PR #11, opened 2020-04-08
chore: add coc link to footer

jbolda:


jbolda:
We may want to move this (see #13), but for now it serves the minimal purpose.

PR #14, opened 2020-04-08
fix header overlap on mobile

mosheduminer:


nklayman:
Fix works for me on a pixel 4, but the header is quite big now. Maybe in the future we should look at auto hiding and revealing on scroll.

mosheduminer:
I agree with @nklayman, the the current setup of three levels of headers (in the docs) takes up a lot of precious vertical space.

nklayman:
And for me on a 18:9 aspect ratio I have extra vertical space, for people on older 16:9 devices their vertical space is even more limited.

mosheduminer:
@nklayman @Laegel what do you think of an implementation similar to the one here: https://css-tricks.com/styling-based-on-scroll-position/ ? I could add it to this PR, or make a separate one.

mosheduminer:
I wrote some JS/CSS to hide the documentation navbar on scroll. I suppose that should go in a separate PR?

nklayman:
Yeah, I'd make a separate PR for that, thanks for adding!

PR #15, opened 2020-04-08
hide headers on scroll

mosheduminer:


nklayman:
I think we should hide the logo bar and about/docs/api/translations nav, but leave the docs navigation when the user scrolls down. This will save more space and I think the docs nav will be the most commonly used navbar. As a bonus a little animation would be cool.

mosheduminer:
Okay, will switch that. And strangely, I have animations, and they work on local, but from some reason they don't show in the preview :(.

jbolda:
It may be something with `start` versus `build`. The preview is pretty much just hosting the static files from `build`.

mosheduminer:
For some reason, `build` changed (in the keyframes) max-height from `initial` to `none`, and opacity from `100%` to `1%`. Changing opacity from `100%` to `initial` in the source did the trick. I'll push after I switch the effect to the top two headers.

mosheduminer:
I'm not quite happy with it, but I can't continue on it now, or in the next few days.

mosheduminer:
Also, the code to hide the headers on scroll is global, I don't know how I would localize it to the docs.

PR #17, opened 2020-04-10
Feature/migrate to v2

Laegel:
Moving from Docusaurus v1 to v2. Still a lot to do before the docs are ready.


nothingismagick:
3:13:24 PM: failed during stage 'building site': Deploy directory 'website/build/tauri' does not exist

nothingismagick:
That is so beautiful. Love the netlify bot too. Wow!

jbolda:
@nothingismagick the build directory is different per version. If we use the file configuration, it will override the "global" settings we have in netlify.

nothingismagick:
I guess in the next step we'd need to fix that standard footer.

nothingismagick (approved):

Laegel:
@nothingismagick Yeah, for now I've not changed the Docusaurus links placeholders, just wanted to have something "ready" first.

PR #58, opened 2020-06-07
Feat/rework configuration

### LINKED ISSUE #30: Improve configuration docs
- Have a better way to show JSON props for tauri.conf.json, https://tauri.netlify.app/docs/config can be cleaner;
- Explain Cargo.toml properties.

Laegel:


docs/api/config.md
```diff
@@ -1,99 +1,152 @@
...
+<Properties rows={[
+  {
+    property: "embeddedServer", type: "object",
+    child: <Properties rows={[{ property: "active", type: "boolean", description: `Set it to <code>false</code> if you plan to serve your application statically.` }]} />
+  },
+  {
+    property: "bundle", type: "object",
+    child: <Properties rows={[
+      { property: "active", type: "boolean", description: `` },
+      { property: "targets", type: "string", description: `The name of the built application. If this is not present, then it will use the name value from your Cargo.toml file.` },
```
lucasfernog: `targets` is an array of the bundles you want to generate; e.g. ["deb", "osx", "msi", "appimage", "dmg"]. By default we bundle everything your target supports (osx/dmg on mac, deb/appimage on linux, msi on windows).

docs/api/config.md
```diff
@@ -1,99 +1,152 @@
...
-Default:
+<Properties rows={[
+  {
+    property: "embeddedServer", type: "object",
+    child: <Properties rows={[{ property: "active", type: "boolean", description: `Set it to <code>false</code> if you plan to serve your application statically.` }]} />
+  },
+  {
+    property: "bundle", type: "object",
+    child: <Properties rows={[
+      { property: "active", type: "boolean", description: `` },
```
lucasfernog: empty description

docs/api/config.md
```diff
@@ -1,99 +1,152 @@
...
+      { property: "targets", type: "string", description: `The name of the built application. If this is not present, then it will use the name value from your Cargo.toml file.` },
+      { property: "identifier", type: "string", description: `A string that uniquely identifies your application, in reverse-DNS form (for example, "com.example.appname" or "io.github.username.project"). For OS X and iOS, this is used as the bundle's CFBundleIdentifier value; for Windows, this is hashed to create an application GUID.` },
+      { property: "icon", type: "string[]", description: `A list of (relative to src-tauri) icon paths to use for your application bundle.` },
+      { property: "resources", type: "string[]", description: `A list of files or directories which will be copied to the resources section of the bundle. Globs are supported.` },
+      { property: "externalBin", type: "string[]", description: `A list of—either absolute or relative—paths to binaries to embed with your application.` },
```
lucasfernog: should mention that each item like `my-binary` will look for `my-binary-${targetTriple}.exe` on Windows and `my-binary-${targetTriple}` on mac/linux.

docs/api/config.md
```diff
@@ -1,99 +1,152 @@
...
+      { property: "copyright", type: "string", description: `A copyright string associated with your application.` },
+      { property: "category", type: "string", description: `What kind of application this is. This can be a human-readable string (e.g. "Puzzle game"), or a MacOS X LSApplicationCategoryType value (e.g. "public.app-category.puzzle-games"), or a GNOME desktop file category name (e.g. "LogicGame"), and cargo-bundle will automatically convert as needed for different platforms.` },
```
lucasfernog: cargo-bundle should be something like "Tauri bundler"

lucasfernog: maybe we should document the possible values:

```
pub enum AppCategory {
  Business,
  DeveloperTool,
  Education,
  Entertainment,
  Finance,
  Game,
  ActionGame,
  AdventureGame,
  ArcadeGame,
  BoardGame,
  CardGame,
  CasinoGame,
  DiceGame,
  EducationalGame,
  FamilyGame,
  KidsGame,
  MusicGame,
  PuzzleGame,
  RacingGame,
  RolePlayingGame,
  SimulationGame,
  SportsGame,
  StrategyGame,
  TriviaGame,
  WordGame,
  GraphicsAndDesign,
  HealthcareAndFitness,
  Lifestyle,
  Medical,
  Music,
  News,
  Photography,
  Productivity,
  Reference,
  SocialNetworking,
  Sports,
  Travel,
  Utility,
  Video,
  Weather,
}
```

lucasfernog (changes requested):

nklayman:
I think it should say which values are required and which are optional.

Laegel:
@nklayman is right, we need to let people know what's required and what's optional.
I think we can stick to the TypeScript way: adding an interrogation mark for optional properties.

But I've got a question: should we keep properties defined as empty or strip them when not wanted?
E.g. "externalBin" in "bundle":
```js
"bundle": {
  // Others...
  "externalBin": []
}
```
, do we need to explicitely keep it as an empty array or can it be removed:
```js
"bundle": {
  // Others...
}
```
?

docs/api/config.md
Laegel: If there's no other possibility than a choice among this list, yup.

docs/api/config.md
Laegel: What happens if we disable this?

nklayman:
I think we should have the option commented out with an example item in the list, ie:

```
"bundle": {
  // Others...
  // "externalBin": ["my-external-bin-executable"] (uncomment and set actual binary name to use)
}
```

This way it is obvious you don't need it but also functions as basic instructions for someone who does want to use it.

Laegel:
That's a nice idea. The problem is that JSON doesn't allow comments, so people would have to uncomment/remove all comments in order to run their apps.
Unless we ship something like https://www.npmjs.com/package/json5.

lemarier:
Good job on the new doc!

> That's a nice idea. The problem is that JSON doesn't allow comments, so people would have to uncomment/remove all comments in order to run their apps.
> Unless we ship something like https://www.npmjs.com/package/json5.

I think you should check [Vercel](https://vercel.com/docs/configuration?query=config#introduction/configuration-reference) (formerly Zeit) as a sample.

nothingismagick:
I'm still missing the `targets` string in the JSON example at the end.

docs/api/config.md
lucasfernog: no bundle, just plain binary file building (faster compile time).

lucasfernog (approved):

PR #70, opened 2020-06-21
Feat/notification docs

Laegel:


docs/api/js.md
```diff
@@ -580,6 +582,47 @@ Makes a DELETE HTTP request
...
+
+Tauri patches the <a href="https://developer.mozilla.org/en-US/docs/Web/API/notification" target="_blank">Notification API</a> but not all methods and properties are available.
+
+Since we're patching the browser API, you don't need to import anything as it lives in the global scope.
+</Alert>
+
+### Types
+
+```ts
+interface NotificationOptions {
```
lucasfernog, line 627: I wonder if we should have all these docs here, even if we're supporting only a couple of props.

docs/api/js.md
```diff
@@ -580,6 +582,47 @@ Makes a DELETE HTTP request
...
+}
+```
+
+### Classes
+
+```ts
+class Notification {
+  static permission: string
+
+  construct(title: string, options?: NotificationOptions)
```
lucasfernog: shouldn't it be "constructor"?

docs/api/js.md
Laegel: I think that it's relevant—for the sake of explicitness—especially if we don't support all these props. If consumers don't know how it works, they'll refer to another doc and will wonder why the options they're passing don't work.
What should be removed?

lucasfernog: then we should add a note that only body and icon are supported

PR #87, opened 2020-07-09
Remove webpack plugin recommendation on mode usage

chippers:
The webpack plugin was used by no-server mode on tauri to bundle the application.  The plugin is now archived and deprecated with a notice that it is no longer needed by tauri.  Recommendations in the #help channel on the discord also include removing the webpack plugin because it's not needed.

This should bring the documentation up-to-date for current server mode recommendations.


nothingismagick (approved):
I thought we removed this long ago. Thanks ❤️

PR #93, opened 2020-07-13
fix: Removed npm usage

Laegel:


lucasfernog:
I guess we can't just force everyone to use yarn, we should teach them to add a `tauri` script on package.json and `npm run` it instead of using npx.

nothingismagick:
No, not npm, npx

docs/usage/development/integration.md
```diff
@@ -30,6 +30,20 @@ npm install tauri --save
   You can install Tauri as both a local and a global dependency, but we recommend installing it locally.
 </Alert>
 
+If you decide to use Tauri as a local package with npm, you will _have to_ define custom scripts to your package.json:
+
+```js title=package.json
+{
+  // This content is just a sample
+  "scripts": {
+    "tauri:init": "tauri init",
```
lucasfernog: I think we just leave a `"tauri": "tauri"` script then the user can run `npm run tauri init` `npm run tauri build`.

lucasfernog (approved):

CreativeCactus:
Is there any reason not to use `npx tauri`? Getting started commands all work fine for me as `npx tauri ...`.

nothingismagick:
We've had very serious problems with npx recently, including it actually using an outdated version. If it works for you, that's great, however we won't promote it again until all security issues are resolved. (also it's slow).

PR #96, opened 2020-07-16
Make the getting started page more welcoming.

adunkman:
We want people to feel comfortable diving in — leaving out statements which make presumptions about the reader’s intelligence can go a long way!

### DIFF
--- a/docs/getting-started/intro.md
+++ b/docs/getting-started/intro.md
@@ -21,9 +21,9 @@ If you find an error or something unclear, or would like to propose an improveme
 
 ### Setting up Your Environment
 
-We are assuming that you know what the command line is, how to install packages on your operating system and generally know your way around the development side of computing. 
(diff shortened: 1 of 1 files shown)


lucasfernog (approved):

lucasfernog:
I like this change. What do you guys think @Laegel @nothingismagick

Laegel:
Hey @adunkman , thanks for the issue. Would you please be a bit more explicit, provide details about what you think is wrong with the current approach?

adunkman:
Sure! By using phrases like "obviously," folks who don’t find that information obvious may feel excluded, or that this product isn’t for them. It can just as easily be left out, to include those folks. 🙂

nothingismagick (approved):
LGTM - and I really appreciate your attention to detail! Thanks @adunkman

PR #103, opened 2020-07-22
feat: Added Ethicalads script and container

Laegel:


nothingismagick:
So, adblocker detection works on some pages, on some there is nothing at all.

But in the console it can't find the ethicalads source... 

GET https://deploy-preview-103--tauri.netlify.app/js/ethicalads.js net::ERR_ABORTED 404

nothingismagick:
So now it works on first load - but not on internal route navigation changes

nothingismagick:
great - now it hooks on every page - sweet

nothingismagick:
the only thing that bothers me a little bit is that on dark mode its hard to read:

image
image

nothingismagick:
Ok - strike that last comment. This seems fine now:

image

nothingismagick (approved):

PR #120, opened 2020-08-18
documentation: fix typo

Toinane:
Just a little typo in API documentation 👍


Laegel:
Thanks @Toinane! Since our Rust docs are generated from our Rust code, this typo lives there. So we need to fix that there if we don't want it to come back again and again.

Toinane:
> Thanks @Toinane! Since our Rust docs are generated from our Rust code, this typo lives there. So we need to fix that there if we don't want it to come back again and again.

Oh I see! So I need to make a PR there 👍

PR #128, opened 2020-09-19
feat(docs): add arch and fedora dependencies

### LINKED ISSUE #88: Add install instructions for linux distributions that are not using apt, e.g. arch, fedora, etc
It would be quite useful to add install instructions for other linux distributions as well, because the names of the needed packages are completly different and there are quite a lot of people out there who using stuff that isn't based on debian.

amrbashir:
This PR adds dependencies for Arch and Fedora. 
Thanks to @Akiyamka for Fedora's dependencies.


Laegel:
Thanks for this PR, @AmrBinBashir !

Issue #33, opened 2020-05-20
Supported web technologies for Webview


Laegel:
Nice suggestion coming from Discord:

> Perhaps someone could create a Union of the browsers that are represented by Webview? Something like: https://caniuse.com/#compare=ie+10,ie+11,edge+80,firefox+74,chrome+80

Requires investigation, definitely an important information.

Laegel:
http://html5test.com/
https://developer.mozilla.org/en-US/docs/Web/API/CSS/supports
https://developer.mozilla.org/fr/docs/Web/CSS/@supports

simonhyll:
I don't think we can make a full on "caniuse" type of page for this in the documentation, but we can certainly add information about how to figure this sort of thing out in the docs. Tracking for 2.0 so we don't forget to bring it up somewhere.

PR #134, opened 2020-10-21
Update setup-windows.md

Dardin-dale:
#33. I am relatively new to Rust, coming from Node/React/Electron background. I've found that using the WebView2 it is easier to port my front-end React code from Electron to Tauri since Electron is also chromium based. It seems to me that WebView2 will eventually become default on Windows but I'm not sure when that will be the case. I see that WebView2 is already linked, but it is something that I missed when going through Tauri my first time.


nklayman:
Added in a quick note about how users of the app may be on Webview1. LGTM now. However, a nice bonus would be some instructions on checking for/forcing webview2 in the distributable.

nklayman (approved):

Issue #124, opened 2020-08-29
[docs] broken link to rust api


drmason13:
The link on this page: https://tauri.studio/en/docs/development/intro
Named "Rust API"
goes to: https://tauri.studio/en/docs/api/rust/tauri/config/fn.get

Which led to the tauri 404: Page Not Found

I'm browsing on Android Chrome.

Laegel:
@drmason13 Thanks for spotting this 404.
The Rust API can still be reached through API > Rust > [crate]. From there, you have a link for each Rust entity (function, struct, ...).

drmason13:
Hiya, I found the broken link.
It's in docs/usage/intro.md

docs/en/usage/intro.md

tauri needs replacing with tauri_api in the URL

https://tauri.studio/en/docs/api/rust/tauri_api/config/fn.get/

Assuming that's the equivalent.

The crate's standard rust docs are also helpful:
https://docs.rs/tauri-api/latest/tauri_api/

Laegel:
Thanks @drmason13 ! This is something that's missing and should be improved, indeed. :+1:

PR #138, opened 2020-12-03
feat(docs) add landing page for rust API (#124)

drmason13:
Started by fixing the broken link, then felt a landing page is better to link to than diving straight into an individual function.
Not wedded to any of the text/content just hope it helps 👍

### DIFF
--- a/docs/en/api/rust.md
+++ b/docs/en/api/rust.md
@@ -1,0 +1,10 @@
+---
+id: rust
+title: 'Rust'
+---
+
+Within the rust folder on this site you can find all of the latest rust api documentation.
+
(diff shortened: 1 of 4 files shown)


drmason13:
Ah the landing page is missing the sidebar! I imagine there's a way to "build" the page properly, I'm afraid I'll need some help to wire all that up as I'm not familiar with docosaurus at all

jbolda:
We appreciate the thoughts and extra set of eyes!

Some of this will likely need to be adjusted upstream though. The "API Docs" are mostly embedded within our crates and then through a combination of rustdoc and https://github.com/tauri-apps/rustdocusaurus, we produce these docs. I haven't had a chance to see where exactly these changes would need to be though.

PR #142, opened 2020-12-15
feat(i18n): Added French docs

Laegel:
French docs build fails because of multiple minor issues:
- `import`s have been put on the same line, triggering syntax errors;
- some frontmatter IDs have been translated and shouldn't have, so we should remove all IDs from English source.


lorenzolewis:
Keeping an eye on https://github.com/tauri-apps/tauri-docs/pull/323 for this as it will impact

lorenzolewis:
Closing this as we're using the Docusaurus and Crowdin integration now.

PR #144, opened 2020-12-30
fix(docs): -- is necessary for `npm run` to pass options to underlying command

rhysd:
In documentation, `--debug` option is passed to `npm run` as follows:

```
npm run tauri build --debug
```

https://tauri.studio/en/docs/development/debugging

However, this does not pass the `--debug` option to underlying `tauri` command. It runs `tauri build` command and release mode is used to bundle the app. This is because the `--debug` option is consumed by `npm run` command. To pass the option to the underlying command, `--` must be put before the option.

```
npm run tauri build -- --debug
```

I confirmed that it works as intended on my local today. This PR fixes the point.

### DIFF
--- a/src/theme/Command.js
+++ b/src/theme/Command.js
@@ -10,13 +10,21 @@ const types = [
   { label: 'Global', value: 'tauri ' },
 ]
 
+function insertDashDashBeforeOption (value, name) {
+  const idx = name.indexOf('--')
+  if (idx === -1) {
+    return value + name
+  }
+  return value + name.slice(0, idx) + '-- ' + name.slice(idx)
+}
+
 export default ({ name }) => {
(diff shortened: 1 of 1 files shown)


rhysd:
I found that simply replacing `--debug` with `-- --debug` does not work because it breaks other tabs. I'll try to fix it and `push -f` the fix.

rhysd:
54b4b4a fixed this issue correctly. I confirmed it on my local machine:

tmp

And I also found that the default value of the tab was not correct. 4812057 fixed it.

Laegel:
Hey @rhysd ! Thanks for that, totally forgot we had to use extra `--` to pass args to the command when using npm. :+1:

PR #147, opened 2021-01-22
Updated config.md with correct description

blankRSD:


Laegel:
Thanks @blankRSD , looks like it was a bad copy/paste here!

PR #182, opened 2021-04-12
Announcement bar have an height bug, it's the fix

MatteoGauthier:
Hey I notice a bug on the height of the announcement bar

image

And when it's fixed

image


Laegel:
Hey @MatteoGauthier , thanks for your PR! 
What's your OS and Chrome version? I can't reproduce that.

MatteoGauthier:
@laegel I'm on Windows 10 1909 with Chrome 89

PR #183, opened 2021-04-14
correct String value 

Ludea:
I notice `window.listen()` takes std::string::String and not $str


lemarier:
Thanks for catching up!

LGTM

PR #193, opened 2021-04-24
added opening curly brackets and missing languages to code blocks

crapStone:


Laegel:
Thanks for that, @crapStone !

PR #200, opened 2021-04-30
added naming for commands

crapStone:


lucasfernog:
This is also documented on the code section here: https://github.com/tauri-apps/tauri-docs/blob/dev/docs/en/usage/guides/plugin.md#writing-a-plugin

Maybe we should use an <Alert> instead?

crapStone:
Sorry I didn't see this.

PR #201, opened 2021-04-30
Add Calciumdibromid to showcase

crapStone:


nothingismagick:
@Laegel - this is great - I think we should however ONLY show projects that are 1.0+ 

what do you think?

6543:
it is at version v0.3.0 but it's already used & core function could be considered as v1 ;)

It's just some "nits" have to be done (refactor pdf generation ...) don't know if put a hard limit  1.0+ is good - but it's your choice anyway ...

crapStone:
The 0.x is only because we wanted to wait for tauri beta (and an issue with PDF generation).
We have a webapp deployed that is already used by the students and i'll finish the migration to `tauri-beta` soon.

After that the only thing stopping us from a 1.0 release is a rewrite of the PDF generation in pure Rust and we will probably take the webapp down, or at least aggressively push the desktop app, because our server is overloaded XD.

If you still think that we should wait until we finished these points I will close this PR and open one after the 1.0 release.

6543:
@crapStone let them decide ...

nothingismagick:
i guess - what I meant - and sorry if I was unclear, is that we should just showcase projects that use tauri 1.0.0 - this project absolutely belongs here. and seriously - PDF generation is such a massive thing, that would be an amazing contribution back to the community! ❤️

6543 (approved):
nice - lgtm will not count, just ok from cabr2 side ;)

nothingismagick:
I guess I assumed you wanted the world to know bout this ;)
https://github.com/Calciumdibromid/CaBr2/blob/14c1fe10f4b8897934af7229451636d97190441f/src-tauri/Cargo.toml#L32

lucasfernog:
package.metadata.bundle isn't read anymore btw.

crapStone:
i got the idea after the big conversation about Angular I started today on the Discord server and our team agreed to this :sweat_smile:

crapStone:
> I guess I assumed you wanted the world to know bout this ;)
> https://github.com/Calciumdibromid/CaBr2/blob/14c1fe10f4b8897934af7229451636d97190441f/src-tauri/Cargo.toml#L32

If you are talking about the `api-all`: this was just a quick-fix because I couldn't find any documentation about the new format

crapStone:
> package.metadata.bundle isn't read anymore btw.

so I should put all of this back into the `tauri.conf.json`?

Laegel:
Awesome work @crapStone & @6543, we'll gladly accept this PR, of course!

@nothingismagick Remove Alpha Tauri projects? I guess we should if we don't want people to find outdated code, as it would be counter productive. But I'd like to let them at least a chance to upgrade to beta before removing them from the showcase!

6543:
well CaBr2 is **now** on **beta**

Laegel:
Just one thing: either change `'cabr2/logo.png'` for `'cabr2/icon.png'` or rename "icon.png" for "logo.png".
And we should also find out why the website isn't building correctly for this PR. 
https://deploy-preview-201--tauri.netlify.app/en/

crapStone:
@Laegel  sry, copy-paste-error XD

crapStone:
We tried to build it locally and it only works when we remove one of the other showcases.
But I couldn't find any indication in the code to why this happens.

6543:
@nothingismagick @Laegel any idear where to look at, that can cause this bug?

Laegel:
Yup sorry, my bad: it lacks a project description but the i18n solution is not good (WIP, got to switch to Docusaurus builtin system instead) and lacks explicitness on how to define it.
To fix it, open tauri-docs/translations/en.json, look for the `pages.showcase.projects` nested property then, on the 4th index, set your project description in a new object.

Laegel:
@6543 @crapStone forgot to tag, cf my previous comment to fix the faulty build.

crapStone:
@Laegel thanks, that fixed it

6543 (approved):
did the job :)

PR #208, opened 2021-05-11
Add Simple STT to showcase

mmpneo:


nothingismagick (approved):

translations/fr.json
```diff
@@ -258,6 +258,12 @@
         },
         {
           "description": "Write Arabic text on apps that don't support bidirectional text."
+        },
+        {
+          "description": ""
```
lucasfernog, line 263: what's up with this empty description? not that it matters since it's just the french unfinished translation

mmpneo: I added it to match the english version

lucasfernog: but why an empty one here?

Laegel: It doesn't matter, you may keep it empty for now as it's not being read, anyway. I'm planning to switch to Docusaurus' builtin i18n soon. :)

mmpneo: because en.json [showcase.projects] had 4 elements in it and fr.json only 3. To match files i added empty description before adding my text

PR #215, opened 2021-05-14
Fix typo in webkit2gtk lib name

cktang88:
From [Wry's page](https://github.com/tauri-apps/wry#debian--ubuntu), the dependency name seems to be `libwebkit2gtk-4.0-dev`. Trying to install `webkit2gtk-4.0` results in a "no package found" error for me.


lucasfernog:
The actual names are below that for each distro, but thanks :)

PR #216, opened 2021-05-14
Update updater.md

dibadibadiudiu:
"relaunch" is no longer in "@tauri-apps/api/app"


lucasfernog:
Thanks for catching this!

PR #222, opened 2021-05-16
Fix gtksourceview3 Fedora package name

sstangl:
The proper package name in the setup instructions is `gtksourceview3`, instead of `gtksourceview-3.0`.

Verified on Fedora 34.


Laegel:
Hey @sstangl ! Thanks for your PR. :+1:

PR #224, opened 2021-05-18
feat: Add benchmarks 

lemarier:


lemarier:
Ok, I think I'm done with that.

Feel free to make any changes, and If you want to add a link in the menu, I didn't add that as I wasn't sure if we wanted to.

Thanks!

PR #235, opened 2021-06-11
WebDriver guide

chippers:
Still a WIP as the interfaces for `wry` and `tauri`/`tauri-runtime-wry` settle and we do a release for them. I want these guides to be ready by the time `tauri-driver` enters pre-alpha so we have some documentation for others to get started with it.

I'll mark for review when `tauri-driver` is ready for the first pre-alpha.


nothingismagick:
@chippers Is there an ETA for this being able to be published?

chippers:
> @chippers Is there an ETA for this being able to be published?
> 

I was planning on publishing when the release with webdriver came out, but ended up being reverted.  I'm in the process of converting the multi window webdriver tests from trying to fix the bug to be usable tests in Tauri and Wry, and plan to either fix or disable the multi window testing for the release this week

lucasfernog:
@chippers this should've been added to the tauri repo first. But since we just released, we can patch in both places.

PR #261, opened 2021-07-24
add `height: 100%;` to showcase cards

lukethacoder:
add `height: 100%;` to cards so they fill the grid nicely.

### before: 
image

### after: 
image


Laegel:
Hey @lukethacoder , thanks for your fix!

PR #267, opened 2021-08-11
Update setup-linux.md

keyneom:
Minor correction on recommended command. It got me while I was attempting to debug an issue. Took a second to realize why the file didn't exist 😂


lucasfernog (approved):
nice catch :) thanks

PR #280, opened 2021-09-01
Add blog tab for future blog posting

wusyong:


lorenzolewis:
Since https://dev.to/tauri is already used for this, I don't think the whole blog functionality is needed (unless there's a want to move away from dev.to). Maybe a redirect in the menu to the dev.to blog instead?

lorenzolewis:
Closing since we're using dev.to for blog. Adjusting wording slightly in #360 to make it a bit more clear.

PR #285, opened 2021-09-26
fix(integration): fix two links

csjkevin:


lucasfernog (approved):

lucasfernog:
Can you PR the same to the tauri repo? That's where the docs file live before getting published.

csjkevin:
> Can you PR the same to the tauri repo? That's where the docs file live before getting published.

@lucasfernog Sure. But I think I didn't find that paragraph in tauri repo. Shall I update it?

lucasfernog:
Hmm i'm not sure, maybe we updated it already.

PR #293, opened 2021-10-15
The default terminal is now zsh

pauleaster:
On a Mac, executing this fails if running zsh, which is now the default:
curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.2/install.sh | bash
Whereas this seems to work:
curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.2/install.sh | zsh

PS: This is my first public PR, so I hope I did it correctly!

### DIFF
--- a/docs/en/getting-started/setup-macos.md
+++ b/docs/en/getting-started/setup-macos.md
@@ -30,7 +30,7 @@ $ xcode-select --install
 We recommend using nvm to manage your Node.js runtime. It allows you to easily switch versions and update Node.js.
 
 ```sh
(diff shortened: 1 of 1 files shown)


nothingismagick (changes requested):
Hmm. Well, maybe we could be even smarter about it and just use whatever the user has that is able to run shell scripts. Maybe confirm if this works for you as it does for me. (also, nvm has been updated, so might as well update the version too):
```
curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.0/install.sh | sh
```

pauleaster:
Yes, this works for me.

nothingismagick:
Thanks!!!

nothingismagick (approved):

PR #294, opened 2021-11-02
add bs redis desktop client project

fuyoo:
add my `redis desktop client project` to the showcase


probablykasper:
You'll need to provide a `description` string for the app in `translations/en.json`. You can see that the deploy preview isn't working, and the log has errors.

PR #301, opened 2021-11-30
add the "BS Redis Desktop Client" project,and the local test passed.

fuyoo:
add the "BS Redis Desktop Client" project,and the local test passed.


.gitignore
```diff
@@ -1,3 +1,6 @@
+# ide
+.idea
+
```
kasper9n, line 3: Any reason for adding this?

translations/fr.json
```diff
@@ -279,6 +279,9 @@
         },
         {
           "description": "ExcelMerger merges Excel files"
+        },
+        {
+          "description": "The Best Surprise Redis Desktop Client. Powered by Rust and Tauri."
```
kasper9n, line 284: Other projects' descriptions are missing here, so I think this would be out of order. I'd suggest just leaving this out (seems like the french docs aren't used yet anyways)

src/pages/showcase.js
```diff
@@ -15,102 +15,109 @@ const projects = [
     title: 'GUIJS',
     image: imgPath + 'guijs/logo.svg',
     link: 'https://guijs.dev/',
-    screenshots: ['0.png'].map((image) => imgPath + 'guijs/' + image),
```
kasper9n, line 18: There's a lot of formatting changes that shouldn't be here, so please remove all those changes. The repo does have prettier configured for formatting.

probablykasper (changes requested):

PR #302, opened 2021-11-30
add the "BS Redis Desktop Client" project

fuyoo:
add the "BS Redis Desktop Client" project


probablykasper:
Thanks, looks good now! (btw, you don't need to open a new PR when you make changes)

PR #314, opened 2022-01-11
Docusaurus upgrade

lorenzolewis:
Upgrade all dependancies. Site needs to be reviewed to ensure all functionality is still as expected.

Motivation behind this is newer versions of Docusaurus provide better accessibility (plus, who doesn't like updates? /s )

The following changes were made to allow build:

- Remove DocVersionSuggestions from DocItem (component no longer exists in module)
- Remove sidebar logic in NavBar component

I don't see any changes visually on the site with the above, but a review would be good in case I missed something.


lucasfernog:
I love updates. I use arch btw.

lucasfernog:
Looks like it broke the release notes component.

lorenzolewis:
I had to set the node version using the `.node-version` file for CI/Netlify to pick up an updated Node version. Not sure if that's the wanted way to do it.

lorenzolewis:
> Looks like it broke the release notes component.

I'll take a look at this tomorrow, should hopefully be a straightforward fix

lucasfernog:
Thanks!

lorenzolewis:
I think some of the CSS is slightly modified as well, seems to be some extra padding around the body that makes it wider than 100%, will add that to the list.

probablykasper:
Ah finally the sidebar dropdowns are fast!!

> Remove DocVersionSuggestions from DocItem (component no longer exists in module)

According to https://github.com/facebook/docusaurus/pull/5052 `@theme/DocVersionSuggestions` was renamed to `DocVersionBanner` (this change appeared in Docusaurus 2.0.0-beta.4).

lorenzolewis:
You're magic @probablykasper, just swapped that in place (after searching for hours this afternoon looking for anything around that change in release notes).

Let's see how this latest commit does with the release notes. I kept getting some high CPU usage and hangs while it was pinging the API in Safari, but Edge seemed to work fine. Will see how it performs in Netlify

amrbashir:
docs sidebar doesn't exist on mobile.
there used to be a menu button on the right bottom corner that opened the docs sidebar.

lorenzolewis:
@amrbashir sidebar on mobile should now be fixed. Note that it's changed positions with the latest version of docusaurus so that it's with the rest of the navigation menu structure (a lot less confusing imo)
Screenshot 2022-01-12 at 10 30 05

What it looks like expanded:
Screenshot 2022-01-12 at 10 30 10

Thoughts on this?

lorenzolewis:
Note that I had to add a dependancy on "@mdx-js/react" which allows for dynamic @ imports in JS from node modules. This should allow for easier automatic updates for any theme updates from docusaurus. I think this will let us remove a lot of older components in `src/theme`, but will probably comb through those in a separate pull request once everything is updated.

lorenzolewis:
Over 100% width issue should now be fixed (caused by some hard-coded search bar CSS)

amrbashir:
Cool, it is almost ready.
there is one issue with search results

amrbashir:
Ideally we should make the search look like the Algolia component used in docusaurus site but that's for another PR. lets focus on fixing the overflow.

FabianLars:
Also, a really small thing, there is some margin or whatever missing on the searchbar/theme switch. It's worse if you hover over the toggle.
Screenshot 2022-01-12 131706 Screenshot 2022-01-12 131830

lorenzolewis:
@amrbashir I think that's just because nothing matches 'dx' (same happens on the current site). Searching for other things seems to be fine.

lorenzolewis:
@FabianLars actually, taking a closer look it seems like the whole nav bar on that side may need a bit of TLC
Before (current live site)
Screenshot 2022-01-12 at 12 38 24

Now (dev site)
Screenshot 2022-01-12 at 12 38 28

lucasfernog:
> @amrbashir I think that's just because nothing matches 'dx' (same happens on the current site). Searching for other things seems to be fine.

I think what he meant was the layout breaking and the results not fully readable. At least it's ok on the website for now. (idk if it was like this before too).

lorenzolewis:
Ah, I see it now. I think the overall search bar experience may take some deeper digging then.  🦖 (docosaurus) changed how it works with the update, so just need to have a read through of the docs. 

The padding for the button itself should be fixed now. Added back in the beta badge and the locale dropdown (using the new 🦖-friendly elements)

lucasfernog:
Let's merge this one then :)

PR #319, opened 2022-01-12
Fix beta badge

lorenzolewis:
This _should_ fix the "beta" button link colour. Need to have it build and deploy to netlify test because I wasn't able to replicate it locally and it only showed up after the upgrade deploy. 

Current:
Screenshot 2022-01-12 at 16 13 01
Dev:
Screenshot 2022-01-12 at 16 13 16


FabianLars (approved):

lorenzolewis:
Alright, this seems to look good to me on netlify now, so I think good to merge if we'd like?

Issue #307, opened 2022-01-06
Add support for no javascript and keyboard navigation


iacore:
Currently, the documentation is unusable without Javascript. For example, on page https://tauri.studio/en/docs/api/cli, the side bar / links are useless when clicked.

Also, when navigating with keyboard, <kbd>Space</kbd> on "Skip to content" does not scroll the page to start of content (it should be a link with `href="#start"`). Navigating backwards is confusing, since the title bar has `display: absolute` or something, and covers the first few lines of content, so I can't see the element in focus. Try Shift+Tab on the above page and you will understand.

nothingismagick:
this is not something we can fix afaik - because we use docusaurus. sorry :(

iacore:
The docusaurus documentation is usable without Javascript. https://docusaurus.io/docs

Here's a better reference of sidebar navigation: https://www.typescriptlang.org/docs/handbook/advanced-types.html

nothingismagick:
Oh wow! I didn't know that!

lorenzolewis:
@locriacyber can you test the navigation sidebar on this test deploy to see if the issue is fixed for you: https://deploy-preview-314--tauri.netlify.app/en/docs/api/cli/

Could you elaborate a bit more on the navigating with keyboard point? Give it a try on the above site as well as it could possibly be fixed.

iacore:
- The "Skip to main content" link still needs JavaScript (incorrect `a[href]`).
- links (`<a>`) do not change color or anything on focus. I don't know which link I'm focusing.
- link color too light, too-little contrast with white background
- nav bar covers main content (please don't use `position: sticky`); Github uses this feature as well, but in Firefox, the browser considers text covered by sticky elements "visible".

![image](https://user-images.githubusercontent.com/74560659/149811629-de86e2e6-52ea-496d-b759-c6f3626ba79f.png)

lorenzolewis:
> The "Skip to main content" link still needs JavaScript (incorrect a[href]).

> nav bar covers main content (please don't use position: sticky); Github uses this feature as well, but in Firefox, the browser considers text covered by sticky elements "visible".

Out of curiosity, can you test these two points on https://docusaurus.io/docs with the same setup? If it has the same behavior there then I'm afraid that's something upstream that we can't do anything about unless they adjust.

For the overall contrast, we're working on a redesign and I'll add the other two styling points to get covered under that

iacore:
[...]

PR #333, opened 2022-01-14
[Draft] Redesign

lorenzolewis:
Initial work on redesign

# To do:

- [ ] Add `:hover` to `<a>` elements #307
- [ ] Add light theming for high-contrast #307


amrbashir:
I don't know how this looks on desktop but on mobile the menu button has a big margin than it should IMO, ideally it should match the margin of the content below.

lorenzolewis:
Going to close this for now due to build minutes

PR #337, opened 2022-01-15
i18n

lorenzolewis:
This is a large chunk to change i18n over to Docusaurus's built-in capability (https://docusaurus.io/docs/i18n/tutorial)

# Changes:

- Remove French docs (none had been translated yet)
- Move English docs up to be directly under the `docs` folder (serves as the default)
- Remove old translation files
- Remove references to old translation files in docusaurus.config.js, index.js, partners.js, and showcase.js, swapping in place Docusaurus's `Translate` HTML element and `translate()` function
- Remove unused users.js file
- Remove `supported-languages.js` (this is now contained by default in the docusaurus.config.js file)
- Add `translate` command to `package.json` that will auto-generate enabled languages
- Add generated translation files derived from `translate` (note: this will require re-work on Crowdin)
- Add MIT license to `package.json` (removes an annoying warning when building)
- Update supported browsers (removes an annoying warning when building)
- Remove `t`, `languages`, and `language` objects from `docusaurus.config.js` that were referencing the old translation files
- Update `netlify.toml` to now simply run `yarn build` (note: this will not yet generated the translations on CI/CD, this will need to be updated later)

# Remaining items (to be done after this PR is merged):

- Update Crowdin integration
- Add translation generation, uploading and downloading to/from Crowdin to CI/CD
- Update actions in https://github.com/tauri-apps/tauri and https://github.com/tauri-apps/typedocusaurus to point to new folders and to stop generating sidebars (since they're now dynamically generated) 

# Testing needed
Please put this deployed site through everything you can to make sure there aren't any breaking changes. We may need to look at things like setting up redirects for /en/ to point to the root depending on how the URLs are generated.

For now I've enabled the French locale to enable testing that, but once deployed that will get disabled since we don't have the translation linked in.

# To Do

- [x] Icons (i.e. roadmap) seem to be broken
- [x] Add redirect from /en/* to /*
- [ ] Remove French translation before merging


lorenzolewis:
Note to self: add flag to only build English in test deploy, all languages for prod deploy after testing

docusaurus.config.js
```diff
@@ -1,40 +1,167 @@
...
   projectName: 'tauri',
-  baseUrl: `/${language}/`,
+  baseUrl: `/`,
   favicon: 'img/favicon-32x32.png',
   url: 'https://tauri.studio',
   onBrokenLinks: 'ignore',
-
+  i18n: {
+    defaultLocale: 'en',
+    locales: ['en', 'fr']
```
lorenzolewis: Remove fr before pulling

FabianLars (approved):
I already looked through it a few times and the only thing that bugged me was the missing redirect (which you added). So because you mentioned it in the meeting i'm gonna approve this from my side already. That's without thinking about crowdin compatibility tho.

PR #338, opened 2022-01-16
Codeblock simplification

lorenzolewis:
# Summary of Changes
- Add reference to custom theme and Rust as an additional language in `docusaurus.config.json`
- Add custom code block theme in `src/theme/CodeBlockTheme.js`
- Remove highlight.js, react-highlight, and react-infinite-scroller from dependencies (not currently referenced in any part of the code)
- Remove custom CSS for code block theming
- Remove custom CodeBlock theme component to fallback to Docusaurus default
- Update `Command.js` to reference Docusaurus CodeBlock

Any customisation can be done in `src/theme/CodeBlockTheme.js`


FabianLars:
Honestly i don't like the "new" colors for codeblocks. Maybe it's my color blindness kicking in, but it's sooo much harder on the eyes.

The main problem for me is the background color. The syntax highlighting colors (after changing bg) are fine, not perfect but fine. Also, the colors make the CLI command docs look like it's "disabled", like commented code in vs code or something.

lorenzolewis:
I haven't really customised it beyond the theme, but I guess I'll go through and replicate it from the live site and then we can always adjust/customise beyond that as part of the larger redesign

lorenzolewis:
@FabianLars let me know what you think of the colours now

FabianLars:
way better, thanks :pray:

Issue #340, opened 2022-01-17
JavaScript API needs an index page


dragonworx:
Following the link for the JavaScript API on `docs/development/intro` leads to a 404 as there is no `https://github.com/tauri-apps/tauri-docs/tree/dev/docs/api/js` page for the JavaScript API on `docs/api/config`.

**Expected behavior:** Clicking the link on `docs/development/intro` would land on an index page showing the Classes, Enums, Interfaces, Modules pages with preferably a short intro and any other useful links.

**Current behavior:** Clicking the link on `docs/development/intro` leads to a 404, requiring additional user navigation to get to JavaScript APIs.

lorenzolewis:
Fixed and deployed. Thanks for catching this!

PR #344, opened 2022-01-19
Fix links

lorenzolewis:
- Add Architecture file
- Fix Architecture broken links
- Fix web driver broken link
- Fix links to development cycle
- Remove `onBrokenLink: ignore` from docusaurus.config.js
- Exclude rust docs from being built
- Add auto-generated index for JS API docs


lorenzolewis:
Fixes #340

PR #358, opened 2022-01-21
fix tauri-cli install command on linux zsh

dardourimohamed:


FabianLars:
What system/versions are you on? I can't reproduce this on any of my systems 🤔 
But, i will close this anyway, because a) the beta.7 version was broken for the last couple of weeks anyway and b) this change would be overwritten ~next week anyway :/
Thanks tho ❤️

dardourimohamed:
I'm on Manjaro

PR #367, opened 2022-01-22
fix(docs): fix minor typo

Edqe14:
This existed on the main repo [comments](https://github.com/tauri-apps/tauri/blob/52723ee8a1/tooling/api/src/os.ts#L8) too but I believe that this should be `os` instead of `fs` based on the rest of the modules descriptions.

### DIFF
--- a/docs/api/js/modules/os.md
+++ b/docs/api/js/modules/os.md
@@ -4,7 +4,7 @@
 
 Provides operating system-related utility methods and properties.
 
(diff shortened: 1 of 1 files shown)


lorenzolewis:
/upstream tauri-apps/tauri

API documentation is kept in https://github.com/tauri-apps/tauri.

An issue is opened there and the changes will get rolled out here once fixed. Thanks!

lorenzolewis:
Apparently the bot is being a bit lazy today, but I put an issue in here for you: tauri-apps/tauri#3264

PR #368, opened 2022-01-23
Add `--locked` To CLI Installation Instructions

zicklag:
I was following the instructions on the site and the CLI failed to build when I didn't provide the locked flag. Not many users will know about that flag so hopefully this helps!


FabianLars (approved):
Thanks. I'm gonna merge this just in case something comes up and we can't release the next version as planned (soon™️).

Also, you may want to install the cli from git instead, quite a few fixes and features in there :)

zicklag:
Awesome, thanks for the tip!

PR #370, opened 2022-01-23
Fix "Platform-specific" md header

HeyITGuyFixIt:
The "Platform-specific" header was a level 2 header, but it probably should be a level 4 header.


FabianLars:
We have the same "problem" in many places. So this is something we need to fix in the api doc comments over at the tauri repo instead. Thanks tho ❤️

PR #373, opened 2022-01-25
add plugin conventions section

JonasKruckenberg:


docs/guides/plugin.md
```diff
@@ -91,6 +91,13 @@ fn main() {
 }
 ```
 
+## Conventions
+
+- Plugins should have a clear name with `vite-plugin-` prefix.
```
FabianLars: vite masterrace 🙌

FabianLars (changes requested):

docs/guides/plugin.md
JonasKruckenberg: noooo I didn't copy and paste why you asking

FabianLars (approved):

PR #376, opened 2022-01-26
docs: yarn getting started page url changed

chartrandf:
Yarn getting started URL changed.


FabianLars (approved):
Thanks for catching that. We were a little bit too aggressive on the "getting" -> "get" renaming 😅

PR #379, opened 2022-01-27
add redirects to handle outdated links in the wild

FabianLars:
There are too many links out there we broke with restructuring PRs. We can't really do double redirects and i didn't want to add one for /en and /, so i just created redirects for /en, because that's what should be the main problem for us anyway.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -3,6 +3,49 @@
   command = "yarn build"
 
 [[redirects]]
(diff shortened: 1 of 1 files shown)


lorenzolewis:
This looks good to me

FabianLars:
Yeah i'm just gonna merge this, it doesn't break anything and testing if i missed something "major" is not really possible with the temporary link...

PR #392, opened 2022-02-04
shift around sidebar for better comprehensive flow

jbolda:
## Motivation

We felt that the docs needed some ordering adjustments to improve the user experience. This sets us up for better user on boarding, and we believe easier "hook" points for the someone to come back to the docs and look for a specific item.

## Approach

This was built upon feedback from both the direct community and extending out to indirect.


FabianLars:
Disclaimer: I know this is more like a general setup and not a ready-to-go-live PR, so most of the stuff down below is more like a brain dump and stuff i probably missed cause i didn't join the meeting in time...
-> Actually, some notes about the bigger picture you guys have in mind with this would be appreciated :pray: 

- Distribution > Updating Packages
  - Doesn't make sense as it's about npm/cargo packages, not the tauri app itself. This is a problem the live version has too as it's position there suggests the same (but "less obvious" than here)
  - It doesn't 100% fit anywhere imo, but i guess Development is the best choice?
- Like amr said on discord: Rename "Your First Tauri App" back to "Introduction" again or whatever else would fit.
  - nvm, actually looking at it, the name is alright, but the position of the item is wrong, should be _after_ the 3 Setups. You can't follow it without having the setup completed so it's kinda weird.
- Also not really a fan of moving back to _getting_-started, but mainly because we just changed that, **not** because i prefer "get" - "getting started" sounds nicer in my angry-german-brain lol and it would remove a redirect again (not that this is worth much if we have to add another 10 lol)

The following things are the consequences of joining the meetup too late...
- are there rough plans/ideas for the Architecture subnav? Pretty sure there are, but i can't come up with stuff myself and only the patterns inside that is kinda weird.
- Debugging and Testing have a weird vibe too, but probably just because they both contain only 1 item, but i get that this pr is just some kind of preparation (although i have no idea what else we can add to Debugging tbh, kinda feels like it would be better suited in Development).
So, it's most likely just my lacking imagination of what's about to come, but to me the sidebar dropdowns are just too much idk, like they don't add much value (yet?) and instead add obfuscation, if that makes sense?

jbolda:
Thanks for the comments @FabianLars!

- Distribution > Updating Packages
  - The thought here is that one wouldn't need look at setting up their up for updates until they are ready or have already published the app. As long as you are working on it, the initial development phase where you would look at the `Development` Section, you don't need to know about it.
- Debugging and Testing
  - These are more containers that we expect to expand upon. Right now we only have one doc, but the expectation that we have is more will be added there. See the image below for some suggested items that will go there.

This a mock that we used as a talking point that @JonasKruckenberg cooked up.
image

Regarding the Architecture Nav, we imagine that that will have an introduction type page explaining how things fit togethor. And then patterns help one decide how to put together the Tauri pieces of the app. The other things we will include are items such as the "Build Tools" and "Frontend Frameworks" as noted in that image, but also things like other languages and how they fit into Tauri (via a sidecar or we can note future expected bindings).

lucasfernog:
"host testing" will also be a topic soon, we're working on a mock tauri runtime for Rust tests purposes.

lucasfernog:
Also if `Build Tools` is actually shown on the docs somewhere, we should consider renaming it, since the name clashes with the Microsoft Visual Studio Build Tools.

FabianLars:
@jbolda Thanks. Makes sense so far. Most of it indeed just breaks down to me not being able to imagine how the docs (content) could look like in the future 😅 (I somewhat can do that now tho, so thanks for that!)

- Distribution > Updating Packages
Just to double check, we're both talking about this page, right? https://deploy-preview-392--tauri.netlify.app/docs/distribution/updating
Or is the content itself meant to be about the updater feature (or whatever else, just not what it currently is)?
Otherwise i still think this doesn't belong there. I already had this discussion with some people about the _current_ docs, where the order of `"App Publishing" into "Updating Packages"` is suggesting that the page is about updating the app, not the dependencies. This "suggestion" would be even worse in the new layout imo.

Tbh, i'd even argue in favor of just removing the instructions on how to update dependencies from the docs all together...

JonasKruckenberg:
I agree with @FabianLars the updating tauri guide feels weird in the distribution section, even if it's a bit too advanced for `Development` that's the section where it feels right.

Or if you feel like it doesn't make sense in `Development` either, we can make it just another guide

jbolda:
Hahaha, I did the same thing actually @FabianLars with `Updating Packages`. Let's move that 👍 .

jbolda:
I think we have covered all of the main concerns in the mob review and discussion here and in discord. Going to merge this in and we can make adjustments from there.

PR #394, opened 2022-02-05
feat: Generate json AST for js api. Add cli and config doc generation.

FabianLars:
Edit: Point 5 got added after the others, so it kinda clashes with them.

0) Ken needs to confirm that this is the correct file 😂 
1) I'm open for filename suggestions 😅 
2) Do we want the file in a subdir instead?
3) Not 100% sure if we want that .gitattributes stuff, but i thought it doesn't hurt to add it.
4) Pretty print the json file, yes or no? (it's not really supposed to be human-readable, right?).
5) **Important**: Do we want/need them versioned (and include every file, not just the last one)? -> `generated/tauri_api-1.0.0-beta.8.ast.json`
    - Can the new tauri-search automatically pick up the new one // can it pick up the one it needs (if it doesn't index for the latest docs)?

https://github.com/tauri-apps/tauri/pull/3335


lorenzolewis (commented):
The only thing I'm not a fan of is having the files at the root. Is there some folder that would make sense to put these type of "auxiliary" files in?

lorenzolewis:
Also, looks like the CI failed. I wish you the very best of luck!

FabianLars:
> The only thing I'm not a fan of is having the files at the root. Is there some folder that would make sense to put these type of "auxiliary" files in?

Yeah that's one of my questions too, i don't have enough imagination to find a folder name for that lol

> Also, looks like the CI failed. I wish you the very best of luck!

Yep, it needs the PR at the main repo to be merged first.

lorenzolewis:
https://docusaurus.io/docs/api/plugins/@docusaurus/plugin-content-docs

Check out the exclude parameter in the config here, I say put it in the respective API folder.

FabianLars:
Ohh, i like that one. It should also handle the versioning for us (once we implement that for the docs). Gonna wait for more feedback before changing that tho (it's getting late anyway)

lorenzolewis (commented):
This looks good from a Docusaurus perspective, waiting on feedback from @ksnyde to make sure this works with the way he's setting up tauri-search

FabianLars:
It's already used in tauri-search, but from this branch, so once we merge this, we should keep the branch a lil longer

lorenzolewis:
Sounds good then

lorenzolewis (approved):

lorenzolewis (dismissed):
So we're somewhat in sync, can you update the update-docs workflow here as well per the conversation in Discord: 
```
"generate-docs": "typedoc src --githubPages false --readme none --entryDocument index.md --hideInPageTOC true --out docs --json docs/js-api.json --pretty false"
```

FabianLars:
Good call, already forgot that wanted i to that lol

FabianLars:
Should be good to go now.
Btw, merging into `next` instead of `feat/v1` in case we release the release candidate before merging the v1 docs. It is also really unlikely that this PR will introduce any merge conflicts.

yankeeinlondon (approved):

PR #403, opened 2022-02-09
v1-rc docs

JonasKruckenberg:
This includes the reworked docs structure while keeping the old theme and landing page


lorenzolewis:
Should this be opened against the v1 branch instead?

lucasfernog:
I believe we're publishing this one for the RC instead of the v1 refactor since the style is not ready yet.

lorenzolewis:
Got ya, been MIA between work and vacation so I'm all kinds of behind 😆

FabianLars:
> since the style is not ready yet.

And more importantly we want v1 to be an even bigger "stunt" (somewhat of a quote from amr on discord :P )

FabianLars (approved):

PR #415, opened 2022-02-11
Isolation Pattern Documentation

chippers:
Old patterns have been renamed to recipes.  New pattern section has Brownfield and Isolation patterns, and a link has been added to the Security page to the Isolation pattern guide.

Semi-related topic: We should really clean up the recipes, many of them don't really make sense


chippers:
Hopefully I fixed all the links cause of the Pattern -> Recipe change. I didn't touch `book.md` cause I wasn't to sure when it was referring too Recipes or just the work patterns

docs/architecture/patterns/isolation.md
```diff
@@ -0,0 +1,153 @@
...
+application may see.
+
+The largest threat model described above that the Isolation pattern was designed in mind with was Development Threats.
+Not only do many frontend build-time tools consist of many dozen (or hundreds) of often deeply-nested dependencies, but
+a complex application may also have a large amount of (also often deeply-nested) dependencies that are bundled into the
+final output.
+
+## When
+
+The best time to use the Isolation pattern is when you have any of the `allowlist` Tauri APIs enabled. As the developer,
```
lucasfernog: the isolation pattern is also useful if you define custom commands right? brownfield is not as dangerous depending on the command and its inputs, but still useful

docs/architecture/patterns/isolation.md
```diff
@@ -0,0 +1,153 @@
...
+you can utilize the secure Isolation application to try and verify IPC inputs to make sure they are within
+some expected parameters. As an example, you may want to check that a call to read or write a file is not trying to get
+to a path outside your application's expected locations. Another example is making sure that a Tauri API HTTP fetch
```
lucasfernog: could also use the hook to perform control flow validations (e.g. ensuring closeSplashscreen() or login() is only called once), though ideally we'd need a hook to tell the isolation app when a message succeeds or fails

lucasfernog (approved):
Amazing writing :) I just added a couple comments but it LGTM

docs/architecture/patterns/isolation.md
chippers: yeah actually CSRF is one of the core ideas i had in mind when creating the isolation pattern, but wasn't sure of the best way to introduce it to an actual application

lucasfernog (approved):

PR #426, opened 2022-02-12
feat: Add App Size best practices

JonasKruckenberg:
~~TODO: Maybe we can find a better name for this Document 🤔~~


lucasfernog:
Something like `Optimizing the bundle size`?

JonasKruckenberg:
> Something like `Optimizing the bundle size`?

Maybe `Reducing the app size`? Bundle can be confusing because it can mean either the frontend js bundle or the whole app bundle

Edit: I like `Reducing the App Size` it fits the rest of the docs

docs/building/app-size.md
```diff
@@ -0,0 +1,354 @@
...
+And example from a typical `tauri.conf.json`:
+
+```json
+{
+  "tauri": {
+    "allowlist": {
+      "all": false,
+      "fs": {
+        "writeFile": true,
+        "writeBinaryFile": true
```
FabianLars: writeBinaryFile doesn't exist anymore

FabianLars (changes requested):

FabianLars (approved):

PR #428, opened 2022-02-12
fix(macos-setup): remove homebrew instructions

FabianLars:


lorenzolewis:
What's the context for this one? Was it because this step is no longer required?

FabianLars:
Yep. But it was probably _never_ needed (or at least not needed for a really long time).

lorenzolewis (approved):

PR #450, opened 2022-02-14
Fix squished featured images on the home page

DavidBradbury:
Hi there, think Tauri is great! Just noticed the images on the front page were squished and thought I'd send in a quick fix.


FabianLars (approved):
wow they really looked weird lol. Thanks for the fix ❤️

PR #453, opened 2022-02-16
feat: autogenerated sidebar

JonasKruckenberg:
This makes the `Docs` sidebar autogenerated from the folder structure as we discussed previously. 
The biggest benefit here is that we can use `README.md` files to have category index pages instead of introduction pages.

I know this break *a lot* of links, but I guess as @lorenzolewis said, now is the time to break links (especially with the new *you know what* website migration).


FabianLars:
Before i actually look at the pr: we need the templates to generate the cli and config docs

JonasKruckenberg:
> Before i actually look at the pr: we need the templates to generate the cli and config docs

Oups 😂 I wondered what they were for. Gonna add them back

FabianLars:
You can add them somewhere else too if the docs folder is inappropriate now, just need to adapt the scripts.

JonasKruckenberg:
It's actually fine in that folder, we only autogenerate from files inside the `docs/manual` folder

lorenzolewis:
Another alternative to having the numbers in the directories (01, 02) is to have a _category_.json file in each directory. Either way is fine but wanted to put this out there as well. 

https://docusaurus.io/docs/sidebar/autogenerated#autogenerated-sidebar-metadata

docs/api/cli.md
```diff
@@ -167,4 +167,4 @@ This command will show the current version of Tauri.
 
 ## CLI usage
 
-See more about the usage through this [complete guide](/docs/development/development-cycle).
+See more about the usage through this [complete guide](/docs/manual/02-Development/Development Cycle).
```
lorenzolewis, line 170: Not a fan in cases like this where the numbers are hard-coded in the URL

FabianLars:
Ah damn i didn't save the comment again. Wanted to ask if we can do the same without the numbers, that's pretty bad if we want to add something in the middle for example

docs/api/cli.md
JonasKruckenberg: They are actually not, see this https://docusaurus.io/docs/sidebar/autogenerated#using-number-prefixes

They get stripped when compiling the mdx, so they are just for development. The numbers just tell docusaurus which order the categories should have without needed an extra config file

JonasKruckenberg: ahh saw you already linked that doc 😄

JonasKruckenberg:
I removed them numbers

JonasKruckenberg:
goooood why would this friggin link not work 😭

docs/fragments/imagemin.md
```diff
@@ -0,0 +1,6 @@
+:::caution
+The `imagemin` library is [officially unmaintained][imagemin is unmaintained].
```
lorenzolewis, line 2: Curious what the purpose of this file is and where the "fragments" folder fits into the rest of the docs structures...?

docs/manual/architecture/README.md
```diff
@@ -192,4 +192,31 @@ The application can poll this server to see if there is a new release. When ther
...
+[2]: https://github.com/tauri-apps/tao
+[winit]: https://github.com/rust-windowing/winit
+[3]: https://github.com/tauri-apps/wry
+[tauri-hotkey-rs]: https://github.com/tauri-apps/tauri-hotkey-rs
+[binary-releases]: https://github.com/tauri-apps/binary-releases
+[tauri-action]: https://github.com/tauri-apps/tauri-action
+[create-pull-request]: https://github.com/tauri-apps/create-pull-request
+[vue-cli-plugin-tauri]: https://github.com/tauri-apps/vue-cli-plugin-tauri
+[tauri-vscode]: https://github.com/tauri-apps/tauri-vscode
+[documentation]: https://tauri.studio/en/docs/guides/plugin
```
lorenzolewis, line 221: Ideally anything like this we would have as a relative link (from the perspective of starting in the docs folder I believe). I'm not 100% sure if the broken-link check will catch it like this because it may consider it an external URL.

docs/manual/getting-started/setting-up-windows.md
```diff
@@ -75,12 +75,12 @@ If you have problems of any kind after following these instructions, we recommen
...
 [Yarn]: https://yarnpkg.com/getting-started"
 [pnpm]: https://pnpm.js.org/en/installation
 [rustup x86_64]: https://win.rustup.rs/x86_64
 [rustup i686]: https://win.rustup.rs/i686
 [Rust]: https://www.rust-lang.org/
 [rustup]: https://rustup.rs/
 [Microsoft Visual Studio C++ build tools]: https://visualstudio.microsoft.com/visual-cpp-build-tools/
-[Linux specific instructions]: /docs/getting-started/setting-up-linux
\ No newline at end of file
+[Linux specific instructions]: /docs/manual/01-Getting Started/setting-up-linux
```
lorenzolewis, line 86: Remove number

sidebars.js
```diff
@@ -11,94 +11,107 @@ const about = [
...
-      'distribution/updater',
-      'distribution/sign-windows',
-    ],
+    type: 'autogenerated',
+    dirName: 'manual', // '.' means the current docs folder
   },
-  'faq',
 ]
 
+  // {
```
lorenzolewis, line 21: Feel free to remove all of this if you're happy with the way the sidebar renders ;)

docs/api/cli.md
lorenzolewis: I think the number prefix still needs removed here

sidebars.js
```diff
@@ -127,7 +140,7 @@ const community = ['community/contributor-guide', 'community/ci-cd']
 
 module.exports = {
   about,
-  docs,
+  manual,
```
lorenzolewis, line 143: If we're now calling this "manual", we should update the navbar/footer to also be Manual instead of Docs. This is originally where I had the ideal of calling it "guides" instead of "docs", but I know the discussion has already been had and "guides" is a different thing.

lorenzolewis (changes requested):

docs/manual/architecture/patterns/isolation.md
```diff
@@ -55,32 +50,32 @@ the messages after being encrypted, new keys are generated each time your applic
...
 and that you probably already use it every day under the hood with TLS.
 
 There is also a cryptographically secure key generated once each time the Tauri application is started. It is not
 generally noticeable if the system already has enough entropy to immediately return enough random numbers, which is
 extremely common for desktop environments. If running in a headless environment to perform
-some [integration testing with WebDriver](../../testing/webdriver/introduction.md)
+some [integration testing with WebDriver]
 then you may want to install some sort of entropy generating service such as `haveged` if your operating system does not
-have one included. <sup>Linux 5.6 (March 2020) now includes entropy generation using speculative execution.</sup>
+have one included. `<sup>`{=html}Linux 5.6 (March 2020) now includes entropy generation using speculative execution.`</sup>`{=html}
```
chippers, line 78: what is this change meant to do? it breaks the superscript rendering. if that was intended just make it it's own line and remove the superscript tags
image

docs/fragments/imagemin.md
JonasKruckenberg: Fragments is in anticipation of the Tauri + <Bundler/Framework> guides. 
Those docs will share a lot of content and extracting common parts into shared files will make it easier to keep them in sync

jbolda (commented):
Some of the categories are now actual pages while some are not. I think this is pretty confusing especially if you are skimming through the list trying to find a specific doc.

I especially noticed it in the first `Getting Started` section. The category is the tutorial, and now the prereqs come after the tutorial which feels odd.

image

I also liked the `Setup` category as it grouped all of the OS in a place that could be easily visually skipped. (Imagine as we add more targets as well.)

image

JonasKruckenberg:
> Some of the categories are now actual pages while some are not. I think this is pretty confusing especially if you are skimming through the list trying to find a specific doc.


Yep agreed. IMO we should have category pages for *all* categories. I will at least add stub introductory pages.


> I also liked the `Setup` category as it grouped all of the OS in a place that could be easily visually skipped. (Imagine as we add more targets as well.)

Agreed as well, I think this got lost in translation, I'll add the section back 👍🏻

docs/manual/architecture/patterns/isolation.md
JonasKruckenberg: Stupid pandoc, fixed!

sidebars.js
JonasKruckenberg: Hmm, so I don't really think this should be an outwards facing change. I mean I like *manual*, but I picked this name bc the folder name `docs` is already reserved for the root folder. For users facing names, I would still cal it just `docs`. Or wdyt?

lorenzolewis:
@JonasKruckenberg is this still something you all were wanting to do?

lorenzolewis:
Since this is being handled by @JonasKruckenberg in another repo, I'm going to close this. I'll work on migrating to auto-generated sidebars in a different PR since there have been a lot of changes.

PR #461, opened 2022-02-19
list yarn first and specify it is our own package manager for v1

chippers:
it's a bit disingenuous to list pnpm first and remove the yarn recommendation descriptions, especially since we use yarn for v1

### DIFF
--- a/docs/getting-started/setting-up-linux.md
+++ b/docs/getting-started/setting-up-linux.md
@@ -101,7 +101,7 @@ If you have any problems with nvm, please consult their [project readme][nvm].
 
 You may want to use an alternative to npm:
 
-- [Yarn] - Used by the Tauri team for v1
+- [Yarn@v1] - Used by the Tauri team for v1
 - [pnpm] - Alternative package manager focusing on decreasing disk space and installation time
 
 ## 3. Rustc and Cargo Package Manager&nbsp;<Icon title="control-skip-forward" color="warning"/>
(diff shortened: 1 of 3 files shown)


FabianLars:
It used to be "preferred by" and that wasn't true at all so we decided on discord to just remove it all together, because "both have their good/bads" (quote from lucas).
The re-ordering happened because the first idea was to show that pnpm is preferred by most of the (active) members now (still included the "preferred by" suffix). And ngl i was a fan of the alphabetical ordering too lol.

I can def live with your phrasing tho and honestly who the f cares so i'm just gonna merge it 🤷 

Edit: Nvm about the merge just yet. Do we want to mention that it's explicitly about yarn v1?

chippers:
as-is it already explicitly mentions v1. the decision on discord was basically dont rock the boat for something that doesn't matter that much, which is why we stopped the pnpm PR

FabianLars:
Yes it mentions tauri@v1 but not yarn@v1 and we link to the new yarn too so we really don't make any suggestions that we ~~don't like yarn berry at all~~ use yarn 1 exclusively.

chippers:
> Yes it mentions tauri@v1 but not yarn@v1 and we link to the new yarn too so we really don't make any suggestions that we ~don't like yarn berry at all~ use yarn 1 exclusively.

i see what you mean

FabianLars (approved):

PR #464, opened 2022-02-19
Fix yarnpkg.com link typo that was returning a 404

andrewplus:
A simple one byte removal from the reference link.


FabianLars:
Fixed/changed in #461. Big thanks anyway!

PR #466, opened 2022-02-20
Removing Node.js as a Tauri requirement

Laegel:
Since Node.js is not a Tauri dependency anymore; removing it from the setup guides makes them lighter.

### DIFF
--- a/docs/getting-started/beginning-tutorial.md
+++ b/docs/getting-started/beginning-tutorial.md
@@ -1,10 +1,14 @@
 import Command from '@theme/Command'
 import Link from '@docusaurus/Link'
 
-# Your First Tauri App
+# Your First Node.js-based Tauri App
 
 :::caution
 You must have completed all the steps required for setting up the development environment on your machine. Please see the [setup page for your operating system][Prerequisites] if you haven't done this yet.
+:::
+
+:::info
+In this guide, we are assuming you already have installed [Node.js](https://nodejs.org/) and a package manager of your choice.
 :::
 
 There are two ways to integrate with Tauri depending on your needs:
--- a/docs/getting-started/setting-up-linux.md
+++ b/docs/getting-started/setting-up-linux.md
@@ -11,7 +11,7 @@ import TabItem from '@theme/TabItem';
 
 <Intro />
 
-## 1. System Dependencies&nbsp;<Icon title="alert" color="danger"/>
+## 1. System Dependencies 🧰
 
 <Tabs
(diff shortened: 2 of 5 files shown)


docs/getting-started/beginning-tutorial.md
```diff
@@ -1,12 +1,16 @@
 import Command from '@theme/Command'
 import Link from '@docusaurus/Link'
 
-# Your First Tauri App
+# Your First Node.js-based Tauri App
```
JonasKruckenberg, line 4: I think Node.js based sets wrong expectations. People will assume the actual app is node based and never stop asking questions 😅

Maybe you can change the alert to something like: "The follow example uses a JavaScript build tool, so make sure to have Node.js and your favorite package manager installed."

docs/getting-started/beginning-tutorial.md
```diff
@@ -1,12 +1,16 @@
...
 
 :::caution
 You must have completed all the steps required for setting up the development environment on your machine. Please see the [setup page for your operating system][Prerequisites] if you haven't done this yet.
 :::
 
+:::info
+In this guide, we are assuming you already have installed [Node.js](https://nodejs.org/) and a package manager of your choice.
```
JonasKruckenberg, line 11: Let's add links to npm, yarn and pnpm here as well, so rust devs don't get lost

docs/getting-started/beginning-tutorial.md
Laegel: Mmmmh yeah, "Your First Node.js-based-frontend Tauri App" would be more accurate but perhaps overkill.
Okay, I'll remove this addition but we should specify that this tutorial is not the only way to write a Tauri app, maybe just the most common. We should point users to examples or make others guides: one without a bundler and perhaps one with a Rust frontend framework.

JonasKruckenberg: > Mmmmh yeah, "Your First Node.js-based-frontend Tauri App" would be more accurate but perhaps overkill.
> 
> Okay, I'll remove this addition but we should specify that this tutorial is not the only way to write a Tauri app, maybe just the most common. We should point users to examples or make others guides: one without a bundler and perhaps one with a Rust frontend framework.

Sounds good to me!

lorenzolewis:
I think all of these points are being covered in https://github.com/tauri-apps/tauri-docs/pull/590, can we close this one?

JonasKruckenberg:
Yeah

PR #467, opened 2022-02-22
invoke uses cmd instead of command as parameter name

jordanbaron:
[invoke](https://github.com/tauri-apps/tauri/blob/82b7f5195697596a84dbc2c05be29f2ee679e254/tooling/api/src/tauri.ts#L68) now uses `cmd` as the parameter name and not `command`, so I updated the docs to reflect that.

### DIFF
--- a/docs/api/js/modules/shell.md
+++ b/docs/api/js/modules/shell.md
@@ -37,11 +37,11 @@ You can change that regex by changing the boolean value to a string, e.g. `open:
 ### Restricting access to the [[`Command`]] APIs
 
 The `shell` allowlist object has a `scope` field that defines an array of CLIs that can be used.
-Each CLI is a configuration object `{ name: string, command: string, sidecar?: bool, args?: boolean | Arg[] }`.
+Each CLI is a configuration object `{ name: string, cmd: string, sidecar?: bool, args?: boolean | Arg[] }`.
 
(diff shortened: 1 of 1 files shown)


FabianLars:
This has nothing to do with invoke, but the change is correct, it is indeed called `cmd`. Thank you ❤️ 

Also the js api docs are generated from the code comments over at the main repo, so i'd appreciate it if you can open a similar PR at https://github.com/tauri-apps/tauri (otherwise it get overwritten on the next release).

jordanbaron:
> This has nothing to do with invoke, but the change is correct, it is indeed called `cmd`. Thank you ❤️
> 
> Also the js api docs are generated from the code comments over at the main repo, so i'd appreciate it if you can open a similar PR at https://github.com/tauri-apps/tauri (otherwise it get overwritten on the next release).

Got it! Sorry, I thought it would have something to do with `invoke`, I’m not that good at navigating big repos yet. Just so I know, what does it have to do with?

jordanbaron:
I will also make a PR for the main tauri repo

FabianLars:
> what does it have to do with?

Basically it's just this line here: <https://github.com/tauri-apps/tauri/blob/dev/core/tauri-utils/src/config.rs#L894> 🤷

FabianLars (approved):
thx again

PR #474, opened 2022-02-25
fix: remove dollar signs from cmdline instructions, closes #454

### LINKED ISSUE #454: bash: $: command not found...
The shell/bash code snippets shouldn’t start with the prompt sign `$`, especially if there is this super practical <kbd>Copy</kbd> button.

```bash
$ sudo dnf install […]
```
To be replaced by:
```bash
sudo dnf install […]
```

This prompt may confuse beginners, and is less and less _everywhere_. 

Found in the [Getting Started page](https://github.com/tauri-apps/tauri-docs/blob/dev/docs/getting-started/setting-up-linux.md), might be found elsewhere.

I’m ready to do a merge request if you agree with the purpose of this issue.

FabianLars:
For reasons see https://github.com/tauri-apps/tauri-docs/issues/454

I'm not 100% happy _everywhere_ but it's still better than before imo.

Edit: Also, vs-code's format-on-save makes it harder to diff again :))

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -13,32 +13,33 @@ A few style guidelines to help us keep a common style across documents:
 1. Use [Reference-style Links]. This makes reading the raw markdown and updating links easier.
 2. Prefer plain markdown over HTML/MDX. This makes reading the markdown easier.
 3. Use present tense rather than future tense. This makes the text easier to comprehend.
-    Make an exception for planned or obsolete features, using future or past tense as appropriate.
+   Make an exception for planned or obsolete features, using future or past tense as appropriate.
 4. Use active voice rather than passive voice. This makes the text more engaging and easier to comprehend.
 
 ## Installation
 
 ```
-$ yarn
+yarn
 ```
 
 ## Local Development
 
 ```
-$ yarn start
+yarn start
 ```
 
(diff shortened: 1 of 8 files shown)


JonasKruckenberg (approved):

JonasKruckenberg:
I like that you updated the todo comment 😂

JonasKruckenberg:
Actually, I'd like to add these back, but maybe in a way where the dollar sign is ignored for copy-paste. 
Turns out the dollar sign is kinds important for syntax highlighting. But don't worry 👍🏻

FabianLars:
Yeah same, but with docusaurus it looks like we need a) some weird css or b) a custom code block thingy and i wanted to get this PR out of the way for the icons&deps doc update.

FabianLars:
But i don't see what you mean with the syntax highlighting thingy.
If i understood you correctly, then the actual problem is that we use `sh` instead of `bash` on the code blocks (only the latter seems to have highlighting).

JonasKruckenberg:
Aaaactually, I learned from the rust book, that `console` as the code block lang gives you much nicer, more consistent syntax highlighting

FabianLars:
doesn't seem to be supported by docusaurus / prism. So we may have to write our own prism lang definition for that one :/

PR #476, opened 2022-02-25
Fix Formatting

Layendan:
Fixed the markdown formatting for the openWith? parameter options for the open function.


JonasKruckenberg (dismissed):

JonasKruckenberg:
These files are gebersted from the sources at tauri-apps/tauri. Can you update the doc comment directly please?

PR #485, opened 2022-03-06
use the same cargo features for AST generation as for docs.rs

FabianLars:


FabianLars:
nvm then. Need to wait till i have access to my linux machine 🤷

amrbashir (approved):

PR #490, opened 2022-03-08
Fixed none-working code samples.

Hupka:


FabianLars (approved):

docs/guides/plugin.md
```diff
@@ -51,15 +51,19 @@ Using the `tauri::plugin::Builder` you can define plugins similar to how you def
 
 ```rust
 use tauri::plugin::{Builder, TauriPlugin};
```
FabianLars: Can you add the `Runtime` import too?

Hupka: On it!

FabianLars:
Last thing, can you revert to 2-spaces indentation? (i don't like it either, but it needs to be consistent)

PR #496, opened 2022-03-12
Revert "fix docs "macOSPrivateApi" case sensitive error"

JonasKruckenberg:
Reverts tauri-apps/tauri-docs#495

My bad, this is actually the wrong place to fix this, both files are autogenerated from code.
Gonna make myself some coffee now 🙂


FabianLars:
merging it was alright, this way we don't have to wait for the next release. it still needs to be fixed in `tauri` additionally tho :shrug:

PR #497, opened 2022-03-12
Fixed a compile time error at documentation code

whizsid:
I got 'no method named `as_str` found for reference `&str` in the current scope' error


FabianLars (approved):
thank you

PR #498, opened 2022-03-13
fix: respect prefers-color-scheme

FabianLars:


JonasKruckenberg:
Hmmmm, so the default Tauri blue with white background is the polar opposite of accessible 😅

Edit: I know this has nothing to do with your PR. But when respecting the color-scheme even more people will see this by default 🤔

PR #505, opened 2022-03-16
fix: cert signing steps

SeanKilleen:
Hi all, 

I think in Tauri's documentation on signing, it is referring to Environment variables that do not match the name of the environment variables that are created by the process.

The scripts reference  `WINDOWS_PFX` and `WINDOWS_PFX_PASSWORD` when the environment variables set above in the action are `WINDOWS_CERTIFICATE` and `WINDOWS_CERTIFICATE_PASSWORD`. 

Here, I am assuming the environment variable names are the ones you want to go with, and am correcting the variable usage for consistency.


FabianLars (approved):
This indeed sounds more logical, thank you!

PR #507, opened 2022-03-17
fix: backport new sponsor design

FabianLars:
We really need more extensive prettier configs and whatever else we can configure. My vscode configs are pretty much the default settings put it still heavily pollutes the PRs every time.


nothingismagick (approved):
Lgtm

PR #510, opened 2022-03-21
Match APPLE_SIGNING_IDENTITY secret name to env variable name

SeanKilleen:
Noticed this while reading the docs and using the example. This is once instance where the env variable name doesn't seem to match the secret name. It was unintuitive to me so submitting this PR to hopefully save others some time.


FabianLars (approved):
thank you!

PR #511, opened 2022-03-21
fix: rename `__TAURI__._invoke` to `__TAURI_INVOKE__`.

mikoto2000:
Renamed `__TAURI__._invoke` to `__TAURI_INVOKE__` in [tauri-app/tauri #2417](https://github.com/tauri-apps/tauri/pull/2417), but not reflected in the document.

### DIFF
--- a/docs/guides/command.md
+++ b/docs/guides/command.md
@@ -32,7 +32,7 @@ Now, you can invoke the command from your JS code:
 // With the Tauri API npm package:
 import { invoke } from '@tauri-apps/api/tauri'
 // With the Tauri global script, enabled when `tauri.conf.json > build > withGlobalTauri` is set to true:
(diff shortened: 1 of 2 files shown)


FabianLars:
Thank you for contributing, unfortunately this change is not quite correct.
`__TAURI_INVOKE_` is primarily meant for _internal_ usage (each of our api functions use it internally too).
`__TAURI__.invoke` still works (the actual npm package equivalent is `__TAURI__.tauri.invoke` btw), if it doesn't then you probably didn't enable `withGlobalTauri` in your tauri.conf.json.

mikoto2000:
@FabianLars 

Thank you for your comment.
I was able to confirm that `__TAURI__.invoke` works. (Failed to save `tauri.conf.json` ...)

PR #521, opened 2022-03-26
feat(roadmap): update finished milestones

lucasfernog:


FabianLars (dismissed):

src/pages/index.js
```diff
@@ -319,17 +319,17 @@ const Roadmap = ({ siteConfig }) => {
         message: '3rd party security audit of core libraries.',
       }),
       icon: 'ti-magnet',
-      color: COLORS.blue,
+      color: COLORS.green,
       targetQuarter: 'Q1 2022',
     },
     {
-      label: 'Iframe with native API',
+      label: 'Isolation pattern',
```
JonasKruckenberg, line 326: Can we rebrand this to "Context Isolation"? 
I know this is a bigger change now, but this is really not a pattern that you adhere to during development, but a feature that can be enabled or disabled.

"Context Isolation" is also much easier to explain to people and users that come from electron will have an easier time adjusting since we name similar concepts the same.

chippers: The pattern idea is from the old recipes, where patterns are things you enable that affect your development environment (in isolation case, an extra application).  There's plans to add more patterns

nothingismagick (approved):

PR #522, opened 2022-03-27
chore: replace from init_env.sh to setup.sh in contributor-guide

keiya01:
init_env.sh was replaced in https://github.com/tauri-apps/tauri/pull/1189 .


FabianLars (approved):
The whole paragraph is wrong and we need to rewrite the whole page anyway😅 

I'm still going to merge this tho, so it's at least somewhat correct. Thanks for your contribution.

PR #538, opened 2022-04-01
Note that Node is optional for Tauri

thadguidry:


lucasfernog:
Thanks for the suggestion, I'm opening #557 to handle other cases of it.

PR #544, opened 2022-04-17
Grammar Mistakes Resolved : Distribution Section

Anishpras:


docs/distribution/sign-macos.md
```diff
@@ -17,7 +17,7 @@ If you are not utilizing GitHub Actions to perform builds of OSX DMGs, you will
 
 ## tl;dr
 
-The Tauri code signing and notarization process is configured through the following environment variables:
+The Tauri code signing and notarization process are configured through the following environment variables:
```
nothingismagick:
```suggestion
The Tauri code signing and notarization processes are configured through the following environment variables:
```

docs/distribution/windows.md
```diff
@@ -18,7 +18,7 @@ By default, the Tauri installer downloads and installs the Webview2 Runtime if i
 You can remove the Webview2 Runtime download check from the installer by setting [tauri.bundle.windows.wix.skipWebviewInstall] to `true`. Your application WON'T work if the user does not have the runtime installed.
 :::
 
-Using the runtime provided by the system is great for security as the webview vulnerability patches are managed by Windows. If you want to control the Webview2 distribution on each of your applications, either to manage the release patches yourself or distribute applications on environments where internet connection might not be available. In that case, Tauri can bundle the runtime files for you.
+Using the runtime provided by the system is great for security as the webview vulnerability patches are managed by Windows. If you want to control the Webview2 distribution on each of your applications, either manage the release patches yourself or distribute applications in environments where internet connection might not be available. In that case, Tauri can bundle the runtime files for you.
```
nothingismagick:
```suggestion
Using the runtime provided by the system is great for security as the webview vulnerability patches are managed by Windows. You may want to control the Webview2 distribution on each of your applications by managing the release patches yourself, for example when distributing applications in environments where an internet connection might not be available. In that case, Tauri can bundle the runtime files for you.
```

nothingismagick (changes requested):
Thanks for the suggestions. See my comments.

docs/distribution/windows.md
FabianLars: I actually like the old version the most, it's just the most "accurate" one (minus the on->in change, that's correct).
To explain what the docs mean with other words: We offer a way to distribute a _fixed_ version of WebView2 (aka. a webview runtime that doesn't auto update and is only used by that single app). For this we had 2 main use cases in mind:
1) The developer wants to have control over the exact webview version the enduser has.
2) The developer wants to distribute their app in offline environments (imo the offline installer would be the better approach but there were issues implementing that).

So yeah, i think the current version (live on tauri.studio) reflects that best, but i think it needed a second "to", so it would look similar to this in the end:
> If you want to control the WebView2 distribution on each of your applications, either to manage the release patches yourself or to distribute applications in environments where an internet connection might not be available, Tauri can bundle the runtime files for you.

Tbf the "on each of your applications" part puts me off, but i can't think of a better phrasing.

lorenzolewis (dismissed):

PR #557, opened 2022-04-27
feat(getting-started): note that nodejs is optional

lucasfernog:


JonasKruckenberg (changes requested):
lgtm, if you make those links you added reference-style-links too 👍🏻

PR #559, opened 2022-05-03
Feat/css cleanup

lorenzolewis:
- Move logic for loading dark header logo from custom CSS to docusaurus config
- Remove unused EthicalAds JS file and references from custom CSS
- Move Meilisearch CSS to SearchBar component module


FabianLars:
the searchbar text alignment is a little bit broken, other than that lgtm

grafik

lorenzolewis:
Thanks @FabianLars :D

Turns out that wasn't as easy as I thought since the class names aren't directly targeted in the module. Doing a bit of a workaround here so the styles are still being treated as "global", just keeping them in a slightly more organised location.

FabianLars (approved):

PR #563, opened 2022-05-04
Feat/benchmarks move to docs

lorenzolewis:
This started as a pet peeve of mine that the benchmark page didn't have the sidebars like the docs, but then turned into a full-blown logic re-write:

- Migrate content from `pages/benchmarks.js` to `docs/benchmarks.md`
- Create `BenchmarkChart` component
- Combine logic from `pages/benchmarks.js` and `utils/benchmarks.js` into the above component
- Add Netlify redirect since changing from a page to a doc changes the url
- Update `sidebar.js` to now point to benchmark doc
- Adapt bulleted list of tests to a table at the top of the benchmark page
- Minor grammar fixes on benchmarks page

Remaining open items (help is much appreciated):

- [ ] Dynamically update `options.theme.mode` based on Docusaurus `useColorMode.isDarkMode` property in BenchmarkChart component
- [ ] "Binary size" raw data looks like there are duplicates specifically for `wry_rlib`. Currently filtering this out and not rendering

Changes:

- [ ] I've removed the links to the GitHub commits for each datapoint as they were finicky before and I don't see the value to them.
- [ ] Also removed the "header" containing the commit hash for the tooltip as it wasn't that most informative IMHO


lorenzolewis:
Pending https://github.com/tauri-apps/benchmark_results/pull/2 to be merged to fix date format issue in Wry data

FabianLars (approved):
We really need that new data 😂 other than that lgtm

PR #567, opened 2022-05-07
Update Release Candidate Announcement Banner Style

hslee2008:
# Problem - 1
desktop-margin

In Desktop and Mobile, the banner has margin of 10px on the top which makes the banner look squashed.

# Problem - 2

mobile-link

In Mobile, the links goes to a new line and it is hard to understand at first sight.

# Solution
For problem 1, remove `margin: 10px`.
For problem 2, add id to the link and use `display: block` when screen width is less than 500px.

# Fixed
desktop-margin-fixed
mobile-link-fixed


lorenzolewis (approved):
Awesome, thanks a million!

PR #569, opened 2022-05-07
Revise Architecture

lorenzolewis:
Overall refinements to the architecture page on the site.

- Remove a duplicated Architecture page
- Add redirect for removed dup page
- Add links to guides and docs where relevant
- Remove the "Workflows" section and replace with "Next Steps" linking to guides covering topics in previous Workflows section
- Capitalisation fixes


lorenzolewis:
Converting back to a draft, planning to put in some sort of a diagram

lorenzolewis:
Added in Mermaid for architecture. Note that the correct syntax is in the diagram to properly flow the diagram but there is currently a bug in Mermaid that makes flowing objects with connections not work. I couldn't find a workaround for this currently.

FabianLars (approved):

PR #575, opened 2022-05-08
Remove unused code in search bar

hslee2008:
useDocusaurusContext is not used in search bar component.
Removing it does not effect the search bar.


FabianLars (approved):
thanks

PR #577, opened 2022-05-08
Feat/lottie-player Dependency

lorenzolewis:
Previously the JS for the Lottie animation on the index page had the raw player JS in the project. This removes that and instead relies on the NPM package for Lottie Player.


FabianLars:
Side note: An automatic bundle size benchmark on PRs would be nice 😅

FabianLars:
Hmmm, can we make it a little bit smaller again? Or do you prefer it that like this?

Oh and is it trivial to prevent the layout shift?

lorenzolewis:
I want to remove Lottie completely for the v1.0 home page ideally

FabianLars:
Okay then let's get this out of the way for now

FabianLars (approved):

PR #579, opened 2022-05-08
Feat/auto generate sidebar

lorenzolewis:
This is work in 2 parts:
- Prep work for the updated docs structure
- Remove a lot of hard-coding of sidebar values in `sidebar.js`

## Changes
- Change "Docs" to be called "Guides" and update link in both navbar and footer
- Rename the "guides" section to "examples" to avoid naming conflict (this is aligned with the move to the updated docs structure)
- Move any subfolders that in `docs` to the `guides` folder to allow dynamic sidebar generation
- Update broken links through documents as a result of this move
- Add `_category_.yml` files to folders to set sidebar label and position (https://docusaurus.io/docs/sidebar/autogenerated#category-item-metadata)
- Add front matter to markdown documents set sidebar label and position (https://docusaurus.io/docs/sidebar/autogenerated#doc-item-metadata)
- Move Getting Started platform-specific guides to a "setup" subdirectory
- Revise Netlify redirects

## Note on Redirects
A lot of old redirections were removed and updated to now land any visitors going to `docs/getting-started`, `docs/development`, etc. to just go to a 404. Although it's technically possible to redirect them to the correct link, the new docs structure does not cleanly map to the current structure and would be broken in the future soon anyway.

In the great words of many: Best to break it now before the launch


JonasKruckenberg:
> Note on Redirects

Question about netlify, isbit possible to see the number of people visiting specific pages? It'd be very helpful in the future to know if we can remove redirects, if they reach low enough traffic for example.

lorenzolewis:
I haven't interacted with it at all before, so not sure on that.

FabianLars:
Is there no way to make this work without moving it to docs/guides/ ?

Or maybe we can get rid of the /docs/ prefix? (similar to how vitejs does it for example), because:
- docs/guides feels rendundant
- docs/about is super weird, same for community

That said maybe we want to pull config out of the api submenu too?

JonasKruckenberg:
> Or maybe we can get rid of the /docs/ prefix? (similar to how vitejs does it for example), because:
> 
> - docs/guides feels rendundant
> 
> - docs/about is super weird, same for community

Agreed, but idk if this is easy to change with how docusaurus works 🤔


> That said maybe we want to pull config out of the api submenu too? 

I like that 👍🏻

lorenzolewis:
Give me a few min, just found a config for it

lorenzolewis:
How does that look on the latest preview now @FabianLars @JonasKruckenberg

JonasKruckenberg:
> How does that look on the latest preview now @FabianLars @JonasKruckenberg 

It's good 👍🏻

lorenzolewis:
Alright, an approval and then it should be good to merge

FabianLars:
Do we want to pull the config.md out of the api page? And if so, in this pr or another one?

lorenzolewis:
We have to update scripts if we do that, so would rather do it separately.

FabianLars:
Alrighty. Just the redirect for the benchmarks in netlify.toml left then as far as i can see

lorenzolewis:
Just going to remove it completely

FabianLars (approved):

PR #583, opened 2022-05-10
Feat/update release notes page

lorenzolewis:
Re-themes the Release Notes page and also moves it from a "page" to a "doc"
- Fix async issue that caused page to lag when loading release notes
- Fall back to using Docusaurus Tab component
- Uses `react-markdown` to render the raw markdown data into HTML
- Remove now un-needed `debounce` utility
- Add link to GitHub for each link (can act as a fallback if needed in the future)


netlify.toml
```diff
@@ -13,3 +13,7 @@ to = "https://docs.rs/tauri/1.0.0-rc.0/tauri/"
 [[redirects]]
 from = "/partners"
 to = "/#sponsors"
+
+[[redirects]]
+from = "/release-notes"
+to = "/docs/release-notes"
```
FabianLars: That's not correct (anymore)

package.json
```diff
@@ -26,7 +26,7 @@
     "react-apexcharts": "1.4.0",
     "react-content-loader": "6.2.0",
     "react-dom": "17.0.2",
-    "react-markdown": "8.0.3",
+    "react-markdown": "^8.0.3",
```
FabianLars: Doesn't matter too much, but renovate will revert that one (same for the lock file)

FabianLars (changes requested):
all in all pretty good imo.
My only problem are the dropdowns. Are they generally modifiable? They don't look clickable & title-ish enough imo.
If we can't really modify them, can we at least change the cursor on hover?

lorenzolewis:
@FabianLars how about now? Set the titles to h3 and added a bit of CSS to make it more "link-like"

lorenzolewis:
Also, does your browser at all lag when data is loading on this page? There was an issue where it kept having performance issues but I updated Docusaurus and I'm not running into the issue now, so just want to confirm.

FabianLars:
>how about now?

I like it :+1: 

> performance

Wayyy better than on the live site :+1:

JonasKruckenberg:
@lorenzolewis did you decide that the search (filter) wasn't worth it? I just noticed it's gone in the preview
(I never really understood/used it much so I'm fine with removing it)

Edit: It's also soooooo laggy 😂

FabianLars:
didn't even know that there was a search :joy:

lorenzolewis:
@JonasKruckenberg is the live site laggy? or you talking this dev version? If the dev version, then I hand that over to you to help with because #reactland

For the search/filter, figured it wasn't worth it and the current way it was implemented just didn't have a nice behaviour for me imho

JonasKruckenberg:
> @JonasKruckenberg is the live site laggy? or you talking this dev version? If the dev version, then I hand that over to you to help with because #reactland


Sorry I meant the filter functionality of the live site, it's just really laggy. The perf for your release notes themselves is good!

lorenzolewis:
Perfect! @FabianLars just needs your review then I think

FabianLars (approved):

PR #584, opened 2022-05-10
Remove usage, generation, and storage for `version.txt` file

lorenzolewis:
This was originally used as part of the old release notes. Removing as it seems nothing depends on it now (especially since the line generating it has been commented out for a bit).


lorenzolewis:
Closing, going to cover this with the versioning PR

PR #586, opened 2022-05-10
Feat/versioning prep

lorenzolewis:
# General Cleanup
- Remove old `version.txt` file that is no longer being used
- Remove lines from script that generated that file
- Remove `community/ci-cd.md` doc that wasn't referenced anywhere
- Remove `community/contributor-guide` doc and change navbar to point to https://github.com/tauri-apps/tauri/blob/dev/.github/CONTRIBUTING.md instead
- Remove community sidebar since there are no longer any docs here
- Remove unused `splash.png` file

## Versioned URL changes
- Change URLs in all docs to be relevant instead of absolute paths
- Move Release Notes from `docs` folder to `src/pages` so that it doesn't get versioned
- `index.js` changes:
  - Load the current version from docusaurus context to concat to any docs urls as a workaround since Docusaurus doesn't seem to support linking versioned docs from pages when not using the `/docs` path
- `docusaurus.config.js` changes:
  - Remove global `version` variable that wasn't being used anywhere
  - Change navbar links for guides and APIs to doc types so they get the properly versioned url
  - Remove "Docs" section from footer (currently the footer component doesn't support dynamically generating the url for a versioned doc like the navbar component does), replaced with links to upstream crates (Tao and Wry)
  - Add `presets.'@docusaurus/preset-classic'.docs.versions` object to add in the path for "v1" based on the current files
  - Added in commented out code to the same position for post-v1 launch to dynamically write the version number for the latest versioned files


lorenzolewis:
Note to self:
- [x] Write guide on how to version and update docs
- [x] Check any redirects if they need to be updated
- [x] File issues with Docusaurus on...
  - [x] Footer links to docs
  - [x] Linking to versioned doc from pages

FabianLars:
Overall quite happy with that, good job!

- ci-cd.cd -> good riddance 👍 
- It's a little unfortunate that we can't use absolute urls, but doesn't matter too much imo. (unless we want to version them too and batch update them in v2, but urgh...)
- v1/about is probably fine too, we could add redirects for that later if we want
- speaking of redirects we probably should consider adding redirects to v1 for a few weeks? i think search crawlers should follow them, right? Edit: This will also prevent our search from breaking, i don't think i can patch it until tomorrow night
- idea: Add tauri to the upstream crates footer, rename the section to _something_ idk and change the github link in the network section to the tauri-apps account. idk, i think just upstream/tao+wry is a little weird 🤔 
- _i forgot my last point..._

lorenzolewis:
@FabianLars added in those redirects and adjusted the footer. I put in a bit of a utility for generating a url for a docs page so we can use that in the future if we need to link to a doc from a non-versioned page (but will still open an issue with Docusaurus).

I'm going to do a bit more testing tomorrow to make sure nothing breaks when we actually deploy a version (and will write up a guide on how to perform a version).

FabianLars:
Just for my understanding, do we wanna merge this before we release stable?

I'm fine with that and it's probably good for seo stuff to update somewhat in-time. On that note, if we do it we need to add another redirect for `/docs/* -> /v1/:splat` for a few weeks imo (to fix the current mess).

lorenzolewis:
Yes to both of those @FabianLars. Better to get all the links out there now. Just added in that redirect

lorenzolewis:
This is ready for review and merge. I need to put in a few upstream issues but will do that on the side, just have some workarounds in place in the meantime.

FabianLars:
The urls default to v2 and manually changing the url to v1 doesn't work either

lorenzolewis:
Silly me, tested it and then apparently pushed my test, fixed now. Also added in an issue about upstream issue https://github.com/tauri-apps/tauri-docs/issues/589

FabianLars (approved):
lgtm

lorenzolewis:
Yay! Somehow that was easier and more difficult than I expected!

PR #590, opened 2022-05-11
Update guide

lorenzolewis:
# Changes
- Add `mdx-mermaid` dependency to simplify mermaid in markdown files
- Remove custom `Mermaid` component
- Remove unused `SetupDocs` component
- Simplify markdown files to use new mermaid syntax
- Switch to prefix-based sidebar order for categories under "guides" (easier to edit in the filesystem now)
- Revise Getting Started -> Prerequisites doc
- Add Concepts category and docs:
  - Inter-process communication
  - Process model
  - Security
- Remove separate Windows, Linux, and macOS setup pages (now combined into the prerequisites page)
- Add CSS for dynamic image switching based on theme


docs/guides/getting-started/prerequisites.md
```diff
@@ -3,22 +3,204 @@ sidebar_position: 1
...
+
+The bootstrapper script will try to determine the correct architecture and version for your system. Still, if you run into issues - especially with Windows on ARM - you can select the correct standalone installer or even a fixed version.
+
+## Setting Up macOS
+
+To install Rust on macOS, open a terminal and enter the following
+command:
+
+```console
+curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf | sh
```
lorenzolewis: @JonasKruckenberg is it possible to just use the command `curl https://sh.rustup.rs | sh`? Seems better for succinctness

docs/guides/getting-started/prerequisites.md
```diff
@@ -3,22 +3,204 @@ sidebar_position: 1
...
+```console
+xcode-select --install
+```
+
+## Setting Up Linux
+
+To install Rust on Linux, open a terminal and enter the following command:
+
+```console
+curl --proto '=https' --tlsv1.2 https://sh.rustup.rs -sSf | sh
```
lorenzolewis: Same here @JonasKruckenberg, use  `curl https://sh.rustup.rs | sh`?

docs/guides/getting-started/prerequisites.md
```diff
@@ -3,22 +3,204 @@ sidebar_position: 1
...
+rustup update
+```
+
+`rustup` can also be used to uninstall Rust from your machine fully:
+
+```console
+rustup self uninstall
+```
+
+<!-- TODO: Unistall Webview2 -->
```
lorenzolewis: Should we open an issue on the repo for this so we come back to it?

docs/guides/getting-started/prerequisites.md
FabianLars: it's the "official" rustup command (https://rustup.rs/). i don't know the reasons, but i can't imagine that they doing this for fun 🤔 

We have a similar problem on windows with the webview2 installation on windows7 because it doesn't use tls1.2 by default there, but getting the webview2 installer requires that. Maybe it's something similar here

Edit: and sSf is some error log thingy idk

docs/guides/getting-started/prerequisites.md
FabianLars: Just remove that part lol. webview2 isn't reallyyy supposed to be uninstalled anyway (and you can't even do that on win11). Also it's a "normal" uninstallation process every windows user should know.

docs/guides/getting-started/prerequisites.md
lorenzolewis: Okay, that makes sense to keep as-is

lorenzolewis:
@FabianLars would be nice to get a review from you on this since the docs were created largely by @JonasKruckenberg. Goal is to get the docs migrated over to the site ASAP so we can get urls fairly stable, then revise the content incrementally.

FabianLars:
Okay so just like in the last PR i need to mention that i absolutely hate the numbered folders, but if there's no (sane) way around it then whatever 🤷 

_gonna start reviewing asap_

lorenzolewis:
I don't have a strong opinion on numbered folders, it just makes it quite a bit easier to make these changes right now. If it's okay to have numbered now and then go back to non-numbered after content is migrated I'd like that (won't affect urls).

docs/guides/01-getting-started/prerequisites.md
```diff
@@ -0,0 +1,204 @@
...
+
+The first step is to install [Rust] and system dependencies. Keep in mind that this setup is only needed for _developing Tauri apps_. Your end-users are not required to do any of this.
+
+## Setting Up Windows
+
+:::info
+For those using the Windows Subsystem for Linux (WSL), please refer to our [Linux specific instructions](#setting-up-linux) instead.
+:::
+
+On Windows, go to [https://www.rust-lang.org/tools/install][install rust] to install `rustup` (the Rust installer). You also need to install Microsoft Visual Studio C++ build tools. The easiest way is to install [Build Tools for Visual Studio 2019]. When asked which workloads to install, ensure "C++ build tools" and the Windows 10 SDK are selected.
```
FabianLars: Multiple people asked for an image for that one. Is it possible to integrate it _nicely_ here or would that look weird?

docs/guides/01-getting-started/prerequisites.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Install WebView2
+
+:::note
+WebView2 is pre-installed in Windows 11
+:::
+
+Tauri heavily depends on WebView2 to render web content on Windows, therefore you must have WebView2 installed. The easiest way is to download and run the Evergreen Bootstrapper from the [official website][download webview2].
+
+The bootstrapper script will try to determine the correct architecture and version for your system. Still, if you run into issues - especially with Windows on ARM - you can select the correct standalone installer or even a fixed version.
```
FabianLars: i'd rather not mention the fixed version here, don't want to give any ideas to people who don't know much about webview2 and its update process.

docs/guides/01-getting-started/prerequisites.md
```diff
@@ -0,0 +1,204 @@
...
+    wget \
+    libappindicator-gtk3 \
+    librsvg2-devel
+sudo dnf group install "C Development Tools and Libraries"
+```
+
+  </TabItem>
+</Tabs>
+
+### Windows Subsystem for Linux (WSL)
```
FabianLars: I'm at a point were i'd love to just remove the wsl section. Or only "support" wsl with wslg _on windows 11_. At the moment it causes nothing but problems we can't even help debugging since none of us uses wsl anymore... (not that it matters, we'd still wouldn't be able to help with most of the setup problems even if we'd all use wsl...)

I'd like to hear some opinions on that one.

FabianLars: Oh and now that i mention windows 11, our wsl docs don't account for it at all.

docs/guides/02-concepts/inter-process-communication.md
```diff
@@ -0,0 +1,58 @@
...
+```
+
+<figcaption>Figure 1-2: An event sent from the Core to the Frontend.</figcaption>
+</figure>
+
+## Commands
+
+Tauri also provides a [foreign function interface]-like abstraction on top IPC messages[^1]. The primary API, `invoke`, is similar to the browsers `fetch` API and allows the Frontend to invoke rust functions, pass arguments, and receive data.
+
+Because this mechanism uses the [JSON-RPC] protocol under the hood to serialize requests and responses, all arguments and return data must be serializable to JSON.
```
FabianLars: technically this is incorrect, isn't it? JSON-RPC is a specification and afaik we're not following that spec. I don't reallyyy have our ipc implementation in my head rn tho 🤔

docs/guides/02-concepts/security.md
```diff
@@ -0,0 +1,81 @@
+---
```
FabianLars: you probably have an idea already, but with the new-ish security page we need a good way to (re-)integrate the second half of the old one (starting here <https://deploy-preview-590--tauri.netlify.app/v1/guides/development/security#tauri-api>) _somewhere_ (it would be good if they'd be easily accessible in/from the corresponding js module docs).

docs/guides/01-getting-started/prerequisites.md
JonasKruckenberg: I'll try!

docs/guides/02-concepts/inter-process-communication.md
JonasKruckenberg: It's not a 100% compatible but it's very close to json-rpc.

docs/guides/02-concepts/security.md
lorenzolewis: There's a yet-to-be migrated sub-page for security:
https://github.com/JonasKruckenberg/tauri-docs-wip/blob/main/src/background/security/context-isolation.md
Should we break out https://github.com/tauri-apps/tauri-docs/blob/dev/docs/guides/development/security.md#tauri-api into a separate page and add it under `concepts/security`?

If there are any things we want to highlight outside of concepts (imagining if someone skips concepts section), then I think we should add that directly in the API documentation

docs/guides/01-getting-started/prerequisites.md
lorenzolewis: Okay if we circle back on this and track it here: https://github.com/tauri-apps/tauri-docs/issues/593 later?

docs/guides/01-getting-started/prerequisites.md
lorenzolewis: Mentioning Windows 11 support here: https://github.com/tauri-apps/tauri-docs/issues/593. If what's here is accurate then I think we leave it as-is and then circle back on it?

docs/guides/01-getting-started/prerequisites.md
FabianLars: @lorenzolewis i didn't mean to integrate it _right now_, i wanted to make sure that we have that in mind and make sure that we _can_ integrate it later.

If with "I'll try" Jonas means he wants to do it right now here are 2 screenshots :P
Visual Studio Installer Light
Visual Studio Installer Dark

lorenzolewis: Ah gotcha. Well I'll still use that issue to track all of these brilliant ideas so they don't get lost ;)

docs/guides/01-getting-started/prerequisites.md
FabianLars: So the problem is, it's only accurate _in theory_. Like it's really just the luck of the draw (edit: is that how you say that? 😂) if it works correctly or not; x servers on windows are just super hacky 🤷 

People reporting issues are somewhat common the last couple of days/weeks and it's never something tauri specific but just x server weirdness, i even recognize one of these issues from back when i used wsl a few years back (i wasn't able to figure it out back then and completely dropped wsl in my setup lol).

lorenzolewis: Do we have enough info to write it _correctly_ now?

Mentioning WSL vs WSL2 vs WSLG, let's not forget that potential enterprise users might not have the luxury of upgrading to win 11 if that's a consideration ;)

FabianLars: i don't think enterprise users (should) use these hacky x servers in the first place lol.

> Do we have enough info to write it correctly now?

Yes and no. In theory it's only missing one tiny thing (`LIBGL_ALWAYS_INDIRECT=1`) to match the broader range of wsl setup docs and it's one of multiple causes of the most common error message.
Buuut x servers on windows, especially together with webkit(!) is so unstable that these instructions are, more often than not, not enough.
It's also really dependent on the system itself and stuff, so it's super hard to debug it from our side (to make matters worse there are system setups where it straight up won't work no matter what you do 🤷).

**Again, my main problem here is that by providing docs for it we're suggesting that a) it's (well) supported and b) that we can provide help on issues.** Which is both not the case, which can be seen by the non-resolved issues (on discord) 😅 

P.S. I'm mostly afraid here because i'm like the only Windows user in the team😂 

P.P.S There's still hope that WSLg can make it into Windows 10 and i'm still praying everyday that MS follows through with it.

docs/guides/02-concepts/security.md
FabianLars: ~~yeah similar to how it's done for the context isolation docs in the wip docs sounds reasonable to me.~~ Edit: I think this contradicts with the statement below.

> then I think we should add that directly in the API documentation

Yeah, most of the (api security) documentation is actually more like a tutorial on how to do stuff than explaining concepts. It's quite common that i send a link to that page to people who want to know how to configure scopes correctly, so that's kindaaa weird.

docs/guides/02-concepts/inter-process-communication.md
FabianLars: Do we want to rephrase it to `this mechanism uses a JSON-RPC like protocol` instead? :P (i'm fine with leaving it as-is but maybe i can prevent a follow up pr with that question)

docs/guides/01-getting-started/prerequisites.md
FabianLars: P.P.P.S. Since i don't run _any_ VMs on Windows anymore i may try out wsl again and see if i stumble upon any common issues.

docs/guides/01-getting-started/prerequisites.md
JonasKruckenberg: Agree

docs/guides/01-getting-started/prerequisites.md
JonasKruckenberg: I agree with @FabianLars this will cause us more problems that it is worth. Remember that this is the doc people will read. They will be excited about wsl and then quickly discouraged if things turn out badly

docs/guides/02-concepts/inter-process-communication.md
JonasKruckenberg: Let me check how json-rpc compliant we actually are!

docs/guides/01-getting-started/prerequisites.md
```diff
@@ -0,0 +1,212 @@
...
+
+The first step is to install [Rust] and system dependencies. Keep in mind that this setup is only needed for _developing Tauri apps_. Your end-users are not required to do any of this.
+
+## Setting Up Windows
+
+:::info
+For those using the Windows Subsystem for Linux (WSL), please refer to our [Linux specific instructions](#setting-up-linux) instead.
+:::
+
+On Windows, go to [https://www.rust-lang.org/tools/install][install rust] to install `rustup` (the Rust installer). You also need to install Microsoft Visual Studio C++ build tools. The easiest way is to install [Build Tools for Visual Studio 2019]. When asked which workloads to install, ensure "C++ build tools" and the Windows 10 SDK are selected.
```
FabianLars: Since the link will download version 2022 now.
```suggestion
On Windows, go to [https://www.rust-lang.org/tools/install][install rust] to install `rustup` (the Rust installer). You also need to install Microsoft Visual Studio C++ build tools. The easiest way is to install [Build Tools for Visual Studio 2022]. When asked which workloads to install, ensure "C++ build tools" and the Windows 10 SDK are selected.
```

docs/guides/02-concepts/inter-process-communication.md
JonasKruckenberg: Alright I have no idea where that code ended up 🤷🏻‍♂️😅 I'll just reword it

docs/guides/01-getting-started/prerequisites.md
lorenzolewis: Good with whatever way you guys decide. If we keep it then we can always add one of those warning call-outs saying it isn't officially supported

docs/guides/01-getting-started/prerequisites.md
```diff
@@ -0,0 +1,212 @@
...
+
+The first step is to install [Rust] and system dependencies. Keep in mind that this setup is only needed for _developing Tauri apps_. Your end-users are not required to do any of this.
+
+## Setting Up Windows
+
+:::info
+For those using the Windows Subsystem for Linux (WSL), please refer to our [Linux specific instructions](#setting-up-linux) instead.
+:::
+
+On Windows, go to [https://www.rust-lang.org/tools/install][install rust] to install `rustup` (the Rust installer). You also need to install Microsoft Visual Studio C++ build tools. The easiest way is to install [Build Tools for Visual Studio 2022]. When asked which workloads to install, ensure "C++ build tools" and the Windows 10 SDK are selected.
```
FabianLars: Hmm i think we may have to swap that around. I think rustup can default to the wrong toolchain if it can't find the build tools, gonna check that real quick.

On that note, it makes sense to swap it around for every OS. It won't look/read as nice but it makes more sense imo.

FabianLars: Okay it still defaults to the correct toolchain but will show an error screen beforehand (you can dismiss it with "y" and can install it like normal)

FabianLars: Turns out it didn't default to the msvc toolchain for others on discord who didn't have ms build tools installed so we really need to swap them around.

JonasKruckenberg: Alright 👍🏻

docs/guides/01-getting-started/prerequisites.md
lorenzolewis: So I'm hearing remove WSL, so the whole section under https://deploy-preview-590--tauri.netlify.app/v1/guides/getting-started/prerequisites#windows-subsystem-for-linux-wsl 
Do we want to keep the callout here on referring to linux instructions if using WSL https://deploy-preview-590--tauri.netlify.app/v1/guides/getting-started/prerequisites#setting-up-windows?

docs/guides/02-concepts/security.md
lorenzolewis: I meant like the tauri rust/js (in-code) API docs. Keep everything in background/security like it is now, but if there are any very important bits, those should ALSO be in our API docs where relevant.

docs/guides/01-getting-started/prerequisites.md
JonasKruckenberg: nope, just make all the wsl go away 😂

lorenzolewis: Will update this now ;)

lorenzolewis: And done

docs/guides/02-concepts/inter-process-communication.md
lorenzolewis: I adapted this wording now

docs/guides/01-getting-started/prerequisites.md
lorenzolewis: Updated this now and improved the headings, give a look and let me know what you think

docs/guides/01-getting-started/prerequisites.md
```diff
@@ -0,0 +1,214 @@
...
+
+# Prerequisites
+
+:::tip Quick Overview
+Check out [Tauri in 100 Seconds](https://www.youtube.com/watch?v=-X8evddpu7M) by [Fireship](https://www.youtube.com/channel/UCsBjURrPoezykLs9EqgamOA) on YouTube for a quick introduction to Tauri
+
+<div className="videowrapper">
+  <iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/-X8evddpu7M" frameborder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+</div>
+:::
```
JonasKruckenberg, line 17: Do you think this makes the most sense here? 🤔 I would have put it on the main page below the feature cards maybe

lorenzolewis: I don't recommend putting it on index since it isn't our content. Once we have guides built out, I want to make a quick "landing" page when the user clicks "Guides" in the navbar that leads to the respective sections. I think moving the video to that page once we have it built out would be better. But having it here for now at least gives it visibility

JonasKruckenberg: Hmm yeah alright 🤔

lorenzolewis:
Do we want to add redirects in for anything?

lorenzolewis:
I added redirect from /examples/ to /features/ and then a general one for getting-started/setup/* to getting-started/prerequisites

JonasKruckenberg (approved):

Issue #372, opened 2022-01-24
Redesign Search


lorenzolewis:
I'd like to move the site to using Docusaurus's Algolia search UI. Docusaurus uses it for their site so you can see the functionality of it there. The current backend would still stay with Meilisearch and not looking to move away.

## Issues to fix

- [x] Mobile bar covers logo https://github.com/tauri-apps/tauri-docs/issues/574
- [x] Mobile results are cut off of edge
<img width="367" alt="Screenshot 2022-05-13 at 21 29 16" src="https://user-images.githubusercontent.com/15347255/168384834-92ac9308-a032-435e-923f-ccfad6b3403b.png">

- [x] Align logo with baseline of text
<img width="232" alt="Screenshot 2022-05-13 at 21 29 44" src="https://user-images.githubusercontent.com/15347255/168384916-3c66ed4e-a7dc-48e0-8899-ff6c4b34d8ac.png">

- [x] Sizes of dark/light mode are different
<img width="263" alt="Screenshot 2022-05-13 at 21 30 15" src="https://user-images.githubusercontent.com/15347255/168384988-5df6a89c-c6aa-41d4-bb73-8730aee58843.png">
<img width="406" alt="Screenshot 2022-05-13 at 21 30 21" src="https://user-images.githubusercontent.com/15347255/168384996-75848f7e-a713-46ca-9ae0-03fe71c5dfae.png">

## Helpful repo files:
- Current searchbar component: https://github.com/tauri-apps/tauri-docs/tree/dev/src/theme/SearchBar

Issue for mobile: https://github.com/tauri-apps/tauri-docs/issues/574

lorenzolewis:
@HyunseungLee-Travis if you're interested in digging into the search bar UI, feel free to give @yankeeinlondon and myself a ping on Discord, he worked on the backend indexing and I'm familiar with the whole site overall

lorenzolewis:
Tagging @OlegGulevskyy as well

lorenzolewis:
After having a chat with @OlegGulevskyy, deciding to not use the AlgoliaSearch UI. Originally thought the UI was something written by Docusaurus but they've just done a wrapper around https://github.com/algolia/docsearch, so it doesn't seem right to take that component. 

Putting https://github.com/tauri-apps/tauri-docs/pull/599 in place in the mean time to at least clean up some of the logic.

lorenzolewis:
Closing this with #600, thanks @HyunseungLee-Travis for the help!

PR #599, opened 2022-05-13
SearchBar Component Cleanup

lorenzolewis:
Want to put this in place as a stop-gap to at least clear up a bit of logic while we continue looking into https://github.com/tauri-apps/tauri-docs/issues/372


FabianLars:
it's alright overall, but it always shows a scrollbar (because of the image?)

lorenzolewis:
Ah didn't notice that (dang hidden scrollbars), let me see...

lorenzolewis:
They have some weird styling choices with that dependency... I love seeing how the dark theme input is wider than the light theme
Screenshot 2022-05-13 at 20 13 03
Screenshot 2022-05-13 at 20 12 58

FabianLars (approved):

FabianLars:
yeah let's just not talk about it 😅

PR #604, opened 2022-05-15
Feat/config

### LINKED ISSUE #257: Error in CLI json
In the en and [fr](https://github.com/tauri-apps/tauri-docs/blob/f488410eb02991522f38e45e8b1c9d4fd7c9f1e3/docs/fr/usage/guides/cli.md) markdowns, there is an error in heading "Flag arguments":
```
{
  "args": [
    "name": "verbose",
    "short": "v",
    "multipleOccurrences": true
  ]
}
```
should be: 
```
{
  "args": [{
    "name": "verbose",
    "short": "v",
    "multipleOccurrences": true
  }]
}
```

### LINKED ISSUE #468: Inline docs/syntax for allowlist scopes in config docs
Example: <https://tauri.studio/docs/api/config#tauri.allowlist.shell.scope> - how should ~~a user~~ Jonas know what `ShellAllowlistScope` is?

lorenzolewis:
Still a few minor tweaks I need to do, but would like to get a few pairs of eyes on this.

## Items to fix in source:

- [x] "HttpAllowlistScope" description
- [x] Fix link in ShellAllowlistConfig.sidecar description

## Remaining items:

- [x] Link anchor in table
- [x] Required values in properties
- [x] Add value formats
- [x] Maybe revise the example config a bit (include something in the `beforeBuildCommand`, `beforeDevCommand`, etc.)
- [x] Elaborate on "Platform-Specific Configuration" section (would like some help with this)
- [x] Elaborate on 4 main objects listed under "Configuration Structure"
- [x] Check bundle targets
- [x] Remove the `replaceAll` filters at the top after source is fixed
- [x] Format default values a little better
- [x] Add if a value is optional (and a note about this)
- [x] ShellAllowedArgs is nested too deeply
- [x] If a default value is a JS object (DebConfig.files)
- [x] MacConfig, WindowsConfig, SecurityConfig, WindowConfig "any of the following values can be used" troubleshooting
- [x] Is PatternKind, Theme supposed to have content?
- [x] Is UpdaterEndpoint used anywhere?


.scripts/generate-config-doc.js
```diff
@@ -1,149 +1,353 @@
 const fs = require('fs')
+const { type } = require('os')
 const path = require('path')
 const schemaPath = path.join(__dirname, '../../tauri/tooling/cli/schema.json')
-const schema = JSON.parse(fs.readFileSync(schemaPath).toString())
+const schemaString = fs
+  .readFileSync(schemaPath)
+  .toString()
+  .replaceAll('<', '&lt;') // Fixes opening HTML brackets
+  .replaceAll('>', '&gt;') // Fixes closing HTML brackets
```
FabianLars: These ones actually break most of the usages since they are in "backtick blocks". Tricky to to handle both cases here 🤔 
Not counting links, only 2 occurrences without backticks.

FabianLars (changes requested):

FabianLars:
I feel like the headers of the allowlist objects are too small, especially being next to the big tables. They're too easy to miss rn.

FabianLars:
The Scope objects are really weird too, but i don't really know how to integrate them better without flattening them into the respective allowlist config table (actually i'd like that, but _no_ idea how to do it.

Edit: Or maybe they should get separate pages like in the js docs 🤔

.scripts/generate-config-doc.js
lorenzolewis: @lucasfernog This was mainly around the `<file>` items in the schema because it was being recognised as HTML tags. I think if those are removed in the schema then this can be removed

lorenzolewis:
> I feel like the headers of the allowlist objects are too small, especially being next to the big tables. They're too easy to miss rn.

I could set it to where the minimum heading size is an h5 instead of h6 to help with that

> Edit: Or maybe they should get separate pages like in the js docs 🤔

I've been looking at using an automatic generator based on the schema (https://github.com/adobe/jsonschema2md) and it could do this, but we need to do some TLC on https://github.com/tauri-apps/tauri/issues/4139 first

lorenzolewis:
Just did a push for the headings, lmk what you think

FabianLars:
I'm talking about these:
grafik

And they are already h4, but imo they don't stand out enough to make clear that they're the headers of the tables

.scripts/generate-config-doc.js
FabianLars: yeah that's the 2 occurrences i was talking about :P

lorenzolewis:
> I'm talking about these: grafik
> 
> And they are already h4, but imo they don't stand out enough to make clear that they're the headers of the tables

@FabianLars how would you feel about leaving this as-is if we look at something like https://github.com/adobe/jsonschema2md in the future that can automatically split things to separate pages? Either that or we have h3 be the minimum heading size, but then that promotes every object up to the overview bar on the right side.

lorenzolewis:
For the scope objects, I don't quote know how to render "anyOf" type of properties (those are the types that generate the below type of table). But at least now it's rendered, in the current live config these aren't even rendered 😛 
Screenshot 2022-05-16 at 13 56 33

FabianLars:
> how would you feel about leaving this as-is if we look at something like https://github.com/adobe/jsonschema2md in the future that can automatically split things to separate pages? Either that or we have h3 be the minimum heading size, but then that promotes every object up to the overview bar on the right side.

Can we give them `font-size: 1.2rem;` or something for now?

> For the scope objects, I don't quote know how to render "anyOf" ...

Yeah the tables are alright-ish, but the text above is a little weird? idk how to explain it, but maybe it's connected to the small titles too.

Another thing i can't really explain: 
grafik

Like the 2 "Any of the following..." don't make sense there at all and i can't really see where they even come from; probably a scheme issue tho?

lorenzolewis:
Changed anchor to be on the same line of properties in the table and then I changed the format of anyOf and oneOf so it's a bullet list instead of a table now (seems to make more sense).

lorenzolewis:
It was bothering me so I overhauled it. Technically having the hierarchy like it was isn't a correct schema because a "definition" can relate to multiple objects (and was causing a lot of issues). So although this is "structured" as cleanly, it's _more correct_.

Just a few minor things I want to work on that I put in the description, but let me know what you think

lorenzolewis:
Ready to merge as soon as https://github.com/tauri-apps/tauri/pull/4153 is merged

lucasfernog (approved):
This is just amazing!!!

lorenzolewis:
@FabianLars you have a requested change open on this still

lorenzolewis:
> This is just amazing!!!

Thanks! Took me 4 iterations and 2 re-writes, but I got there in the end! :D

FabianLars (approved):
LGTM (for now :P )

lorenzolewis:
Quick, merge before he changes his mind!

Issue #589, opened 2022-05-11
Workaround for linking docs from page or footer


lorenzolewis:
Currently have to use `getCurrentVersion()` from utils/scripts.js to retrieve the "live" version of docs to assist with building a dynamic URL for a versioned doc. Raised https://github.com/facebook/docusaurus/issues/7402 upstream and I believe we rely on them for a fix.

PR #619, opened 2022-05-21
Feat/move about section

lorenzolewis:
# Changes
- Move content from `docs/about` to `pages/about`
- Fix links on `index.js` that were pointing to `docs`
- Add dropdown for about section to navbar in `docusaurus.config.js` since sidebars are not displayed for page types
- Remove about section from `sidebar.js`
- Fix links in `architecture.md` and `intro.md`. See note below on hard-coded URLs
- Adjust redirect to handle this move

Note: Currently the URLs for any docs contained in pages require the version number hard-coded if they are markdown files. See https://github.com/tauri-apps/tauri-docs/issues/589 for the upstream issue on this.


FabianLars:
The Mermaid thingy on the architecture page is broken

lorenzolewis:
Well shucks. This might be a fun one

lorenzolewis:
Alright, should be fixed now. Thanks for that catch!

FabianLars (changes requested):
The relative links to the v1 docs in the about pages are wrong (404 -> about/v1/...) other than that lgtm

lorenzolewis:
Gah! I guess I can't rely on the link checker when it's not in the docs folder. I may consider just moving these fully to `.js` types in the future because there's a lot of workarounds having to be put in place (and I don't like the hard-coded v1).

Anyways, thanks for catching that, should be fixed now @FabianLars

FabianLars:
about/architecture has the same problem :P

lorenzolewis:
Not anymore ;) And this time I checked all the pages, but those are the only two with internal links I found

FabianLars (approved):

PR #633, opened 2022-05-29
Feat/code theme

lorenzolewis:
Improve contrast for code blocks. Note: we can still customise our theme in the future to match branding, but this will at least give us a quick increase in separating code blocks from the rest of the content 

Before:
Screenshot 2022-05-29 at 20 23 48

After:
Screenshot 2022-05-29 at 20 24 01

Before:
Screenshot 2022-05-29 at 20 24 13

After:
Screenshot 2022-05-29 at 20 24 27


FabianLars (approved):
i changed it somewhat recently to the vs themes because of the same reason 😂

lorenzolewis:
This one still isn't perfect, but I think it's better than what we currently have ;)

FabianLars:
yeah it's just weird that it turned out to be unreadable _again_ in the end. Let's see how long this theme is usable 😂

PR #637, opened 2022-06-04
added SystemTrayEvent type to system tray example

onixldlc:
missing one of the type that was used in the example (im quite new to rust i dont know what to call these the error says its type so thats what i went with)


FabianLars (approved):

FabianLars:
thanks

PR #647, opened 2022-06-14
feat(tray): update Linux setup section

lucasfernog:


docs/guides/features/system-tray.md
```diff
@@ -23,17 +23,12 @@ The `iconAsTemplate` is a boolean value that determines whether the image repres
 
 #### Linux Setup
 
-On Linux, you also need to enable one of the `ayatana-tray` and `gtk-tray` Cargo features:
+On Linux, you need to install one of `libayatana-appindicator` or `libappindicator3` packages.
+Tauri determines which package to use at runtime, with `libayatana` being the preferred one if both are installed.
```
FabianLars, line 27: > being the preferred one if both are installed.

Aren't they mutually exclusive? Pretty sure that apt fails to install if the other package is already present (can't test it rn since i'm on windows)

FabianLars: anyway, doesn't really matter (mainly because i don't know how to phrase it any better).

lucasfernog: well at least on Arch I can install both IIRC

FabianLars: well, good enough for me then 😄

lucasfernog: Also I **think** if you have an AppImage using the gtk one and the system has ayatana, it'll use ayatana (which doesnt make much sense lol).

FabianLars: i think your [PR](https://github.com/tauri-apps/libappindicator-rs/pull/35/) fixed that

FabianLars (approved):

PR #661, opened 2022-06-16
Fix link to Debian/Linux building guide

shrimpza:
Very small fix, which had the link to Linux in the building landing page point to Windows.


FabianLars (approved):
thanks!

we will probably change that part of the docs in the future because it overlaps with distribution/linux

PR #662, opened 2022-06-16
Add french screenshots for translation

Pierstoval:
Will be used on translations that I'm currently contributing to on Crowdin


lorenzolewis:
Going to close this for now as we'll be using Crowdin's integration for any images. We haven't added the static dir that holds the images yet, but would be a good issue to add in. 

https://github.com/tauri-apps/tauri-docs/issues/663

PR #664, opened 2022-06-16
swap unaffliated youtube video with quick overview

jbolda:
## Motivation

I swapped a quick overview of Tauri instead of the YouTube video link. The video contained outdated and misleading content. We have previously discussed that we should not be enabling poor benchmarks, and linking this video directly enables that.

Also if we want a video, we should create this under our namespace rather than affiliating ourselves with someone outside of the Tauri team.

### DIFF
--- a/docs/guides/README.md
+++ b/docs/guides/README.md
@@ -4,12 +4,4 @@ sidebar_position: 1
 
 # Introduction
 
-## Quick Overview
-
-Check out [Tauri in 100 Seconds](https://www.youtube.com/watch?v=-X8evddpu7M) by [Fireship](https://www.youtube.com/channel/UCsBjURrPoezykLs9EqgamOA) on YouTube for a quick introduction to Tauri
-
-<div className="videowrapper">
-  <iframe width="560" height="315" src="https://www.youtube-nocookie.com/embed/-X8evddpu7M" frameborder="0" allow="accelerometer; autoplay; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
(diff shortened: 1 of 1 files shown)


lucasfernog (dismissed):

Laegel:
This page feels a bit... empty. Don't we have some older content to fill it a bit?

jbolda:
Yea, I don't disagree, but I think this is better than the alternative. It also didn't seem to make sense to expand on ideas that are covered in-depth in the later sections.

lorenzolewis:
We could build out a quick blurb for each section of the sidebar as a sort of summarised TOC (can do this dynamically). And then something around `create-tauri-app` as well at the very very top (@JonasKruckenberg feel free to add in some more thoughts here).

JonasKruckenberg:
Yeah nothing much to add, sounds good to me 👍

jbolda:
> We could build out a quick blurb for each section of the sidebar as a sort of summarised TOC (can do this dynamically). And then something around `create-tauri-app` as well at the very very top (@JonasKruckenberg feel free to add in some more thoughts here).

This feels like a potential good approach. We kind of already do this on the main page already though.

lorenzolewis:
> We kind of already do this on the main page already though.

I think once we get the Features section in particular revised, the index page will be updated and the "Features" bit of index will just hop directly into each of their respective guides.

lorenzolewis:
Going to close this one for now and keep track of it in https://github.com/tauri-apps/tauri-docs/issues/667

PR #677, opened 2022-06-17
Change "Brownfield" to "Independent"

lorenzolewis:
This is motivated by wanting to make our index page and docs more accessible to new users. "Independent" is a more recognisable term than "Brownfield".

This also has an update to the pattern name, so if we want to go ahead with that change we should look at updating the `PatternKind`in the `Tauri` config file. If we don't want to do that and keep the rename to more of a marketing-/external-rename then we could do that as well.


nothingismagick:
Brownfield has a specific meaning in project management. It is talking
about an 'old' project that is being ported. Independent is just wrong.

PR #681, opened 2022-06-18
Revise `create-tauri-app` usage

amrbashir:
- `npx` and `npm x` are the same, so they shouldn't have different tabs.
- I switched all npm usages to `npm create tauri-app` because it is similar to other package managers and `npm` had the `create` command for so long now.


FabianLars (approved):

lorenzolewis:
Looks great, thanks a million!

PR #683, opened 2022-06-18
Update link, break in mobile

hslee2008:
# New line in mobile
Link should move into new line for easier reading and clicking.

(before)
before

(after)
afer


FabianLars:
The link itself is just a placeholder until we release the blog post for v1. The mobile break is a nice fix tho

hslee2008:
Ok. I will change the link back.

FabianLars (approved):
thanks

PR #686, opened 2022-06-19
really? :)

ethanwillis:
for https://tauri.app/v1/guides/getting-started/setup/vite/ you are using relative paths which are correct when referring to a specific folder in a file path. However, when parsed to a URL this means that you need to go up an additional folder in order for the generated URL to be correct. However(again), if you do this then you will break docs links like this when people are browsing via github. The current best solution for you... just fully spell out the file path. 

I checked a few other files.. and didn't see the same mistake, but it's probably present elsewhere. 

https://docusaurus.io/docs/markdown-features/links !!!!

### DIFF
--- a/docs/guides/getting-started/setup/vite.mdx
+++ b/docs/guides/getting-started/setup/vite.mdx
@@ -58,13 +58,13 @@ pnpm create vite
   </TabItem>
 </Tabs>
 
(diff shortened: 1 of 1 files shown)


ethanwillis:
Deleting. will leave broken links for someone else to fix.

Issue #679, opened 2022-06-17
Move Tauri CLI guide


lorenzolewis:
Context: https://discord.com/channels/616186924390023171/987434624349261874/987477601180348429

We should move the CLI guide to he under "Quick Start". This helps with the overall flow of the getting started section and also shows that cargo is just another option to integrate with Tauri. 

CC @FabianLars

PR #690, opened 2022-06-19
Feat/cargo guide

lorenzolewis:
Tagging @JonasKruckenberg How should we handle installing the api here (is it even needed since I modelled this after the HTML guide)?

@lucasfernog @amrbashir @jbolda would love your feedback on this as well


lorenzolewis:
Closes #679

PR #706, opened 2022-06-24
Docs: Correct output path for msi on windows

arturh85:
noticed the wrong path for windows in the docs


lucasfernog (approved):
Thanks for catching this :)

PR #709, opened 2022-06-27
fixed isAbsolute doc spelling error

BeeMoe5:


FabianLars:
Thank you!
The js api docs are generated from the main tauri repo, i included this fix in https://github.com/tauri-apps/tauri/pull/4482. It will be automatically pulled on the next release (which is is right around the corner).

PR #712, opened 2022-06-28
docs: fixed arguments

MentalGear:
readDir doesn't take a datatype


FabianLars:
Thanks! The js api docs are generated from the main repo, so if you wanna fix this, feel free to head over to https://github.com/tauri-apps/tauri/blob/dev/tooling/api/src/fs.ts#L360

MentalGear:
In that case @FabianLars , please note that the "Edit Page" in the docs points to this git repo. It should then be adjusted to the file in the source for editing.

lorenzolewis:
@MentalGear good point, I put this as a separate issue in #714 to track

PR #715, opened 2022-06-29
Fixed spelling error

humanfriend22:
Fixes a spelling error in the code example for the isAbsolute function.


FabianLars:
Thank you :)
Duplicate of #709, the fix should be published today.

PR #718, opened 2022-06-30
fix: broken link in docs

archsolar:


archsolar:
nvm, it suddenly works.

PR #722, opened 2022-07-02
Remove note recommendation to target macOS 10.15

probablykasper:
I don't see a good reason to recommend building apps that are unsupported/blocked from running on older OS-es. We don't seem to recommend against supporting/running on Windows 7, which has gone even longer without security support.

Newer macOS versions drop support for older hardware, which means that supporting older OS-es can also potentially help reduce e-waste.

### DIFF
--- a/docs/guides/building/macos.md
+++ b/docs/guides/building/macos.md
@@ -19,10 +19,6 @@ GUI apps on macOS and Linux do not inherit the `$PATH` from your shell dotfiles 
 The minimum version of the operating system required for a Tauri app to run in macOS is `10.13`. If you need support for newer macOS APIs like `window.print` that is only supported from macOS version `11.0`
 onwards, you can change the [`tauri.bundle.macOS.minimumSystemVersion`]. This will in turn set the `Info.plist` [LSMinimumSystemVersion] property and the `MACOSX_DEPLOYMENT_TARGET` environment variable.
 
-:::caution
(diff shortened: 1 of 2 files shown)


JonasKruckenberg:
The problem (and reason why we added the 10.15 recommendation) is that we use some APIs that have been introduced in that release. So the situation is somewhat different to windows here 🤔

probablykasper:
Then it's not really an issue with security updates like the note says. It would probably be better to mention when a feature is only supported on a certain OS version, like we currently do with `window.print` requiring macOS 11.0+

lucasfernog:
We use some APIs added in 10.15 but we always try to fallback or check if the API exists - so we should still work on 10.13. If you find a crash please file an issue.

probablykasper:
Yeah that would make sense. Good to merge?

docs/guides/building/macos.md
```diff
@@ -19,10 +19,6 @@ GUI apps on macOS and Linux do not inherit the `$PATH` from your shell dotfiles
 The minimum version of the operating system required for a Tauri app to run in macOS is `10.13`. If you need support for newer macOS APIs like `window.print` that is only supported from macOS version `11.0`
```
lorenzolewis, line 19: It isn't directly related to this change, but we do specify macOS 10.15 and above in the tauri readme (although that's more meant that we support 10.15 and above, not that it's a minimum)

kasper9n: Ah interesting

lorenzolewis (approved):

PR #743, opened 2022-07-07
Pin netlify build, remove commit comment

lorenzolewis:
Made specifically just for @amrbashir :)


amrbashir (approved):

lorenzolewis:
I don't actually know which specific message this turns off because the docs aren't that great, so give me a shout if you still get the notifications you want after the run tonight

PR #749, opened 2022-07-10
correct relative URL for code signing link

RyKilleen:


lorenzolewis:
This would mean the target file is one directory up from the source file, but they are siblings so the current code is correct (plus the link isn't broken unless I'm missing something)

RyKilleen:
This PR was made to fix broken documentation I experienced myself last night.

This might be the wrong fix, but it is a broken link currently on the site.

FabianLars:
Can you show us where the broken link is on the live site? Because the one you changed here isn't broken.

RyKilleen:
I've attempted clearing cache and using incognito to rule out any aggressive client-side caching, no dice:

https://tauri.app/v1/guides/distribution/macos/

https://user-images.githubusercontent.com/4615840/178149100-d1a23f8e-587e-4a89-add6-82876170b6d1.mp4


It seems that the relative URL is being made relative to the file, not the file's folder.

FabianLars:
ahhh now i see it, it's because of the trailing `/` in your link. without it it works fine.

RyKilleen:
Pretty sure the trailing slash is normalized by Chrome whether I add it or not. I experience the issue whether I arrive there organically or go there directly.

RyKilleen:
Another option is to delete the link completely and rely on the "Next Section:" button like the Linux / Linux Code Sign pages do.

FabianLars:
> Pretty sure the trailing slash is normalized by Chrome whether I add it or not

Ah true, so we need to handle both types of links then (using the sidebar doesn't add it for example). `../` would break the links for the other variant tho i think.

> Another option is to delete the link completely and rely on the "Next Section:" button like the Linux / Linux Code Sign pages do.

Well, in this case, the plan is to move the non code signing guides to the Building category.

@lorenzolewis tldr: trailing slashes break our relative links :(

lorenzolewis:
Is it if the source page has a trailing slash and then you try to link to a target page via a relative link you get this issue?

RyKilleen:
The issue is that any URL paths that are relative are made relative to the current URL. According to the Docusaurus docs, instead of using URL paths, you could reference file paths. You can reference docs instead without the complications of browser relative pathing, which is resolved against the directory and not the browser URL. eg:

```md
[code signing guide]: sign-macos.md
// or
[code signing guide]: ./sign-macos.md
```

I'm not very familiar with Docusaurus specifically, but the gist of it is that either path endings need to be standardized in your front end so you can use relative paths consistently, or you can switch to referencing files so that the static build produces appropriate urls for you based on relative paths in the file directory.

lorenzolewis:
When we had the file extension before it was messing up i18n somehow, so we went through and removed them all. I wonder if it will work if we just have `sign-macos`?

Does the "HTML/CSS/JS" link at the bottom of this page behave the same for you: https://tauri.app/v1/guides/getting-started/setup/? It's using the same format as what's currently on the page in this PR

FabianLars:
That one works fine, because it's not the same. The equivalent would be a link to vite from the html guide. But you can have see a similar problem by clicking on the interprocess link all the way at the bottom of the html guide (if you have a trailing slash)

lorenzolewis:
The trailing slash is actually added by Netlify. I can remove it by disabling pretty urls. I can't think of any side effects of this, any come to mind for you @FabianLars ?

Source: https://github.com/slorber/trailing-slash-guide

lorenzolewis:
This won't fix the problem if you're already on a page with a trailing slash, though. Just helps to reduce the chance of navigating to one.

lorenzolewis:
I'm going to close this in favour of https://github.com/tauri-apps/tauri-docs/issues/754 since this will require some link cleaning across the repo to fix I'm pretty sure. Should be quick, but just need to do some testing first to make sure we don't break anything

PR #766, opened 2022-07-13
Minor Fixes

lorenzolewis:
- Adjusts capitalisation for blog title
- Remove Lottie-player dep (no longer used
- Remove themify icons (no longer used)
- Removed dynamic-css plugin (only used for removed themify icons)
- Remove unused `#starfield` attribute from CSS
- Remove used `Icon` and `OSList` from theme modules
- Move crowdin CLI down to dev dep
- Fix index package svgs (extra border)


src/css/custom.css
```diff
@@ -110,6 +110,10 @@ h2 > [class^='ti-'] {
   min-height: 52px;
 }
 
+[class*='announcementBar_'] a:hover {
+  color: var(--ifm-button-color) !important;
+}
```
Beanow: Doesn't look like this resolves as a color for me on Firefox
image
What's the expected result?

FabianLars: It changes the link color only on hover

FabianLars:
This makes the announcement links inconsistent with the other broken links 👀 (Quick Start and Learn More)

src/css/custom.css
Beanow: Right, should have shown with hover emulated.
It's supposed to be the same as the text beside it I assume? So no hover effect essentially.
image

lorenzolewis:
> This makes the announcement links inconsistent with the other broken links 👀 (Quick Start and Learn More)

Lol, love just casually throwing in the other ugly CSS issues here 😂

A few more custom-CSS things than I'd like, but that's what we get when we use blue as a primary brand colour. How's this look for you both now? Updated the original comment with a few more changes I did (general cleanup).

src/css/custom.css
lorenzolewis: Updated this, lmk what you think

FabianLars:
1) The `Quick Start` buttom still has the "invisible on hover" issue.
2) The announcement link doesn't look like a link without hovering on it (and why would someone move their mouse on it if it doesn't look clickable)

lorenzolewis:
Can you post a screenshot of what you're talking about with #1?

Any ideas on how to help with #2? I tried bolding but it doesn't work

FabianLars:
grafik

FabianLars:
about 2) maybe just re-add the underline and make it disappear on hover? Basically the reverse of all other links.
or add a `here`, i think that could suggest a clickable link too

lorenzolewis:
Give that a try (when the build finishes). The button CSS on the infirma site (backing all the Docusaurus CSS) is different than what they actually ship :)

But maybe that will change once I update to the docusaurus rc because I saw they pulled an update through.

For the announcement bar I just kept the underline all the time because I felt like that makes more sense than inverting the underline stuff, but let me know what you think

lorenzolewis:
Well, now that 2/2 of the extra problems fixed themselves, this was a fun little cleanup PR at the very least, merge ready?

FabianLars (approved):

PR #771, opened 2022-07-15
Feat: DocCard Image

lorenzolewis:
Adds a snazzy image to doc cards if opted-in

Screenshot 2022-07-15 at 15 41 10

# Changes

- Hide TOC on setup page (right sidebar)
- Add `sidebar_custom_props values `doc_card_image` and optional `doc_card_image_dark` to HTML and Vite guides
- Add HTML (light and dark) and Vite logo assets
- Remove unused TS def file for DocCard (reverted in Docusaurus)
- Adapt Swizzled DocCard component to check for a custom props `doc_card_image` and `doc_card_image_dark` and if present to render them (can use only `doc_card_image` if the main image works fine for either colour mode)
- Hide the emoji icon for a DocCard if an image is used
- Add additional CSS to DocCard to allow for image (using flex because magic 🪄)

# Implementation Notes

- Currently images have to be stored in the `static` folder due to MD relative URLs having special parsing magic that I couldn't figure out
- This is using our custom CSS for `gh-light-mode-only`/`gh-dark-mode-only` under the hood to toggle instead of using a react hook
-


JonasKruckenberg (approved):
The light/dark variation of the HTML logo is a nice touch

Issue #754, opened 2022-07-10
Trailing Slash URLs


lorenzolewis:
Relevant to https://github.com/tauri-apps/tauri-docs/pull/749

If we don't include the file Urls in our links then they don't get post-processed by Docusaurus and can lead to some issues with Url navigation if a backslash is included in the browser's current URL.

Previously I removed all file extensions because it was causing some issues with the i18n'ed sites. We've had quite a bit of work on i18n since then so we may be safe to add these back in (and will also opt us into broken md link alerts since they aren't checked if the extension isn't included).

Currently I'm not sure how to link to a target page anchor for a MD page AND get the broken link check to work, but that's a minor thing and won't break anything at this point.

PR #775, opened 2022-07-16
Fix broken links

Seeker14491:


lorenzolewis:
I believe this is caused by https://github.com/tauri-apps/tauri-docs/issues/754. I'm assuming that you have a trailing slash in your URL when you try to navigate to a page and the destination leads to a 404?

lorenzolewis:
@Seeker14491 can you test the same links here and let me know if they're fixed? https://deploy-preview-777--tauri.netlify.app

Seeker14491:
The links seem to be fixed now.

PR #776, opened 2022-07-16
Feat/configurations guide

### LINKED ISSUE #252: tauri.conf.json, package.json, Cargo.toml
There are many places to configure things, and we need to do not only a better job documenting all the possibilities, but also explain how they interact and impact a project.

I am thinking we need a whole section about this, with a general overview and then individual pages for each of these.

lorenzolewis:
Closes #252 , still want to have a section that explains what each of the generated folders are outside of the guides


lorenzolewis:
@FabianLars checking in if GitHub timed out for you again on this one

FabianLars:
shit, got nobody/nothing to blame for this time, just forgot it, sorryyy

docs/guides/architecture/configuration-files.md
```diff
@@ -0,0 +1,91 @@
...
+## `package.json`
+
+This is the Node.js package file. If the frontend of a Tauri app is developed using Node.js-based technologies this file is used to configure the frontend such as which dependencies are used. It's also common to use the `"scripts"` section to store the command used to launch the frontend used by your Tauri application.
+
+An example of a barebones `package.json` file for a Tauri project might look a little something like this:
+
+```json title=package.json
+{
+  "scripts": {
+    "dev": "command-for-your-framework"
```
FabianLars:
```suggestion
    "dev": "command-for-your-framework",
    "tauri": "tauri"
```

FabianLars:
So, the technical stuff etc looks correct, i still think some sentences sound a little rough, but my english is too bad to really point it out x)

lorenzolewis:
@FabianLars I did some re-wording. lmk if there are any bits that stick out or are rough around the edges

docs/guides/architecture/configuration-files.md
```diff
@@ -0,0 +1,98 @@
...
+
+Since Tauri is a toolkit for building applications there can be many files to configure project settings. Some common files that you may run across are `tauri.conf.json`, `package.json` and `Cargo.toml`. We briefly explain each on this page to help point you in the right direction for which file to modify.
+
+## `tauri.conf.json`
+
+This is the file used by the Tauri process. You can define build settings (such as the [command run prior to `tauri build`][before-build-command] or [`tauri dev`][before-dev-command]), set the [name and version of your app][package-config], [control the Tauri process][tauri-config], and configure any plugin settings. You can find all of the options in the [`tauri.conf.json` API reference].
+
+## `Cargo.toml`
+
+Cargo's manifest file used to declare Rust crates your app depends on, metadata about your app, and other Rust-related features. If you do not intend to do backend development using Rust for your app then you may not modifying it much, but it's important to know that it exists and what it does.
```
FabianLars:
```suggestion
Cargo's manifest file is used to declare Rust crates your app depends on, metadata about your app, and other Rust-related features. If you do not intend to do backend development using Rust for your app then you may not be modifying it much, but it's important to know that it exists and what it does.
```

docs/guides/architecture/configuration-files.md
```diff
@@ -0,0 +1,98 @@
...
+[features]
+# by default Tauri runs in production mode
+# when `tauri dev` runs it is executed with `cargo run --no-default-features` if `devPath` is an URL
+default = [ "custom-protocol" ]
+# this feature is used used for production builds where `devPath` points to the filesystem
+# DO NOT remove this
+custom-protocol = [ "tauri/custom-protocol" ]
+```
+
+The most important parts to take note of are the `tauri-build` and `tauri` dependencies. Their versions must be the same as the Tauri CLI in order to ensure that there won't be any issues running your app. If any of these are not the same version number you are highly likely to encounter issues.
```
FabianLars: welp, this is not correct anymore. And not that big of a problem now that the cli and tauri-build are more decoupled. There is still a requirement to have them somehow match, but no idea how one would phrase that.

docs/guides/architecture/configuration-files.md
```diff
@@ -0,0 +1,98 @@
...
+
+Cargo version numbers use [Semantic Versioning]. Running `cargo update` will pull the latest available Semver-compatible versions of all dependencies. For example, if you specify `1.0.0` as the version for `tauri-build`, Cargo will actually detect and download version `1.0.4` because it is the latest Semver-compatible version available. Tauri will update the major version number whenever a breaking change is introduced, meaning you should always be capable of safely upgrading to the latest minor and patch versions without fear of your code breaking.
+
+If you want to use specific crate version you can use exact versions instead by prepending `=` to the version number of the dependency:
```
FabianLars:
```suggestion
If you want to use a specific crate version you can use exact versions instead by prepending `=` to the version number of the dependency:
```

docs/guides/architecture/configuration-files.md
```diff
@@ -0,0 +1,98 @@
...
+  "dependencies": {
+    "@tauri-apps/api": "^1.0",
+    "@tauri-apps/cli": "^1.0"
+  }
+}
+```
+
+It's common to use the `"scripts"` section to store the command used to launch the frontend used by your Tauri application. The above file specifies the `dev` command that you can run using `yarn dev` or `npm run dev` to start the frontend framework.
+
+The dependencies object specifies which dependencies Node.js should downloaded when you run either `yarn` or `npm install` (in this case the Tauri CLI and API).
```
FabianLars:
```suggestion
The dependencies object specifies which dependencies Node.js should download when you run either `yarn` or `npm install` (in this case the Tauri CLI and API).
```

docs/guides/architecture/configuration-files.md
lorenzolewis: I really don't like that we have different version numbers in that monorepo, it makes things a lot harder to keep on top of. Let me take a stab at rephrasing this.

lorenzolewis: @FabianLars thoughts on this phrasing?
```suggestion
The most important parts to take note of are the `tauri-build` and `tauri` dependencies. Generally they must both be on the latest minor versions as the Tauri CLI, but this is not strictly required. If you encounter issues while trying to run your app you should check that any Tauri versions (`tauri` and `tauri-cli`) are on the latest versions for their respective minor releases.
```

docs/guides/architecture/configuration-files.md
```diff
@@ -0,0 +1,98 @@
+---
+sidebar_position: 1
+---
+
+# Configuration Files
+
+Since Tauri is a toolkit for building applications there can be many files to configure project settings. Some common files that you may run across are `tauri.conf.json`, `package.json` and `Cargo.toml`. We briefly explain each on this page to help point you in the right direction for which file to modify.
+
+## `tauri.conf.json`
```
amrbashir:
```suggestion
## `tauri.conf.json`, `tauri.conf.json5` and `Tauri.toml`
```

lorenzolewis:
```suggestion
## Tauri Config (`tauri.conf.json`, `tauri.conf.json5`, or `Tauri.toml`)
```

docs/guides/architecture/configuration-files.md
```diff
@@ -0,0 +1,98 @@
...
+
+This is the file used by the Tauri process. You can define build settings (such as the [command run prior to `tauri build`][before-build-command] or [`tauri dev`][before-dev-command]), set the [name and version of your app][package-config], [control the Tauri process][tauri-config], and configure any plugin settings. You can find all of the options in the [`tauri.conf.json` API reference].
```
lorenzolewis:
```suggestion
This is the file used by the Tauri process. You can define build settings (such as the [command run prior to `tauri build`][before-build-command] or [`tauri dev`][before-dev-command]), set the [name and version of your app][package-config], [control the Tauri process][tauri-config], and configure any plugin settings. You can find all of the options in the [`tauri.conf.json` API reference].

:::note

The default Tauri config format is `.json`. `.json5` and `.toml` formats can be enabled by adding the feature flag `config-json5` or `config-toml` (respectively) to the `tauri` dependency in `Cargo.toml`. The structure and values are the same across all formats, however the formatting should be consistent with the respective file format.

`.toml` is only available from Tauri 1.1.0 and above.

:::
```

docs/guides/architecture/configuration-files.md
FabianLars: Why did i not get a notification 😢 But yeah sounds good x)

docs/guides/architecture/configuration-files.md
FabianLars: This should probably mention that json5 and toml have their own feature flags _which are disabled by default_. The flags are `config-json5` and `config-toml`

lorenzolewis: Where do those flags go? In the `Cargo.toml` file?

FabianLars: Yeah on the tauri dependency, like this thingy here https://github.com/tauri-apps/tauri/blob/dev/examples/sidecar/src-tauri/Cargo.toml#L14

lorenzolewis: And added to the regular dependency, not the build dependency, right?

FabianLars: yeah

FabianLars: Actually tauri-build has that flag too, but i've never seen it being used 🤔 Gonna double check if it's enough on tauri

lorenzolewis:
Updated the Tauri Config section to look like the below screenshot. @FabianLars let me know if something should be added for `tauri-build` feature flag.

Screenshot 2022-08-12 at 17 38 23

FabianLars:
Okay so it's a little weird. In theory it needs it because the `cargo:rerun-if-changed` lines are behind a flag check (and also if we want to be on the safer side with its dependency on tauri-utils which otherwise would rely on tauri enabling its feature). "In theory" because while normally these rerun thingies are needed to tell cargo when to rerun the build script, it seems to always rebuild it anyway.
That sounds like a bug, and i think i saw an issue or PR about it too but can't seem to find it.

Anywayyy there's an even bigger bug because even tho it seems to rerun the build script every time, it doesn't actually apply the changes. _Edit:_ Or is that an issue with the cli? What i tested was changing the app name and modifying the allowlist.

Btw, i tested this on the latest released versions, Can't try the git version right now. Also i only tested it on json5 rn (with and without the feature flag).

cc @lucasfernog

lucasfernog:
Sorry i'm a bit lost, can you guys explain what's the issue?

FabianLars:
@lucasfernog 
1) documentation question: Do users need to add the `config-json5` feature flag to tauri-build too, or is it enough on tauri? https://github.com/tauri-apps/tauri/blob/dev/core/tauri-build/src/lib.rs#L203 suggests that it is indeed needed. (even though the script re-runs on json5 changes if the flag is disabled)

2) Bug: While you're in `tauri dev` mode, changes to the config file won't get applied unless you manually rerun `tauri dev` (even though the file watcher correctly triggers a rebuild). For example changing the allowlist doesn't update the feature flag in Cargo.toml without a manual restart.
Edit: To clarify: it does try to rebuild automatically, but you get an issue about incorrect feature flags.

lorenzolewis:
For #1 ideally it should only be for one (either `tauri` or `tauri-build`) and only exist as a feature flag on one to avoid confusion I'd think. One less thing to have to double-check is synced up.

lucasfernog:
I think we can make it required only on tauri. Need to check that though.

lucasfernog:
@FabianLars can you file an issue for #2?

lorenzolewis:
Okay, if that's the direction we go then I think this is accurate once 1.1 is released, right? Any other feedback on it?

P.S., after this is merged I'm going to start working on separating out "concept"/"reference" type things like we were discussing on Discord before

FabianLars (approved):

PR #777, opened 2022-07-16
Revise links, minor codeblock formatting

### LINKED ISSUE #754: Trailing Slash URLs

lorenzolewis:
Fixes https://github.com/tauri-apps/tauri-docs/issues/754 (hopefully)

Tested this against a local i18n build for ko, zh, and fr and it seems to work fine.

### DIFF
--- a/docs/.templates/cli.md
+++ b/docs/.templates/cli.md
@@ -70,4 +70,4 @@ This command will show the current version of Tauri.
 
 ## CLI usage
 
-See more about the usage through this [complete guide](../guides/development/development-cycle).
+See more about the usage through this [complete guide](../guides/development/development-cycle.md).
--- a/docs/api/cli.md
+++ b/docs/api/cli.md
@@ -201,4 +201,4 @@ This command will show the current version of Tauri.
 
 ## CLI usage
 
-See more about the usage through this [complete guide](../guides/development/development-cycle).
+See more about the usage through this [complete guide](../guides/development/development-cycle.md).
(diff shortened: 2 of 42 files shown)


lorenzolewis:
@FabianLars this should fix the trailing slash problem. Can you double-check it with the below links?

https://deploy-preview-777--tauri.netlify.app/v1/guides/getting-started/setup/vite/ (link to prerequisites page)
https://deploy-preview-777--tauri.netlify.app/v1/guides/distribution/macos/ (link to signing page)

FabianLars:
On the first linked page, all the way at the bottom the info box: The link to the interprocess doc doesn't work if there's a trailing slash. (idk if it's just this link or others too rn)

lorenzolewis:
Give it a try now @FabianLars , should be fixed for HTML and Vite pages

FabianLars (approved):

Issue #790, opened 2022-07-21
More specific directions on user guide walkthrough


skishlin:
Hi,

I'm working on making videos/a user guide of using Tauri for absolute beginners. I'm pretty raw when it comes to computer coding and web development, but I spent the last 2 years teaching myself Data Analytics, Python and SQL and I've been learning Rust since February.  The user guide I have in mind is for people coming to the programming/development world with a very limited knowledge base like myself who might be coming to Tauri either for amateur recreation, or as a necessity for work-related projects and need a more layman (see: Troglodyte) approach to utilize it quickly and efficiently for a project they might be working on. 

I noticed that the guide, while informative about the various options available to create the front and backend framework using CLI, doesn't specify that the various methods are either or, rather than steps to be followed sequentially.  I will be filing a PR. 

Thanks

FabianLars:
I'll close this as a duplicate of #781, even tho it's slightly different, but at least now the 2 issues are linked to each which should be enough for us to track it.

Issue #781, opened 2022-07-17
[docs] I just went through the quickstart and noticed a few issues


paustint:
I ended up having to start over because I followed the guide which led me astray.

I was following https://tauri.app/v1/guides/getting-started/setup/vite

I started with `yarn create tauri-app`
* `Create the Rust Project` -> this was already done as part of `yarn create tauri-app`, so when I attempted to do this I got an error saying that the `src-tauri` directory was not empty and I did a force install as suggested. Instead all I really needed to do was edit `tauri.conf.json` and change the port (but I was not sure if other stuff would have been different, which is why I opted for force the first time)
  * In addition, the instructions to update `src-tauri/tauri.conf.json` do not honor yarn/npm/pnmp configuration 
* This command offered to add `@tauri-apps/api` for me (and I accepted), so I did not need to explicitly do this step called out in the docs

Also, on the Development Cycle page - this is called out:
>In your project repository, you SHOULD commit the "src-tauri/Cargo.lock" along with the "src-tauri/Cargo.toml" to git because Cargo uses the lockfile to provide deterministic builds. As a result, it is recommended that all applications check in their Cargo.lock. You SHOULD NOT commit the "src-tauri/target" folder or any of its contents.

Ideally the cli would manage this by default for projects structured in the standard way - `src-tauri/target` should be added to `.gitignore` by default since the cli generated the file to start with.

FabianLars:
Just to clear it up a bit. The docs currently mainly describe how to set it up without create-tauri-app (that's why it also shows how to use create-vite). So really, the only somewhat valuable information here is the example vite config, which is not necessary for using vite with tauri.
But i guess we need to make that clearer than just the small infobox at the top (which doesn't even mention that specifically).

> Ideally the cli would manage this by default for projects structured in the standard way - src-tauri/target should be added to .gitignore by default since the cli generated the file to start with.

It is managed automatically, there is a `.gitignore` file in the src-tauri dir, the comment is there so that people don't add the Cargo.lock file themselves.

Thanks for the feedback btw!

lorenzolewis:
Reading through the `create-tauri-app` callout at the top I guess we don't do a good job explaining that it's an ALTERNATIVE to manually setting up tauri with the rest of the guide. So we can do some work to update the phrasing to highlight that.

paustint:
>..You SHOULD NOT commit the "src-tauri/target" folder or any of its contents.

[...]

PR #791, opened 2022-07-21
Enhancement of vite page

skishlin:
Please see issue #790


FabianLars:
I do agree that we have to change _something_, see #781, but i'm not a big fan of this particular change, we'll keep in mind tho, and thank you so much for contributing!

PR #798, opened 2022-07-24
Fix incorrect section level

sa2taka:
Fixes to be the same structure as other sections.
e.g. 

```
### pictureDir

...

#### Platform-specific
```

### DIFF
--- a/docs/api/js/modules/path.md
+++ b/docs/api/js/modules/path.md
@@ -489,7 +489,7 @@ ___
 
 Returns the path to the suggested log directory.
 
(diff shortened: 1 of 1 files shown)


lorenzolewis:
Thanks for this. The JS docs are generated upstream from the tauri repo directly, so I've put in a PR there and this will automatically get fixed once that's merged.

lorenzolewis:
Upstream has now been merged and this will get fixed in the next release, going to close this for now. Thanks for catching!

PR #804, opened 2022-08-05
Fix/blog dates

### LINKED ISSUE #783: Blog posts all say they were published on the current date
Looks like the YAML frontmatter on the blog posts dont have a published at date and something is filling in today's date.

lorenzolewis:
Fixes #783 and a broken link

Note that blog post urls are changed, but this should be more sustainable for the future


FabianLars:
We should add redirects, or at least for the 1.0 post

lorenzolewis:
Makes sense, just added it in

FabianLars:
hmmm doesn't seem to work 🤔

lorenzolewis:
(Mumbles _again_ at CI/CD) It's because the CI workflows get the contents for that from the target branch (so dev) instead of the branch that's being merged. So I need to update CI for that final step I think for it to accurately update for deploy previews

lorenzolewis:
I need to fix some broken i18n stuff there as well so let me take a stab at that this weekend and then can use this as a test to see if it's fixed

lorenzolewis:
@FabianLars could you double check on your end? Redirect should be fixed now.

FabianLars (approved):

PR #816, opened 2022-08-12
feat: embed classes/enums/interfaces in the module

amrbashir:


amrbashir:
Looks like it won't deploy until broken links are fixed which I am not going to fix now. I just wanna get some initial feedback, so build locally to see the changes.

PR #823, opened 2022-08-16
Feat/move release notes

lorenzolewis:
This is in prep of creating the "Concepts" section as another L1 navbar item to help cleanup the amount of top-level links we have


amrbashir:
tbh, I like the position of where it was, it doesn't makes sense under API since it includes release notes of the tools too not just the APIs

lorenzolewis:
> tbh, I like the position of where it was, it doesn't makes sense under API since it includes release notes of the tools too not just the APIs

Hmm yeah, this does make sense. I see the value in having the page in general, but I wonder if we could move the link down into the footer instead? We could rename the title of the footer if we wanted to. 

Screenshot 2022-08-16 at 22 02 43

Unfortunately Netlify doesn't give us analytics across all pages, only the "top" performing pages. But I can confirm this isn't one of the "top" performing pages based on those metrics ;)

amrbashir:
Yeah I am aware that it is not often visited, but we need to keep it for historical reasons. Can't we just move it at the end of the order after `Community`, and place `Concepts` before it?

amrbashir (approved):

FabianLars (approved):

PR #829, opened 2022-08-20
Feat/references section

lorenzolewis:
This refactors some of the site into a "References" section. 

This restructure will allow us to re-flow the "Guides" section of the docs to be read from top to bottom as a developer through the lifecycle of creating a Tauri app. The "References" section is important and will be referenced and linked to a lot in the "Guides", but is not _strictly_ critical to creating a Tauri app.

In the future we could create more hands-on guides for in-depth how-tos that showcase best practices using our APIs and Guides in blog posts that can act as a natural bridge between the Guides and References sections.

## Changes

- `guides/architecture/inter-process-communication/*` (as well as Brownfield and Isolation patterns) moved to `references/architecture/inter-process-communication/*`
- `guides/architecture/process-model` moved to `references/architecture/process-model`
- `about/architecture` moved to `references/architecture`
- Moved all architecture recipes, although these are only drafts and not published on the site (so no redirect is included)
- `about/security` moved to `references/architecture/security` (as the source page is on Tauri security, not application development security)
- `about/benchmarks` moved to `references/benchmarks`
- `guides/development/security` removed as it was a duplicate of `guides/architecture/security` largely
- Update broken links across the site caused by this
- Pagination next button for setup guides now go to the next section (`development/development-cycle`)
- Create index page for References section
- Fix sidebar sort order caused by file moves
- Update nav bar:
  -  Move the old "About" dropdown to the end of the leading group of options 
  - Remove Architecture, Security, and Benchmarks links as they are now auto-generated in the sidebar of the "References" section
  - Add a "References" item to be immediately after the "Guides" item
- Add Netlify redirects for links that would be broken (testing that these are correct would be much appreciated)
- Add a `references` sidebar slice to auto-generate the sidebar for the whole section


amrbashir (approved):
LGTM! Although I would maybe make the `References` has a dropdown menu to show its subpages which will give a better idea what is this about. Also maybe swap `API` position with `References`, I think `Guides` and `API` are naturally related and they should be next to one another.

FabianLars:
Or where About used to be = the first one on the left. If references is supposed to be some groundwork/base information that could make sense too idk

FabianLars (approved):

lucasfernog (approved):

lorenzolewis:
Moving API next to guides makes sense. I'll have to see about the drop down because from what I remember I think that has to be statically coded.

lorenzolewis:
@amrbashir I changed the order to put References third (after Guides and API), but I couldn't get a dropdown to work how I'd like to (you apparently can't have the main item link anywhere if it's a dropdown, only the children can link), so maybe we can circle back to that after a few pages pending this PR are merged and we see what it could look like.

PR #833, opened 2022-08-23
docs(clipboard): Added allowlist section to the Clipboard module

mateo-gallardo:
Added missing bit about adding the available Clipboard APIs to `tauri.conf.json`.


FabianLars:
Thank you! Just a small issue, this file is autogenerated from this one https://github.com/tauri-apps/tauri/blob/dev/tooling/api/src/clipboard.ts 😅 So i'd appreciate it very much if you could change that over there instead ❤️

mateo-gallardo:
Oops, my bad. [Here it is.](https://github.com/tauri-apps/tauri/pull/5014)

PR #839, opened 2022-08-27
Add create-tauri-app new command

amrbashir:


lorenzolewis:
My first reaction to this is this is too many tabs for a call-to-action (yay for multiple CTA acronyms). And having `curl` and `wget` both as well defeat the call to action of one thing as well (and mess with the layout).

Screenshot 2022-08-30 at 21 15 56

I think there should only be one option in the node ecosystem (just for the index page for simplicity sake) and _maybe_ drop PowerShell (although if we remove yarn and pnpm and keep npm there could still be enough room).

For cargo, is there a shorter way to do `cargo install create tauri-app && cargo create-tauri-app` because it would be nice to also have this as a "one-liner" like the others.

One last thing, we've been getting a lot of feedback on confusion that we have the CTA banner on the HTML and Vite pages because people read this as "run the CTA command AND THEN continue down the page" instead of either option. Maybe we write up a separate CTA page (at the same level of HTML and Vite) and then this "Quick Start" button on the home page goes there instead? Then we could have the additional methods (PowerShell, all the Node options, etc.) and some more details there. Just bouncing ideas around.

amrbashir:
> My first reaction to this is this is too many tabs for a call-to-action (yay for multiple CTA acronyms). And having `curl` and `wget` both as well defeat the call to action of one thing as well (and mess with the layout).

I don't see the issue with many tabs, only one of them is active. 

As for `wget` and `curl` we can explicitly specify `OR` between the two commands, it is normal for linux users to make this choice but we can just remove `wget` and only provide curl.

> I think there should only be one option in the node ecosystem (just for the index page for simplicity sake) and _maybe_ drop PowerShell (although if we remove yarn and pnpm and keep npm there could still be enough room).

Why remove `powershell`? this defeats the purpose of why I made this rewrite in the first place (which is to reach as many communities as possible and no one feels excluded.

For npm, js ecosystem is accustomed to have the 3 package managers instructions directly, they may get confused and don't notice it if its under another tab. 

> For cargo, is there a shorter way to do `cargo install create tauri-app && cargo create-tauri-app` because it would be nice to also have this as a "one-liner" like the others.

This is considered a one liner. Windows doesn't have `&&` so we need both commands on separate lines to work in a cross-platform manner.

> One last thing, we've been getting a lot of feedback on confusion that we have the CTA banner on the HTML and Vite pages because people read this as "run the CTA command AND THEN continue down the page" instead of either option. Maybe we write up a separate CTA page (at the same level of HTML and Vite) and then this "Quick Start" button on the home page goes there instead? Then we could have the additional methods (PowerShell, all the Node options, etc.) and some more details there. Just bouncing ideas around.

As I explained in discord. This is not needed and will serve nothing tbh as there is no further instructions after the project has been bootstrapped.

docs/guides/getting-started/setup/_fragments/_tauri-init.mdx
```diff
@@ -10,28 +10,28 @@ Our Tauri CLI uses Cargo under the hood so you rarely need to interact with it d
 If you haven't installed the Tauri CLI yet you can do so with with one of the below commands. Aren't sure which to use? Check out the [FAQ entry].
 
 <Tabs groupId="package-manager">
-<TabItem value="npm" label="npm" default>
+<TabItem value="npm" label="NPM" default>
```
amrbashir: `npm` should be the correct one, that's how they refer to it in the official docs.

docs/guides/getting-started/setup/_fragments/_tauri-init.mdx
```diff
@@ -10,28 +10,28 @@ Our Tauri CLI uses Cargo under the hood so you rarely need to interact with it d
...
-<TabItem value="yarn" label="yarn">
+<TabItem value="yarn" label="Yarn">
 
 ```shell
 yarn add -D @tauri-apps/cli
 ```
 
 </TabItem>
-<TabItem value="pnpm" label="pnpm">
+<TabItem value="pnpm" label="PNPM">
```
amrbashir: same thing for `pnpm`

lorenzolewis: Pssssshhhhhh what are capitalisation rules? Node ecosystem just throw them out the window

lorenzolewis:
@amrbashir I fixed the capitalisation bits, are you okay with the other changes?

docs/guides/getting-started/setup/_fragments/_tauri-init.mdx
amrbashir: yeeeeeet

src/theme/Command.js
```diff
@@ -6,9 +6,9 @@ import TabItem from '@theme/TabItem'
 
 const types = [
   { label: 'npm', value: 'npm run tauri ' },
-  { label: 'yarn', value: 'yarn tauri ' },
-  { label: 'pnpm', value: 'pnpm tauri ' },
-  { label: 'cargo', value: 'cargo tauri ' },
+  { label: 'Yarn', value: 'yarn tauri ' },
+  { label: 'npm', value: 'pnpm tauri ' },
```
amrbashir:
```suggestion
  { label: 'pnpm', value: 'pnpm tauri ' },
```

amrbashir:
I can't approve my own PR but it is good to go now.

src/theme/Command.js
```diff
@@ -21,18 +21,32 @@ function insertDashDashBeforeOption(value, name) {
...
+      <TabItem value="npm">
+        <CodeBlock className={`language-shell`}>npm create tauri-app</CodeBlock>
+      </TabItem>
+      <TabItem value="Yarn">
         <CodeBlock className={`language-shell`}>
           yarn create tauri-app
         </CodeBlock>
       </TabItem>
-      <TabItem value="pnpm">
+      <TabItem value="npm">
```
amrbashir:
```suggestion
      <TabItem value="pnpm">
```

lorenzolewis (approved):

PR #840, opened 2022-08-30
Fix incorrect heading level under logDir

aubrey-wodonga:
What it says on the tin. `Platform-Specific` subheadings should be level-4 headings.


lorenzolewis:
Hey @simonxciv thanks for this! The API files are generated in the main Tauri repo and we've fixed this in there https://github.com/tauri-apps/tauri/blob/dev/tooling/api/src/path.ts. Those changes will get pulled in and updated on the next Tauri release.

Cheers!

PR #820, opened 2022-08-15
fixes #819 wrong command in docs for cargo update

thadguidry:
fixes #819 

Example:
```
E:\GitHub Repos\openrefine-tauri\src-tauri>cargo upgrade
error: no such subcommand: `upgrade`

        Did you mean `update`?

E:\GitHub Repos\openrefine-tauri\src-tauri>cargo update
    Updating crates.io index
    Updating anyhow v1.0.56 -> v1.0.61
    Removing async-broadcast v0.3.4
    Removing async-channel v1.6.1
<snip>
```

PR #670, opened 2022-06-16
fix: cargo upgrade to cargo update

milos018:

PR #841, opened 2022-08-31
update `cargo upgrade` command

thedoublejay:
Was trying out the docs, I think `cargo upgrade` was renamed to `cargo update`. https://doc.rust-lang.org/cargo/commands/cargo-update.html

### DIFF
--- a/docs/guides/development/updating-dependencies.md
+++ b/docs/guides/development/updating-dependencies.md
@@ -48,7 +48,7 @@ cd src-tauri
 cargo update
 ```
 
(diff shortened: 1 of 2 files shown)


FabianLars:
Thank you for contributing! Unfortunately this change is not correct, see #670 and #820.
I'll remove the note about updating dependencies from the prereq page altogether when i can.

The part in updating-dependencies will most likely stay with the additional clearification that the mentioned `cargo-edit` tool is a _separate_ tool that needs to be installed first.

PR #845, opened 2022-09-01
Fix typo in _tauri-init.mdx

thiagodebastos:
This commit addresses a typo where a `pnpm` example is incorrectly labelled as `npm`, resulting in what looks like two examples using `npm`.


lorenzolewis (approved):
Thanks for this!

PR #838, opened 2022-08-26
feat(docs/guides) Added NextJS guide

0xle0ne:

PR #850, opened 2022-09-03
Add Next.js Framework Guide

knownasnaffy:
This will add a setup guide for using Nextjs framework as a frontend tool with Tauri app. It follows the `nextjs-ts` template in the `create-tauri-app`. Some instructions have been copied from [Pull#838](https://github.com/tauri-apps/tauri-docs/pull/838) by @leon3s


knownasnaffy:
The update-docs workflow is failing, is there a fault in my code?

FabianLars:
Don't worry about it, it's failing because of an issue in the tauri repo

knownasnaffy:
@FabianLars Thanks for the reply

lorenzolewis:
Thanks for this! I'm going to close this out in favour of #838 that we just merged. In general we also try to keep these guides very minimal (no CSS for instance) so that we can let the reader focus on just the bits specific to Tauri. 

Thanks again! <3

PR #851, opened 2022-09-03
Update mermaid

lorenzolewis:
@FabianLars you want to check this out and double-check the charts are rendering like they should? It looks like a big diff but it's because I had to turn part of the Docusaurus config into a function, so not really much changed.

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -213,166 +213,171 @@ const footerLinks = [
   },
 ]
 
-const siteConfig = {
-  title: 'Tauri Apps',
-  tagline:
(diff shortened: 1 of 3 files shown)


FabianLars:
Yep, output looks correct locally. No idea how to fix that build error tho (i see the same one locally on `yarn build`)

lorenzolewis:
I'll try a local bulls later tonight, they changed a lot of infra but I didn't think to test build before

FabianLars (approved):

FabianLars:
embarrassing that it took me so long to try upgrading it further 😂

PR #859, opened 2022-09-08
provide example error message for brew conflict, closes #674 

### LINKED ISSUE #674: FAQ: Brew and Potential Conflicts with Tauri
Installation issue because of Brew (see https://discordapp.com/channels/616186924390023171/987144563388997662/987298158692413440 for more details) to add the FAQ.

FabianLars:
(the index.js change was no accident)

### DIFF
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -61,7 +61,7 @@ function Cards() {
       label: translate({ message: 'Cross Platform' }),
       description: translate({
         message:
-          'Bundle binaries for all major desktop platforms (mobile & WASM coming soon).',
+          'Bundle binaries for all major desktop platforms (mobile coming soon).',
       }),
       link: '/guides/building/cross-platform',
       isDoc: true,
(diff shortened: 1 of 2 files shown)


src/pages/index.js
```diff
@@ -61,7 +61,7 @@ function Cards() {
       label: translate({ message: 'Cross Platform' }),
       description: translate({
         message:
-          'Bundle binaries for all major desktop platforms (mobile & WASM coming soon).',
+          'Bundle binaries for all major desktop platforms (mobile coming soon).',
```
lorenzolewis, line 64: Since this is a fairly high-visibility page, do we have a link to an issue or discord discussion we could maybe just throw as a comment on this PR for why we're removing it?

FabianLars: > do we have a link [...]

No and i didn't want to make a fuss about it to not cause any panics, because actually _nothing_ changes. We can't create tauri apps (with webviews) that run in wasm apps anyway, and many people interpreted this as "wasm support in the browser" which was always supported.
Also it was removed once before already: https://github.com/tauri-apps/tauri-docs/pull/168 and re-added accidentally(?) here https://github.com/tauri-apps/tauri-docs/pull/337

lorenzolewis: Nah, it was there before that in the old translation thingy, so was before that when it got re-introduced. But fine with #168 anyway.

lorenzolewis (approved):

PR #860, opened 2022-09-08
Fix recipes pages

olivierlemasle:
This PR fixes the "recipes" pages. These pages are still marked as "draft" and may be outdated or obsolete.

These pages should be either deleted or completed and published.

- Fix broken link
- Fix Mermaid diagrams
- Fix a "rater" color

### DIFF
--- a/docs/references/architecture/recipes/about-recipes.md
+++ b/docs/references/architecture/recipes/about-recipes.md
@@ -7,6 +7,6 @@ draft: true
 
 Tauri recipes are descriptions of use-cases that are entirely configurable within the `src-tauri/tauri.conf.json` file. These are not the limits of what Tauri can do, and there are probably more out there. If you discover one, why not get in touch and help us update this collection!
 
-If you haven't read about the general design of Tauri, then it would make the most sense for you to visit the ["Quick Start"][quick start] and become familiar with the basic architecture and terminology used in these recipes.
+If you haven't read about the general design of Tauri, then it would make the most sense for you to visit the [Quick Start] guide and become familiar with the basic architecture and terminology used in these recipes.
 
 [quick start]: ../../../guides/getting-started/setup/README.mdx
--- a/docs/references/architecture/recipes/about-recipes.md
+++ b/docs/references/architecture/recipes/about-recipes.md
@@ -7,4 +7,6 @@ draft: true
 
 Tauri recipes are descriptions of use-cases that are entirely configurable within the `src-tauri/tauri.conf.json` file. These are not the limits of what Tauri can do, and there are probably more out there. If you discover one, why not get in touch and help us update this collection!
 
-If you haven't read about the general design of Tauri, then it would make the most sense for you to visit the ["Getting Started"](../../getting-started/beginning-tutorial) and become familiar with the basic architecture and terminology used in these recipes.
+If you haven't read about the general design of Tauri, then it would make the most sense for you to visit the ["Quick Start"][quick start] and become familiar with the basic architecture and terminology used in these recipes.
+
+[quick start]: ../../../guides/getting-started/setup/README.mdx
--- a/docs/references/architecture/recipes/bridge.md
+++ b/docs/references/architecture/recipes/bridge.md
@@ -50,23 +50,21 @@ The Bridge recipe is a secure pattern where messages are passed between brokers 
 
 ## Diagram
 
-import { colors } from '@theme/Mermaid'
-
 ```mermaid
 graph TD
   H==>F
   subgraph WEBVIEW
-  F-.-E
+    F-.-E
   end
   D-->E
   E-->D
   B-->D
   D-->B
   subgraph RUST
-  A==>H
-  A-->B
-  B-.-C
-  B-.-G
+    A==>H
+    A-->B
+    B-.-C
+    B-.-G
   end
   A[Binary]
   B{Rust Broker}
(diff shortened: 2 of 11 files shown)


lorenzolewis:
Hey there! Thanks for the PR! 

We have these marked as drafts because they're from the early days before the 1.0 release and the feature wasn't fully fleshed out. 

I don't have an issue with merging something like this in so we can use it in the future (plus you've already put in the work ;) ).

I'll go through and put in a few comments on a few details in a review.

docs/references/architecture/recipes/about-recipes.md
```diff
@@ -7,4 +7,4 @@ draft: true
 
 Tauri recipes are descriptions of use-cases that are entirely configurable within the `src-tauri/tauri.conf.json` file. These are not the limits of what Tauri can do, and there are probably more out there. If you discover one, why not get in touch and help us update this collection!
 
-If you haven't read about the general design of Tauri, then it would make the most sense for you to visit the ["Getting Started"](../../getting-started/beginning-tutorial) and become familiar with the basic architecture and terminology used in these recipes.
+If you haven't read about the general design of Tauri, then it would make the most sense for you to visit the ["Quick Start"](../../../guides/getting-started/setup/) and become familiar with the basic architecture and terminology used in these recipes.
```
lorenzolewis: If you could update this to use reference-style links that would be brilliant! You can find some more details in the first point here: https://github.com/tauri-apps/tauri-docs/blob/dev/.github/CONTRIBUTING.md#style

For links going to another internal (markdown) page, the link should end in the `.md` file extension so that broken-link detection is build-in (so it would be `../../../guides/getting-started/setup.md` instead of `../../../guides/getting-started/setup/`)

docs/references/architecture/recipes/bridge.md
```diff
@@ -52,8 +52,9 @@ The Bridge recipe is a secure pattern where messages are passed between brokers
 
 import { colors } from '@theme/Mermaid'
 
-```mermaid
-graph TD
+<!-- prettier-ignore-start -->
+
+<Mermaid chart={`graph TD
```
lorenzolewis: Across all of these, we prefer to keep the old way as it allows the graph to be viewed outside of the rendered site (such as on GitHub). You can see an example of this here: https://github.com/tauri-apps/tauri-docs/tree/dev/docs/references/architecture/inter-process-communication

docs/references/architecture/recipes/bridge.md
```diff
@@ -79,7 +80,9 @@ graph TD
   style D fill:#ccc,stroke:#333,stroke-width:4px,color:white
```
lorenzolewis: One thing I learned during a Mermaid dependency update is that we can write a custom theme and can pull out any of these Style line you see here into that theme. I think this would also allow these charts to be rendered in the GitHub UI similar to the one I linked above. 

Not at all concerned about that for this PR, but if that's something you'd be interested in helping out with then feel free to give a shout!

lorenzolewis (changes requested):

docs/references/architecture/recipes/bridge.md
olivierlemasle: Ah, yes, it is rendered in GitHub; I did not see that.
The issue is that it is not rendered by Docusaurus... (when using `yarn && yarn start`, the recipes pages are rendered without the diagrams) :disappointed: 

For the `<Mermaid chart={...} />` syntax, I took it from https://github.com/tauri-apps/tauri-docs/blob/f743e35c3436d953485c5a8932d93d957c0def51/docs/references/architecture/readme.md?plain=1.

olivierlemasle: Oh, but actually it works for `references/architecture/inter-process-communication/` so of course it should also work here. I guess there's another issue. I'll check it.

lorenzolewis: I just checked, and I think the issue is caused by the `style` lines. If I remove them then they render correctly. I assume they only work if they're in a `<Mermaid />` component (which is probably why Architecture/readme has that instead of the code block format).

But regardless, those docs were created a while ago which is why they are formatted like that, the preferred way would be to use code block syntax so that we get that nice GitHub rendering.

docs/references/architecture/recipes/bridge.md
olivierlemasle: Actually, I've just found out that it's these `style` lines that break Docusaurus rendering.
With them, the  ` ```mermaid` syntax works on GitHub but not with Docusaurus.
Without them, both work (but with no styling). I'll look at those themes.

olivierlemasle: (sorry, I did not see your other comment before !)

docs/references/architecture/recipes/about-recipes.md
olivierlemasle: Ok, done

docs/references/architecture/recipes/bridge.md
olivierlemasle: I reverted to the previous syntax, and used classes to style the elements. It is rendered with the proper styles with Docusaurus, and degrades gracefully on GitHub.

olivierlemasle:
I've updated the PR regarding your comments.
Diagrams are now rendered with the proper styles within Docusaurus, and degrade gracefully on GitHub (work without custom colors).

docs/references/architecture/recipes/about-recipes.md
```diff
@@ -7,4 +7,6 @@ draft: true
 
 Tauri recipes are descriptions of use-cases that are entirely configurable within the `src-tauri/tauri.conf.json` file. These are not the limits of what Tauri can do, and there are probably more out there. If you discover one, why not get in touch and help us update this collection!
 
-If you haven't read about the general design of Tauri, then it would make the most sense for you to visit the ["Getting Started"](../../getting-started/beginning-tutorial) and become familiar with the basic architecture and terminology used in these recipes.
+If you haven't read about the general design of Tauri, then it would make the most sense for you to visit the ["Quick Start"][quick start] and become familiar with the basic architecture and terminology used in these recipes.
```
lorenzolewis:
```suggestion
If you haven't read about the general design of Tauri, then it would make the most sense for you to visit the [Quick Start] guide and become familiar with the basic architecture and terminology used in these recipes.
```

lorenzolewis (approved):

Issue #865, opened 2022-09-10
[design] Searchbar too wide for navbar at 1366px wide display


electrolyte-orb:
## You can see here
at 1366px wide display(1080p scaled up by ~150%), the navbar becomes inaccessible:

![Tauri Issue](https://user-images.githubusercontent.com/55880176/189470834-cc26d89a-764a-49c6-85ed-f7306d7a7d41.png)

lorenzolewis:
Closed by #866

PR #866, opened 2022-09-10
Fixed Navbar overlapping due to wide searchbar

electrolyte-orb:
Kindly refer to: #865

### DIFF
--- a/src/theme/SearchBar/style.css
+++ b/src/theme/SearchBar/style.css
@@ -21,7 +21,7 @@ div[data-ds-theme='dark']
 }
 
 /* Media Queries for specific width screen */
-@media only screen and (max-width: 1185px) {
+@media only screen and (max-width: 1400px) {
   .searchbox {
     width: 200px !important;
   }


lorenzolewis:
Thanks for this! There's still some issues when you go a bit smaller, but I think that's an issue with the searchbar as well as the trailing navbar items since they have added icons. I'll go ahead and merge this in, thanks a lot!

lorenzolewis (approved):

electrolyte-orb:
# This is just too much!
image

Mate, you can see there are so many elements for just a Search function(that is controlled directly by JS) and you can also see that each child `div` (which is totally irrelevent if JS is used correctly) has its own dimension and CSS styles, I dont understand why? For most of time they are invisible.

The search results are rendered into `<span class="dsb-dropdown-menu dsb-with-1"...`, means why are they not using divs for results and using divs for uneccessary bloating of styles?

#### I originally planned to make this issue of its own, but at this point the code is so complex that I cannot fix it. Maybe someone smarter will :)

lorenzolewis:
The searchbar was pulled in from `docs-searchbar` so most of that is directly from that project. The whole component is on the radar to update and simplify.

PR #869, opened 2022-09-10
Update JS Doc Generation

lorenzolewis:
Switches over to the very alpha-y branch we're maintaining of https://github.com/tgreyuk/typedoc-plugin-markdown.

# Changes

## Add `js-doc-generator` package
- By using this here we're able to control how the docs render without needing to commit to the main tauri repo
- `typedoc.json` config in this package points to the `tsconfig.json` and CLI entry point files relatively (assumes that the `tauri` and `tauri-docs` repos are next to each other on disk)

## Docusaurus Updates
- Replace latest 1.0.5 docs with a new structure
- Update internal sidebar and navbars based on new structure
- Add custom CSS for blockquotes to help them standout

## CI/CD (`update.docs.yml`)
- Remove old, commented out code for `tauri-search-bot` deployment (old name for `tauri-discord-bot`), meilisearch indexing command, and version.txt that is no longer used, 
- Now checks out and builds our fork of https://github.com/tgreyuk/typedoc-plugin-markdown
- Update docs generation to use new package instead of using the Tauri repo (this will allow us to remove some dependencies from the Tauri repo if we'd like)

# Open questions

- [ ] Is the `js-doc-generator` package set up as it should be? @amrbashir 
- [ ] Do we need to set up netlify redirects for API docs? @FabianLars


amrbashir:
wait until I make some popcorn 😈

lorenzolewis:
Let's make a bet on how many commits this will take

FabianLars:
Never was this happy that i had to leave my house :joy: have fun :v:

amrbashir:
> Let's make a bet on how many commits this will take

I never bet but say more than 30 commits

lorenzolewis:
Challenge accepted

lorenzolewis:
Only up to 6 commits so far 😅

lorenzolewis:
It's alive and it builds!!! After hours in npm dep land because how do you even update dependencies without renovate?!?

FabianLars:
> Do we need to set up netlify redirects for API docs? @FabianLars

Yes. The `js/modules/` -> `js/` one is a must and in theory the subpage merges too, ~~but i don't see a way to link to the correct anchor so idk if we should bother at all.~~ Edit: Yes we should, a "inaccurate" redirect is better than a 404.

amrbashir (approved):

lorenzolewis:
There we go @FabianLars I splatted for modules, but the interfaces, enums, and classes have different urls so they just redirect back to the JS API folder root.

netlify.toml
```diff
@@ -78,3 +78,19 @@ to = "/v1/references/benchmarks"
 [[redirects]]
 from = "/references/*"
 to = "/v1/references/:splat"
+
+[[redirects]]
```
FabianLars, line 82:
```suggestion
[[redirects]]
from = "/v1/api/js/classes/window.webviewwindow/*"
to = "/v1/api/js/window"

[[redirects]]
```

That's one of the most linked js pages so we should consider adding a specific redirect at least for that one.

FabianLars:
idk how easy it would be to change that, but i think the headers are way too small here 
grafik

lorenzolewis:
Which header @FabianLars ? There are so many 😅

FabianLars:
The isDecorated one (and isFullscreen etc). It's not clear enough that it's the start of a new section if that makes sense?

lorenzolewis:
I set the font size to not go below an h4 now. It gets tricky because it's going down into the h5 level which is accurate, but yay hierarchy. 

Screenshot 2022-09-13 at 16 11 40

Also cleaned up a bunch of old CSS while I was at it (a lot of them were for the old config page before it was reworked).

lorenzolewis:
Here's what it looks like @FabianLars 

Screenshot 2022-09-13 at 16 46 47

FabianLars:
Yeah that works. i have a feeling tho that we end up using horizontal lines sooner or later x)

PR #874, opened 2022-09-11
Feat/search and navbar

lorenzolewis:
- Remove the search item from navbar (this doesn't stop it from being rendered, just moves it to the very end of the navbar)
- Switch to using built-in Docusaurus theming for the search bar input
  - This allows us to remove the `useColorMode` hook that was also causing a bit of flickering on first load or when changing colors
- Rebuild the search results CSS from the ground up
  - It should behave a lot better on mobile and at the very least we have a lot more precise control as we're no longer importing the vendor CSS from `docs-searchbar.js`
- Collapse trailing navbar text if screen size would cause wrapping

I want to tweak the CSS overtime on the search results, but I think the results we get back from the indexer need optimised first (the CSS isn't concatenating a lot of the values, they're concatenated when sent back from the indexer)

Before:
Screenshot 2022-09-11 at 21 45 44

After:
Screenshot 2022-09-11 at 21 45 09

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -134,10 +134,6 @@ const navbarItems = [
       },
     ],
   },
-  {
-    type: 'search',
-    position: 'right',
-  },
 ]
 
 const footerLinks = [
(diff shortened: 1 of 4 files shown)


FabianLars:
the only thing is that i think the breakpoints need to trigger a little earlier when shrinking the window so you don't see this stuff:
grafik
grafik

FabianLars:
Follow up on the above message: This is only a problem if i resize the window itself (firefox), if i resize the mobile emulator window (the devtools thingy) the breakpoints are correct :/

lorenzolewis:
I should have known better than to have it that close when it comes to dimensions. What are standard font and point sizes anyway?

FabianLars:
The new one is a littleee bit too early, but it's good enough to leave it like this. 👍 And the other breakpoints seems to be from docusaurus so the only other thing would be to remove some other element so that the logo and `guides` don't overlap, buuut yeah, let's just leave it. That 10px range should be small enough i guess 🤷

FabianLars (approved):

lorenzolewis:
Yeah, unfortunately the only things Docusaurus recommends are to remove navbar items to fix it 😅 https://github.com/facebook/docusaurus/issues/7635

RudraSen2:
@lorenzolewis, what browser are you using in the picture?

lorenzolewis:
Safari

PR #885, opened 2022-09-14
Add instructions to build for arm on windows

FabianLars:


lorenzolewis (approved):
Don't know anything about this from a technical perspective, but from a formatting POV it looks good

PR #886, opened 2022-09-14
fix: Don't use `app` in sidecar guide

FabianLars:
_"because."_ - Fabian-Lars, professional commit description writer, 2022

Btw the reason for this is that the default app name in Cargo.toml is `app` too which causes conflicts, see https://github.com/tauri-apps/tauri/issues/4780

### DIFF
--- a/docs/guides/building/sidecar.md
+++ b/docs/guides/building/sidecar.md
@@ -20,7 +20,7 @@ Here is a sample to illustrate the configuration. This is not a complete `tauri.
   "tauri": {
     "bundle": {
       "externalBin": [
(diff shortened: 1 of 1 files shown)


lorenzolewis:
Do you think this should be moved to `features` instead of `building` like it is? And then throw a link to it somewhere in the building page?

FabianLars:
no idea, i think we talked about that once before but someone didn't like it, can't remember. I wanted to talk about the building/distribution categories on discord later anyway

lorenzolewis:
sounds good

lorenzolewis (approved):

PR #888, opened 2022-09-14
Fix External link target, TAO/WRY caps, blog link

lorenzolewis:
CC @probablykasper

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -51,6 +51,7 @@ const navbarItems = [
       {
         label: 'Rust (via Docs.rs)',
         href: 'https://docs.rs/tauri/1/',
+        target: '_self',
       },
     ],
   },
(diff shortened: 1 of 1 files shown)


FabianLars (approved):
i hate the caps change btw but i guess it's the default

probablykasper (changes requested):
Looks like it's still showing those "open in new tab" icons

lorenzolewis:
Those I prefer to keep because it's at least a visual indicator that it's an external domain 🤷

probablykasper:
hmm, I feel like it might be confusing, but feel free to keep it if you think that's best, or if someone else agrees

amrbashir:
I agree to keep it, I really like the indicator for external sites (I think it is also good for accessibility), so I can use middle click instead of normal click.

lorenzolewis:
@probablykasper is that alright? (would you like to begrudgingly click the approve button?) <3

probablykasper (commented):
Yes 👍

probablykasper (approved):
Ok for real this time 🤦‍♂️

PR #893, opened 2022-09-14
Add guide about `resources`

FabianLars:


docs/guides/building/resources.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+## Accessing files in JavaScript
+
+In this example we want to bundle additional i18n json files that look like this:
+
+```json title=de.json
+{
+  "hello": "Guten Tag!",
+  "bye": "Auf Wiedersehen!"
```
lorenzolewis, line 49: ❤️🇩🇪

FabianLars: took me way to long to come up with an example...

thegame4craft:
how is the file & folder structure in `Embedding additional files`?
is it like
```
src/
    ... (frontent-files)
src-tauri/
    lang/
        de.json
    main.rs
```
(shorten)

FabianLars:
yes, but it doesn't matter as long as you change the paths in the scope and resolveResource accordingly.

thegame4craft:
ok because i get errors like
```
resources.ts:16 
        
    Uncaught (in promise) path not allowed on the configured scope: \\?\C:\Users\<USERNAME>\projects\TauriApp2\tauri-app\src-tauri\target\debug\locales\language.json
```

FabianLars:
Ah yes, i copied the wrong scope config... Gonna update that asap, thank you! as a quick fix just change it to `"scope": ["$RESOURCE/*"]`

FabianLars:
Merging due to popular demand. We can fix any spelling mistakes etc in a follow-up PR 🤷

PR #894, opened 2022-09-15
Add info about config plugins and scripts

NathanaelRea:


docs/guides/getting-started/setup/vite.mdx
```diff
@@ -94,6 +94,12 @@ export default defineConfig({
 })
 ```
 
+:::info
+
+Note that if you are not using vanilla JavaScript, you should modify this config and add the necessary plugins your framework requires.
```
FabianLars: @NathanaelRea What do you think about something like this, since create-vite's templates already contain the plugins?
```
Note that if you are not using vanilla JavaScript, you must keep the framework-specific plugins that were already in this file.
```
cc @lorenzolewis because i'm not happy with my own suggestion either 😂

docs/guides/getting-started/setup/vite.mdx
```diff
@@ -127,6 +133,14 @@ Now that we have scaffolded our frontend and initialized the Rust project the cr
...
 ```
 
+Be sure that tauri is accessible to your package manager of choice. For example if you are using npm, the `package.json` file should include tauri as a script:
+```json title=package.json
+{
+  "scripts": {
+    "tauri": "tauri"
+  },
+```
+
```
FabianLars: Writing this warning here would be too late since `tauri init` needs it too. I think it should be part of the `Install Tauri CLI` tabber instead.

FabianLars (changes requested):

docs/guides/getting-started/setup/vite.mdx
NathanaelRea: I mainly wanted to make it clear that the vite.config.ts given can be modified. It says "update the file with the following", but then it gives the entire file, which might lead the user to just overwrite their file with the one provided.

docs/guides/getting-started/setup/vite.mdx
NathanaelRea: Agreed. I just wanted to make sure the user didn't think they needed to install tauri as a dependency or dev-dependency.

docs/guides/getting-started/setup/vite.mdx
FabianLars: Yeah we're on the same page

FabianLars (approved):
Sorry for the delay! Thanks for contributing :)

PR #899, opened 2022-09-15
feat(blog): add 1.1.0 announcement post

lucasfernog:

PR #908, opened 2022-09-19
chore(blog): add tauri-egui 0.1 announcement post

lucasfernog:


blog/2022-09-19-tauri-egui-0-1.mdx
```diff
@@ -0,0 +1,267 @@
...
+tauri-egui = "0.1"
+```
+
+Now you need to enable the plugin:
+
+```rust
+fn main() {
+  tauri::Builder::default()
+    .setup(|app| {
+      app.wry_plugin(tauri_egui::EguiPluginBuilder::new(app.handle()));
```
amrbashir, line 30: this is not relevant to the blog post but why is it named `wry_plugin` when it got nothing to do with `wry`?

amrbashir (approved):

blog/2022-09-19-tauri-egui-0-1.mdx
lucasfernog: it's a low level plugin (event loop) to the `tauri-runtime-wry` crate. `wry` exposes `tao`, so there's something to do with wry 😂 though i don't see any good name for this and too late to change it now.

amrbashir: `event_loop_plugin` would be a good name but yeah too late now.

FabianLars:
Do we want to update the banner at the top too?

lucasfernog:
> Do we want to update the banner at the top too?

Too soon? Is there a good way to keep both announcements?

FabianLars:
> Is there a good way to keep both announcements?

Probably not, [we're not even that much of a fan of a single one](https://github.com/tauri-apps/tauri-docs/pull/899#issuecomment-1248648317) 😅 

> Too soon?

Well, it would be ~3 days earlier than we had in mind, but at least the hero banner on the frontpage still shows 1.1.
idk, it's your call, we can change it later or even not at all 🤷

FabianLars:
ah well, we could do something like "announcing the release of tauri 1.1 and tauri-egui 0.1"

lucasfernog:
> ah well, we could do something like "announcing the release of tauri 1.1 and tauri-egui 0.1"

changed it. What do you think?

FabianLars (approved):

PR #912, opened 2022-09-22
fix: truncate blog post summaries on blog overview page

FabianLars:


amrbashir:
this post should also be truncated https://tauri.app/blog/2022/07/12/tauri-programme-turns-1-and-board-elections

FabianLars:
i didn't find a good break point and hoped that it'd be short enough. i guess i'll just cut it at the first line break? (before the `The Tauri Board has a minimum of three` line)

in the context of summary length i thougt about including the # What is Tauri section in the 1.0 summary, but that would result in the same length as that one so nevermind.

amrbashir (approved):

PR #919, opened 2022-09-25
Mention prevent close API in docs

joepio:
Adds a section to the `system-tray` feature page to describe how you can have an app that keeps running in the background.

### DIFF
--- a/docs/guides/features/system-tray.md
+++ b/docs/guides/features/system-tray.md
@@ -201,8 +201,8 @@ If your app should run in the backgorund, you can call `api.prevent_close()` lik
 ```rust
 tauri::Builder::default()
     .on_window_event(|event| if let tauri::RunEvent::ExitRequested { api, .. } => {
-    api.prevent_exit();
-  })
+      api.prevent_exit();
+    })
 ```
 
 [template image]: https://developer.apple.com/documentation/appkit/nsimage/1520017-template?language=objc
--- a/docs/guides/features/system-tray.md
+++ b/docs/guides/features/system-tray.md
@@ -193,5 +193,18 @@ Note that you need to add `icon-ico` or `icon-png` feature flag to the tauri dep
 app.tray_handle().set_icon(tauri::Icon::Raw(include_bytes!("../path/to/myicon.ico").to_vec())).unwrap();
 ```
 
+### Keep app running in the background after closing all windows
+
+By default, tauri closes the application when the last window is closed.
+If your app should run in the backgorund, you can call `api.prevent_close()` like so:
+
+```rust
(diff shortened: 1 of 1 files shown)


docs/guides/features/system-tray.md
```diff
@@ -193,5 +193,18 @@ Note that you need to add `icon-ico` or `icon-png` feature flag to the tauri dep
...
+
+By default, tauri closes the application when the last window is closed.
+If your app should run in the backgorund, you can call `api.prevent_close()` like so:
+
+```rust
+tauri::Builder::default()
+    .on_window_event(|event| if let tauri::WindowEvent::CloseRequested { api, .. } = event.event() {
+        event.window().hide().unwrap();
+        api.prevent_close();
+    })
```
amrbashir: this hidesthe window or prevents closing the window where if you want to keep the app running without any windows, then a better example would be the example here https://docs.rs/tauri/latest/tauri/struct.App.html#method.run

amrbashir (changes requested):
#

docs/guides/features/system-tray.md
joepio: I agree, we can make it more minimal. I've updated the PR

amrbashir: It is not about how minimal the example. The example you are using prevents the **window** from being closed, what you want instead is to prevent the **app** from being closed which is what the example I linked above shows.

joepio: Whoops, you're very right, sorry. Should be good now.

docs/guides/features/system-tray.md
```diff
@@ -193,5 +193,17 @@ Note that you need to add `icon-ico` or `icon-png` feature flag to the tauri dep
...
+### Keep app running in the background after closing all windows
+
+By default, tauri closes the application when the last window is closed.
+If your app should run in the backgorund, you can call `api.prevent_close()` like so:
+
+```rust
+tauri::Builder::default()
+    .on_window_event(|event| if let tauri::RunEvent::ExitRequested { api, .. } => {
+      api.prevent_exit();
+    })
```
amrbashir: RunEvent is not accessible in `on_window_event` hook, it should be

```suggestion
tauri::Builder::default()
  .build(tauri::generate_context!())
  .expect("error while building tauri application")
  .run(|_app_handle, event| match event {
    tauri::RunEvent::ExitRequested { api, .. } => {
      api.prevent_exit();
    }
    _ => {}
  });
```

amrbashir (changes requested):

amrbashir (approved):

PR #920, opened 2022-09-27
fix: correct links to frameworks on /integrate page

lars-berger:


FabianLars (approved):
How that went live is beyond me lol. Thank you!

PR #927, opened 2022-10-01
add note regarding my experience with EV+OV certs

goenning:
This is how it looks.

Screen Shot 2022-10-01 at 12 01 34


FabianLars (approved):
Thank you

PR #929, opened 2022-10-01
use tabs for package manager commands on updating-dependencies page

NyCodeGHG:
I updated the "Updating Dependencies" Page to use Tabs for the different javascript package manager commands and I also added pnpm equivalents besides the npm and yarn commands.


FabianLars (approved):
thank you

PR #930, opened 2022-10-02
Add optimization tip for "compression" feature

mon:
Until cargo allows us to selectively remove a default feature, this doc will unfortunately have to be kept up to date with whatever features get added to the `tauri` crate.

With every optimization in this document (and this new one), I got the svelte template app down to 1.3MiB. An extra ~150KiB can be saved by compiling for i686 instead of x86_64, but I don't think it's worth documenting.

Still a long way from the 600KiB on the homepage (UPX gets it to 500KiB but Windows AV raises a false positive), but acceptable for now!


FabianLars (approved):

FabianLars:
Thank you and sorry for the delay.

PR #937, opened 2022-10-08
Updated return type for 'exists' in js fs docs

Kimbo0o:
This PR updates the return type of the `exists` function within `/api/js/fs`.
It should be boolean instead of void.

The api was already fixed in [API Pull Request](https://github.com/tauri-apps/tauri/pull/5252/files), but the documentation wasn't adjusted yet.

### DIFF
--- a/docs/api/js/fs.md
+++ b/docs/api/js/fs.md
@@ -283,7 +283,7 @@ A promise indicating the success or failure of the operation.
 
 ### `exists`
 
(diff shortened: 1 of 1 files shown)


FabianLars:
> The api was already fixed in [API Pull Request](https://github.com/tauri-apps/tauri/pull/5252/files), but the documentation wasn't adjusted yet.

Yeah because the fix wasn't released yet. Once it is the docs will be updated automatically. Thank you nonetheless!

PR #942, opened 2022-10-12
Add more headers to gain more readability

tiziodcaio:


FabianLars (changes requested):
This file is auto generated on each Tauri release by this file: https://github.com/tauri-apps/tauri-docs/blob/dev/.scripts/generate-config-doc.js taking the config from here: https://github.com/tauri-apps/tauri/blob/dev/tooling/cli/schema.json (this one is also auto generated from the rust source code)

So this PR would be overriden on the next release. I'm not sure how easy it would be to integrate these changes into the generator files but feel free to look into it if you want :)

Thanks for contributing nonetheless!

tiziodcaio:
I suspected damn... I see.... now I study if we can do something
Thanks

tiziodcaio:
I'm sorry the re-request was an error

FabianLars:
As far as i can see it's looking good so far! The only thing i'm not quite happy with is using :::note for the type, but i didn't like the prior design either... Do you have another idea? (we do have access to CSS here btw) If not then i'm be fine with merging it as-is for now, maybe someone else can think of something nice later 🤷

tiziodcaio:
> As far as i can see it's looking good so far!

Huge tanks!!

> The only thing i'm not quite happy with is using :::note for the type, but i didn't like the prior design either... Do you have another idea? (we do have access to CSS here btw) If not then i'm be fine with merging it as-is for now, maybe someone else can think of something nice later 🤷

Hmmm at the moment not, but if I arrange something I make a propose no problems

amrbashir:
The `:::note` is indeed a bit annoying and I'd prefer the old one if nothing better comes to mind.

tiziodcaio:
Ok no problem, now when i've time I will revert...

FabianLars (approved):
Let's merge it and iterate from there if needed. Thank you!
(Edit: Waiting for CI to finish)

PR #946, opened 2022-10-18
correct macOS 12. version

JonasKruckenberg:
The webkit version actually didn't change since 12.5.1 🤷🏻‍♂️


FabianLars:
@JonasKruckenberg 
The Safari version for macos 12.5.1 is wrong (i think)? What's `16.5.1` supposed to actually be? x)

JonasKruckenberg:
Okay let me check that

JonasKruckenberg:
Okay I just checked with a colleague, macOS 12.5.1 also has `613.3.9.1.16`

FabianLars:
And that's Safari 15.6 right?

JonasKruckenberg:
15.6.1 is what Safari reports on my coworkers machine, so I assume it's the same for 15.6

FabianLars:
Ah alrighty, so it's just a Zahlendreher, thanks :v:

FabianLars (approved):

PR #949, opened 2022-10-24
Fix announcement bar width, format html

hslee2008:
# Before (Width: 500px)

This is probably what the website looks like on 100% of all the phones out their.
Does not seem like there is any need for breaking in screen width 500px.

# After (Width is now set to 350px)

# HTML is not properly formatted
before

# After
after


FabianLars (approved):
thanks

PR #962, opened 2022-11-03
Explain about mocking IPC request to sidecar bin

yonathan06:
IPC requests to sidecar binaries behave differently then others
I needed to dig into the source code to find that it uses an event emitter of the spawned process, so to mock the behaviour we must send events to the event emitter to resolve the promise


FabianLars (approved):

FabianLars:
Thank you!

PR #985, opened 2022-11-15
fix broken reference

jukrb0x:
the ref to [minimumSystemVersion](https://tauri.app/v1/api/config/#macconfig.minimumsystemversion) is broken in the doc


FabianLars:
All the other links are broken too 😅 

Anyway, this file will be deleted in #989 and we have to batch-fix all these links in multiple files soon.

Thanks for contributing nonetheless!

PR #991, opened 2022-11-24
init v2

FabianLars:
let's see if it builds

Todo:
- js-api doc generation
- config doc generation
- check if `LATEST_VERSION` env var in build.yml is still needed


lorenzolewis:
I'm extremely impressed and proud of you @FabianLars for getting it to work in one build

FabianLars:
well, to be fair i didn't do anything, you set it all up back then :D

FabianLars:
Ah actually the translations don't seem to work anymore

lorenzolewis:
Translations only work on prod deploys, otherwise we were getting into build fails for broken links from translated documents. You move a file in English, but for a deploy preview we don't post any updated files to Crowdin so then we can't download any translated files. Sort of a catch 22 so went with just "faking" building other languages for the preview where it builds them but just from copied English versions instead of other languages. 

tldr: works as expected

FabianLars:
ah right, totally forgot that stuff. thanks :v:

docusaurus.config.js
```diff
@@ -2,19 +2,17 @@ const path = require('path')
...
-var lastestReleasedVersion
+const unreleasedTauriVersion = 'v2'
+let lastReleasedVersion
 
 // Checks if Docusaurus has been versioned before and sets versions accordingly
 try {
-  lastestReleasedVersion = JSON.parse(
-    fs.readFileSync('versions.json', 'utf-8')
-  )[0]
+  lastReleasedVersion = JSON.parse(fs.readFileSync('versions.json', 'utf-8'))[0]
```
lorenzolewis: Now that you'll actually have this `versions.json` file generated and committed, you can remove line 6 above and remove the conditionality on 309-312 below. They were just there since you don't have the versions.json file until you've first made a version cut

lorenzolewis:
For the JS and CLI configs, I think everything should work out of the box, you just need to point the `update-docs` action to the correct Tauri branch when it runs.

lorenzolewis:
And then lastly not sure if you want to keep the filepath for versioned docs to be `versioned-docs/version-v1` like it is now or prefer `versioned-docs/v1` instead

FabianLars:
> For the JS and CLI configs, I think everything should work out of the box, you just need to point the update-docs action to the correct Tauri branch when it runs.

Yeah but only for the `current` docs, not for v1 which will still receive updates before v2 is actually released.

> And then lastly not sure if you want to keep the filepath for versioned docs to be versioned-docs/version-v1 like it is now or prefer versioned-docs/v1 instead

I figured it'd be the easiest to go with the default but i can take a look if we can change it because it indeed looks ugly

lorenzolewis:
> Yeah but only for the current docs, not for v1 which will still receive updates before v2 is actually released.

Ah, didn't realise that was the default. 

For v1 updates post 2.0, how often and what types of updates would those be that would actually be API changes or things that need to update the guides for? I would think you actually don't need a live v1 that's built every time and I'd almost treat v1 as a pre-built archive at that point. Can always manually make a fix if needed, but I feel like making pipelines work for old versions _could_ be more trouble than it's worth.

FabianLars:
> For v1 updates post 2.0, how often and what types of updates would those be that would actually be API changes or things that need to update the guides for?

None. This is only for the time where v2 is in pre-release phase where v1 will still receive changes (which may or may not be quite some time).
For now we can just change the pipeline to work in v1 (or both) but it will get a bit more complicated if we actually release v2 pre-release packages so that v1 and v2 api docs will actually diverge

PR #995, opened 2022-11-28
More idiomatic NixOS setup

bailey-busc:
This is a simple change that removes the hand-rolled function in favor of just using [lib.makeLibraryPath](https://github.com/NixOS/nixpkgs/blob/6949cecbd6b3b9e78ae91b8b97dc001075378abc/lib/strings.nix#L177), which accomplishes the same thing.


FabianLars (approved):

FabianLars:
thank you :)

PR #1001, opened 2022-12-02
feat: add mobile documentation

lucasfernog:


FabianLars:
btw, feel free to merge _whenever_ you want so that the preview at <https://next--tauri.netlify.app/> is updated.
Or maybe we should change the approach a bit again so that PRs do get a preview too 🤔

lorenzolewis:
We aren't linking out to next on the main site right now, so I'd say merge in whatever WIP you have. But we could also activate deploy previews against the `next` branch as well if you'd like.

112RG:
Thanks for writing docs for mobile.

I have some additions that might be helpful such as
When using Vite you need type: "module" in your package.json to use the latest version of internal-ip. Also another note is that you cant use TailwindCSS with ESM. I would suggest using internal-ip 6.2.0 since this was before sindresorhus converted all other npm packages to ESM

PR #1027, opened 2022-12-13
Move mobile bundler from roadmap to features

mrquantumoff:
Since in Tauri v2 there is mobile bundling support, I think it's time to move "mobile bundler" from roadmap to features


FabianLars:
Honestly i think it's too early. We just released the very first alpha and it still has many issues and many users are not able to successfully use it yet, so i feel like listing it as a feature already would be a bit dishonest.

PR #1028, opened 2022-12-14
Add version dropdown with a link to the alpha docs

FabianLars:


lorenzolewis:
@FabianLars how do you feel about doing this in sections similar to Docusaurus's site:

Screenshot 2022-12-22 at 17 26 58

Have the top section be labeled "Stable" with 'v1' and the bottom section be "Alpha" with 'v2'?

https://docusaurus.io/docs/api/themes/configuration#navbar-docs-version-dropdown

FabianLars:
Yeah had something similar in mind at first but it was a bit awkward with just 2 entries, i'll push your version so we can see it live

lorenzolewis:
I'm honestly not too worried about it because by the time v2 is out hopefully we have a whole new site and it will all be completely different

lorenzolewis:
@FabianLars how about this?

Screenshot 2022-12-22 at 18 29 56

FabianLars:
Sure, works for me

lorenzolewis (approved):

PR #1047, opened 2022-12-25
Update sveltekit output directory

MCMattia:
The sveltekit output directory that needs to be configured when using `pnpm tauri init` is not `../build`. It is instead `../svelte-kit`. You can also check it here in the sveltekit [documentation](https://kit.svelte.dev/docs/project-structure#other-files).

Thank you guys for the awasome work with tauri !


MCMattia:
I was wrong, the svelte adapter-static creates the build directory. Ignore this pull request please

PR #1052, opened 2023-01-02
add debugging guide for intellij clion

### LINKED ISSUE #1053: [docs] Method for debugging in IntelliJ
I see in the documents there some method to debug with breakpoints in VSCode, however, I do not see anything regarding IntelliJ or CLion. Is there a well known method for running tauri in debug mode within IntelliJ using their configurations? I am at a point now where this is greatly needed over println everywhere.

JonasKruckenberg:
closes #1053


lorenzolewis:
Can we start to put assets under `/static`? This is how Astro will need to handle it and that will help me resolve deltas a lot easier later :)

JonasKruckenberg:
Sure thing

docs/guides/debugging/clion.md
```diff
@@ -0,0 +1,51 @@
+# Debugging in CLion
+
+In this guide, we’ll be setting up IntelliJ CLion for debugging the [Core Process of your Tauri app].
+
+## Prerequisites
+
+You need to install the [IntelliJ Rust Plugin](https://plugins.jetbrains.com/plugin/8182-rust/docs) that Rust features are enabled.
```
FabianLars: > that Rust features are enabled.

Something is missing and i have no idea what 😅

JonasKruckenberg:
@lorenzolewis I'm done with this PR now, also moved some more images to the static folder. Though I think we might want to revisit this after the astro v1 version bc I would really like to build optimised versions of images, so `static` isn't really applicable then anymore

docs/guides/debugging/clion.md
```diff
@@ -0,0 +1,51 @@
...
+## Top-level Cargo Workspace
+
+By default Tauri places the Rust project in a subdirectory called `src-tauri`. CLion doesn't recognize Cargo projects that are not at the top-level like this, so we need to create a top-level Workspace that just points to our main `Cargo.toml` file: 
+
+```toml title=Cargo.toml
+[workspace]
+members = ["src-tauri"]
+```
+
+Before you proceeed, make sure that your project is fully loaded. If the indexing is finished and the Cargo tool window shows all the modules and targets of the workspace, you’re good to go.
```
FabianLars: This is not true. Arguably it may be the easier solution but you can also tell CLion where the rust project is. For example if you open the main.rs file it should tell you that there is no Cargo project with a button saying something like "Attach project".

docs/guides/debugging/clion.md
```diff
@@ -0,0 +1,51 @@
...
+
+```sh
+pnpm vite dev
+```
+
+> Note that CLion currently has no support for background tasks in Run/Debug configurations like VS Code has, so for now you need to manually run the development server.
+
+## Launch a Debugging Session
+
+With the development server running and the Run/Debug Configuration selected in the Switcher, you can now start a new debugging session by clicking Debug `Ctrl+D` on Windows/Linux, `⌃D` on macOS). CLion will automatically recognize breakpoints placed in any Rust file in your project.
```
FabianLars: The shortcuts seem to be wrong, for me it lists these:
grafik

Edit: In the editor window itself there is no shortcut at all: 
grafik

FabianLars:
Also another thing. On Windows i get this error:
```
Unable to run debugger

GNU debugger cannot be used with MSVC Rust toolchain.
```

Can be fixed by changing the default debugger to Visual Studio, which was added quite recently to CLion as it seems.
Can't seem to find a way to switch it to the gnu rust toolchain without changing it globally 🤔 I thought maybe we can make the debugger use the gnu toolchain since clion bundles all the necessary gcc stuff anyway (setting up mingw or cygwin is quite annoying imo)

docs/guides/debugging/clion.md
FabianLars: Hmm, maybe nevermind. the debugger cries about missing Cargo.toml if we don't set the working directory to the src-tauri sub dir

JonasKruckenberg: Yeah but even then, when I tested it I couldn't attach that nested folder at all. It just wouldn't recognise it

docs/guides/debugging/clion.md
JonasKruckenberg: I took these from their official docs 🤔 maybe it's a weird Germany specific keymap? When you go to settings > keymap what does it say your keymap is?

( for me it says "macOS" for example, but CLion has a number of presets that can emulate an editor basically)

JonasKruckenberg:
> Also another thing. On Windows i get this error:
> 
> ```
> Unable to run debugger
> 
> GNU debugger cannot be used with MSVC Rust toolchain.
> ```
> 
> Can be fixed by changing the default debugger to Visual Studio, which was added quite recently to CLion as it seems. Can't seem to find a way to switch it to the gnu rust toolchain without changing it globally 🤔

As per [this blog post](https://blog.jetbrains.com/clion/2019/10/debugging-rust-code-in-clion/#switching_debuggers_and_renderers) it seems that they only support gnu targets for debugging on windows which would be a pity. 

I have to way to check this however so I would appreciate if you could maybe add a quick paragraph explaining this for windows users

docs/guides/debugging/clion.md
JonasKruckenberg: yeah fixed it!

docs/guides/debugging/clion.md
FabianLars: you need to either select the src-tauri dir or the Cargo.toml file. It works 100% with the plugin, but for this use-case it'd require us the use src-tauri as the working dir for the Run command so idk if it's worth.

docs/guides/debugging/clion.md
FabianLars: Keymap: `Windows` (there's no macos or linux option so i can't check if it's Windows specific)

JonasKruckenberg: Yeah I think this is a windows specific setting then, maybe their docs are outdated

FabianLars:
Yeah i saw that blog post too but they added VS support in 2022 or something and it seems to work. I can't compare the output to the gnu debugger because i don't want to install the latter :P 
The gnu debugger is always the default tho, and i can't find i way to set it per run/debug config, only project-wide 🤔

JonasKruckenberg:
> The gnu debugger is always the default tho, and i can't find i way to set it per run/debug config, only project-wide 🤔

Project-wide is completely fine though IMO

docs/guides/debugging/clion.md
FabianLars: Maybe it's better to focus on the Button (for the mouse) for the guide then? Since it shows the shortcut people can connect the dots themselves.

JonasKruckenberg: Yeah let's just remove the shortcuts
```suggestion
With the development server running and the Run/Debug Configuration selected in the Switcher, you can now start a new debugging session by clicking Debug. CLion will automatically recognize breakpoints placed in any Rust file in your project.
```

lorenzolewis (approved):

PR #1057, opened 2023-01-04
Improve allowlist docs

matthme:
The allowlist documentation has been confusing for me so I extended it.
It is based on my current understanding of it and its correctness needs to be confirmed by people actually understanding it in case this PR is considered useful.

### DIFF
--- a/docs/api/config.md
+++ b/docs/api/config.md
@@ -390,7 +390,9 @@ Type: `object`
 ### AllowlistConfig
 
 Allowlist configuration. The allowlist is a translation of the [Cargo allowlist features](https://docs.rs/tauri/latest/tauri/#cargo-allowlist-features).<br>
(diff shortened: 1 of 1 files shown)


FabianLars:
I really like this change. The only problem here is that it's auto generated from here https://github.com/tauri-apps/tauri/blob/dev/core/tauri-utils/src/config.rs#L2143 - i'd appreciate if you could open a PR against the main repo too 🙏

FabianLars:
p.s. all commits must be signed for us to be able to merge them: https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits - i _could_ merge it in tauri-docs regardless but not in the main repo.

matthme:
Okay, makes sense. I'll make a (signed) PR to the main repo then.

matthme:
https://github.com/tauri-apps/tauri/pull/5985

FabianLars (approved):

PR #1060, opened 2023-01-08
Add extra prerequisite in order to develop under windows

JurrianFahner:
I'm a newbie on Tauri development. And I was trying to get started, but got stuck. I used the powershell method. 

The default toolchain I had configured (gnu) is not compatible with the getting started guide. However it is [assumed that the msvc toolchain is set as default](https://github.com/tauri-apps/tauri/issues/3953#issuecomment-1107807436). When I did configure the msvc toolchain, the problems were resolved. 

Maybe it also needs to be added in the powershell script. Where it is mentioned that it is important to install `tauri-cli` and `trunk`.

### DIFF
--- a/docs/guides/getting-started/prerequisites.md
+++ b/docs/guides/getting-started/prerequisites.md
@@ -58,6 +58,7 @@ rustup default stable-msvc
 ```
 
 :::
+
 ### Setting Up macOS
 
 #### 1. CLang and macOS Development Dependencies
@@ -314,7 +315,8 @@ If you don't see this information, your Rust installation might be broken. Pleas
 [github discussions]: https://github.com/tauri-apps/tauri/discussions
 [download webview2]: https://developer.microsoft.com/en-us/microsoft-edge/webview2/#download-section
 [rustup.sh]: https://sh.rustup.rs
-[Nix Flakes]: https://nixos.wiki/wiki/Flakes
-[direnv's Flakes integration]: https://nixos.wiki/wiki/Flakes#Direnv_integration
-[Nix Shell]: https://nixos.wiki/wiki/Development_environment_with_nix-shell
-[direnv's Shell integration]: https://nixos.wiki/wiki/Development_environment_with_nix-shell#direnv
+[nix flakes]: https://nixos.wiki/wiki/Flakes
+[direnv's flakes integration]: https://nixos.wiki/wiki/Flakes#Direnv_integration
+[nix shell]: https://nixos.wiki/wiki/Development_environment_with_nix-shell
+[direnv's shell integration]: https://nixos.wiki/wiki/Development_environment_with_nix-shell#direnv
+[`trunk`]: https://trunkrs.dev
--- a/docs/guides/getting-started/prerequisites.md
+++ b/docs/guides/getting-started/prerequisites.md
@@ -49,7 +49,17 @@ winget install --id Rustlang.Rustup
 
 #### 3. MSVC toolchain as default
 
-In order to be able to install tools like `tauri-cli` and `trunk` you need to make sure that the proper toolchain is set as default:
+In order to be able to install tools like `tauri-cli` and `trunk` you need to make sure that the proper toolchain is set as default.
+
+To display the current configured default toolchain please run the following command:
+
+```powershell
+rustup default
+```
+
+The command should return either `i686-pc-windows-msvc`, `x86_64-pc-windows-msvc`, or `aarch64-pc-windows-msvc`.
+
+If the command returns a different toolchain, please configure the default toolchain in the following way: 
 
 ```powershell
 rustup default stable-msvc
--- a/docs/guides/getting-started/prerequisites.md
+++ b/docs/guides/getting-started/prerequisites.md
@@ -49,7 +49,7 @@ winget install --id Rustlang.Rustup
 
 :::caution MSVC toolchain as default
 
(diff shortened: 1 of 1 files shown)


FabianLars:
Thank you! I think we should make that clear on the actual install step. This way users wouldn't end up with 2 toolchains installed either. What do you think?

JurrianFahner:
> Thank you! I think we should make that clear on the actual install step. This way users wouldn't end up with 2 toolchains installed either. What do you think?

I think we need to assume that the user would start with rust today. If somebody starts today the person will go to https://www.rust-lang.org/tools/install and needs to download `rustup-init.exe`.

If you leave everything as default it will lead to the following configuration (I used a clean windows 11 system to test it):
image

So I think the nominal case wouldn't lead to problems. However for a more experienced rust user who has played a while with rust already and who somehow uses a different toolchain, it is important that this user can find the solution to configure the right toolchain.

So I don't think we need to make it clear in the actual install step (because the nominal case wouldn't lead to a problem). Since it is an edge case I updated the docs, so a user can check whether the proper toolchain is selected.

FabianLars:
> So I don't think we need to make it clear in the actual install step (because the nominal case wouldn't lead to a problem)

It's actually not _that_ uncommon for some reason. Maybe rustup detects existing mingw installations or something. But anyway, it's too common to not have a note that asks users to make sure msvc is selected at install time already. This can be a separate PR but if we integrate it here we could refactor your new section to be explicit about the "If you already have Rust installed make sure msvc is the default" stuff

JurrianFahner:
@FabianLars I think your comment has been implemented, can you have a second look?

docs/guides/getting-started/prerequisites.md
```diff
@@ -47,6 +47,24 @@ Alternatively, you could use `winget` to install rustup using the following comm
...
+```
+
+The command should return either `i686-pc-windows-msvc`, `x86_64-pc-windows-msvc`, or `aarch64-pc-windows-msvc`.
+
+If the command returns a different toolchain, please configure the default toolchain in the following way: 
+
+```powershell
+rustup default stable-msvc
+```
+
```
FabianLars: To make it a bit more clear what direction i thought we should go here:
`````suggestion
:::caution MSVC toolchain as default

For full support for Tauri and tools like [`trunk`] make sure the MSVC Rust toolchain is the selected `default host triple` in the installer dialog.<br/>Depending on your system it should be either `x86_64-pc-windows-msvc`, `i686-pc-windows-msvc`, or `aarch64-pc-windows-msvc`.

If you already had Rust installed, you can make sure the correct toolchain is installed by running this command:

```powershell
rustup default stable-msvc
```

:::
`````

This way we could catch both cases:
- the installer selected gnu for whatever reason (user had mingw installed but not msvc or something)
- the user already had rust installed, but with the wrong toolchain

P.S. no need to accept my suggestion as-is, just wanted to visualize it.

File: docs/guides/getting-started/prerequisites.md [hunk 2]
```diff
@@ -47,6 +47,17 @@ Alternatively, you could use `winget` to install rustup using the following comm
 winget install --id Rustlang.Rustup
 ```
 
+:::caution MSVC toolchain as default
+
+For full support for Tauri and tools like [`trunk`] make sure the MSVC Rust toolchain is the selected `default host triple` in the installer dialog.<br/>Depending on your system it should be either `x86_64-pc-windows-msvc`, `i686-pc-windows-msvc`, or `aarch64-pc-windows-msvc`.
```
lorenzolewis: Is trunk supposed to be a link here? If not then remove the brackets around it. I'd also remove the `<br/>`, I think it's fine to be one paragraph.

docs/guides/getting-started/prerequisites.md
```diff
@@ -47,6 +47,17 @@ Alternatively, you could use `winget` to install rustup using the following comm
...
+
+If you already had Rust installed, you can make sure the correct toolchain is installed by running this command:
```
lorenzolewis:
```suggestion
If you already have Rust installed, you can make sure the correct toolchain is installed by running this command:
```

docs/guides/getting-started/prerequisites.md
lorenzolewis:
```suggestion
For full support for Tauri and tools like [`trunk`] make sure the MSVC Rust toolchain is the selected `default host triple` in the installer dialog. Depending on your system it should be either `x86_64-pc-windows-msvc`, `i686-pc-windows-msvc`, or `aarch64-pc-windows-msvc`.
```

FabianLars: Yeah it was supposed to be a link but i forgot to add that to my review comment...

docs/guides/getting-started/prerequisites.md
FabianLars: i wanted to make clear that this means, "if you had it installed prior to reading this guide". Does this still come across using `have`? (i should've worked harder in school 😞)

lorenzolewis: They convey the same thing, just different tenses. "have" is present tense (would which we prefer overall for writing style). Since you have "already" in front of it then that conveys it was installed prior to reading.

FabianLars: Okay, thank you for the explanation ❤️

FabianLars (approved):
thank you

PR #1064, opened 2023-01-13
docs 2.0: prototype for tabs via rehpye

Aierie:
Prototype demonstrating rehype plugin setup to allow specifying tabs in HTML comment syntax. There's some areas for improvement - a few are:

1. More robust HTML AST processing - it seems like the current plugin setup doesn't handle nodes where a string precedes a tab-comment directly (such as `something<!--- begin-tab-group --->`) 
2. Javascript needs to handle roving tabindex when focus is moved around the tab group using arrow keys
3. Doesn't handle all cases of invalid tab groups yet (such as unclosed tabs)

Opening this PR to demonstrate the high level approach - is this worth polishing up, and if so how? Are [directives](https://github.com/remarkjs/remark-directive) a good alternative to explore?

### DIFF
--- a/src/rehype-tabs/initializeTabs.ts
+++ b/src/rehype-tabs/initializeTabs.ts
@@ -1,0 +1,75 @@
+function initTabs(tabList: HTMLElement) {
+  let tabs: HTMLElement[] = Array.from(tabList.querySelectorAll('[role=tab]'))
+  let selectTab = (selected: HTMLElement) => {
+    for (let tab of tabs) {
+      let panel = document.getElementById(tab.getAttribute('aria-controls')!)!
+      if (tab === selected) {
+        tab.setAttribute('aria-selected', 'true')
+        tab.removeAttribute('tabindex')
+        panel.removeAttribute('hidden')
+      } else {
+        tab.setAttribute('aria-selected', 'false')
+        tab.tabIndex = -1
+        panel.hidden = true
+      }
+    }
+  }
+
+  let onKeydown = (event: KeyboardEvent) => {
+    let tab = event.currentTarget as HTMLElement
+    let index = Number(tab.dataset.index)
+    let next_index: number | undefined = undefined
+
+    switch (event.key) {
+      case 'ArrowLeft':
+        next_index = index - 1 < 0 ? tabs.length - 1 : index - 1
+        break
+      case 'ArrowRight':
+        next_index = index + 1 > tabs.length - 1 ? 0 : index + 1
+        break
+      case 'Home':
+        next_index = 0
+        break
+      case 'End':
+        next_index = tabs.length - 1
+        break
+      default:
+        break
+    }
+
+    if (next_index !== undefined) {
+      tabs[next_index].focus()
+      event.stopPropagation()
+      event.preventDefault()
+    }
+  }
+
+  let firstTab: HTMLElement | undefined = undefined
+  for (let index in tabs) {
+    let tab = tabs[index]
+    if (!firstTab) {
+      firstTab = tab
+    }
+
+    tab.tabIndex = -1
+    tab.setAttribute('aria-selected', 'false')
+    tab.setAttribute('data-index', index)
+
+    // TODO: implement roving tabindex that handles keyboard focus change properly
+    tab.addEventListener('keydown', onKeydown)
+    tab.addEventListener('click', (event: MouseEvent) => {
+      selectTab(event.currentTarget as HTMLElement)
+    })
+  }
+
+  selectTab(firstTab!)
+}
+
+for (let tabList of document.querySelectorAll(
+  '[role=tablist][data-docs-tablist]'
+)) {
+  initTabs(tabList as HTMLElement)
+}
+
+// Without this Typescript complains about not being able to compile this under --isolatedModules
+export {};
--- a/src/rehype-tabs/plugins.js
+++ b/src/rehype-tabs/plugins.js
@@ -53,6 +53,9 @@ let ctr = 0
 function rehypeTabs() {
   return (tree) => {
     visit(tree, (maybeStartNode, maybeStartIndex, parent) => {
+      if (isTabGroupEnding(maybeStartNode) || isTab(maybeStartNode)) {
+        throw new Error('Tab group missing opening')
+      }
       if (isTabGroupBeginning(maybeStartNode)) {
         let tabs = []
         let startIndex = maybeStartIndex
--- a/src/rehype-tabs/plugins.js
+++ b/src/rehype-tabs/plugins.js
@@ -1,0 +1,136 @@
+import { visit } from 'unist-util-visit'
+import { h } from 'hastscript'
+import rehypeRaw from 'rehype-raw'
+
+const TAB_GROUP_BEGINNING = /^- begin-tab-group -$/
+const TAB_GROUP_ENDING = /^- end-tab-group -$/
+const TAB_REGEX = /^- tab-title=(.+?) -$/
+function isTabGroupBeginning(node) {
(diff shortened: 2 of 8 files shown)


Aierie:
cc: @lorenzolewis

lorenzolewis:
I'm on mobile so apologies for any typos. 

I like the general direction. I know this is a draft but we'll want to make sure to pull things into separate files. 

From what you've done so far, so you think Styles Blocks (from the Directive plug-in you linked) would be better than HTML-style comments?

Lastly I'd say to try to keep the number of dependencies to the absolute minimum so that it's 1) easier to maintain and 2) potentially easier to port upstream directly to Astro if they're open to it.

Aierie:
Thanks for the quick response! I agree with the points you've raised re: dependencies and keeping code in separate files and will work on that. 

> From what you've done so far, so you think Styles Blocks (from the Directive plug-in you linked) would be better than HTML-style comments?

I gave the syntax issue some more thought and I think something like this could be good? Have a custom way of declaring a tab group, where top level content should only be list items. Take text in the first line of each list item as the title of the tab.

:::tab-group
- npm
  ```shell
  npm install --save-dev @tauri-apps/cli
  ```
  more content within the same list item
  
- yarn
  ```shell
  yarn add -D @tauri-apps/cli
  ```

- pnpm
  ```shell
  pnpm add -D @tauri-apps/cli
  ```

- Cargo
  ```shell
  cargo install tauri-cli
  ```
:::

Rationale:
1. Minimizes custom syntax and is easy to fall back to an unordered list - I wonder if it'd be easier on i18n processing too?
2. Clearer in markdown previews that a tab is there
3. Seems easier to detect invalid content (child node of the tab-group is not a list or there are 2 lists in the tab-group) and less risk of unclosed/mistaken tab groups

What do you think? (Just throwing this out there, no strong opinions on this - will continue working towards HTML comment syntax first)

Aierie:
Also for consideration - Astro seems to be [considering/preparing to support](https://discordapp.com/channels/830184174198718474/1058058026638123118) [Markdoc](https://markdoc.dev/docs/overview), though there is no precise timeline for it yet. There is a [community-created component](https://github.com/flexdinesh/blogster/tree/main/packages/astro-markdoc-renderer) at the moment.

Expressing tabs in Markdoc is shown in the [official Markdoc examples](https://markdoc.dev/docs/examples#tabs), the syntax in the markdown itself looks like this:

> {% tabs %}
> 
> {% tab label="React" %}
> React content goes here
> {% /tab %}
> 
> {% tab label="HTML" %}
> HTML content goes here
> {% /tab %}
> 
> {% /tabs %}

lorenzolewis:
Thinking through MarkDoc, I like the angle that it tries to solve, but I think you open up a lot more room for complexity and errors. One of the reasons we want to move away from MDX is because we kept having issues (especially with Crowdin for i18n) where something was modified into invalid MDX syntax and then broke the rendering pipeline. Successfully using MarkDoc or MDX I feel require solid support in all the different places you'll be modifying content (which unfortunately isn't the case).

To add a bit more perspective, you'll have the markdown-style documentation being modified in 3 places:
- mdBook (for when we move technical guides into upstream repos)
- Crowdin (i18n)
- In the `tauri-docs` repo directly

The first 2 in this list WILL have issues with MDX and MarkDoc. 

That being said, taking the Markdoc syntax as a reference (`{% tabs %}`) and building on top of that could be a better choice than the HTML-style comments (`<!-- tabs -->`). 

If we do Markdoc syntax instead of HTML comments then they will be shown in the markdown previews such as mdBook and GitHub directly. I think this a good thing because it removes "hidden logic" since I believe HTML comments aren't rendered in the preview for each.

The other option you gave is the `:::` syntax. I think this is a better option for "sections" of content where you don't need control over the individual items _inside_ of a section (think code snippets or the info banner annotations we have now with Docusaurus). When you introduce "sub-components" within those then they can start to get a bit messy I feel.

@FabianLars do you have any feedback on these options?

tl;dr: I think I'm liking the `{% tabs %}` syntax most out of these options.

src/rehype-tabs/initializeTabs.ts
```diff
@@ -0,0 +1,75 @@
+function initTabs(tabList: HTMLElement) {
```
lorenzolewis, line 1: How much of this file could be turned into an Astro component instead?

I'm using Docusaurus's implementation as reference, but they have two components: `TabGroup` and `TabItem` as React components. We could do a similar thing and have `src/components/Tabs/TabGroup.astro`, `src/components/Tabs/TabItem.astro`, and then `src/components/Tabs/styles.css` there for styles shared between the two. The last will also help so that you have isolated CSS and aren't doing global CSS which is for sure what we'll want.

src/styles/tabs.css
```diff
@@ -0,0 +1,22 @@
+.tab-container {
```
lorenzolewis, line 1: Similar comment as the `.ts` file where you have the tab logic, move these styles into a `src/components` folder for Tabs specifically

src/rehype-tabs/initializeTabs.ts
lorenzolewis: With this approach you might be able to remove the `hastscript` dep as well.

src/rehype-tabs/plugins.js
```diff
@@ -0,0 +1,139 @@
+import { visit } from 'unist-util-visit'
```
lorenzolewis, line 1: Can this be a TypeScript instead of JS file?

lorenzolewis: Also same comment as the `initializeTabs.ts` file, this could ideally be turned into an Astro component.

lorenzolewis (commented):
Overall pretty good first stab. I put more detailed feedback in line, but I'd take a look at the way Docusaurus isolates this type of feature logic into Components. They're using React components but I'm fairly confident all of this can be done with Astro components (and then you just use rehype to inject that astro component into the Markdown).

See https://github.com/facebook/docusaurus/tree/main/packages/docusaurus-theme-classic/src/theme/TabItem and https://github.com/facebook/docusaurus/tree/main/packages/docusaurus-theme-classic/src/theme/Tabs for reference.

Feel free to send me a message on Discord if you have any questions or clarifications. And thanks a million!

FabianLars:
> `@FabianLars` do you have any feedback on these options?

Hmm, i agree with everything you said actually.
It would have been nice if html comments are shown in previews because i think the existing tools have less of a problem with comments than the new markdoc syntax. Like i'd imagine that crowdin would just skip html comments? But yeah, if this breaks previews then i guess html comments are a no-go.

lorenzolewis:
> > `@FabianLars` do you have any feedback on these options?
> 
> Hmm, i agree with everything you said actually. It would have been nice if html comments are shown in previews because i think the existing tools have less of a problem with comments than the new markdoc syntax. Like i'd imagine that crowdin would just skip html comments? But yeah, if this breaks previews then i guess html comments are a no-go.

I wouldn't necessarily say it _breaks_ previews. I think it is just a bit less explicit with what's happening since previews hide comments by default. Crowdin doesn't render HTML-style comments so I think that's another reason to use something else.

FabianLars:
> Crowdin doesn't render HTML-style comments so I think that's another reason to use something else.

Isn't that in favor of using html comments? if crowdin doesn't render them it's less likely to break.

edit: Actuallyyy nevermind, we do have tabs with actual written content too, not just cmd line examples so these needs translations too

lorenzolewis:
Yup, that's what I'm thinking. Gets a bit tricky because technically you can have content in the special syntax that you want to be able to translate.

Aierie:
Closing for now until we're clearer on how we want to proceed with i18n + special markdown syntax.

PR #1072, opened 2023-01-22
Update vite.mdx

gblue1223:
When I run the command `npm create vite@latest`, it creates a new folder. So the `npm run ...` commands must executed in the folder.


FabianLars:
Thank you for contributing :)

The guide expects you do enter the vite project folder and create the tauri project inside of that.
So it's like `npm create vite && cd vite-project` ... all the other commands.
Otherwise the npm commands listed in the docs, like `npm install` and `npm run tauri init` won't work either (since they must be executed in the vite-project, where the package.json file is, too)

PR #1081, opened 2023-01-26
crucial fix on console you must input always cargo

miguelgargallo:
crucial fix on console you must input always cargo as well as we know is a common thing to add, but for newcomers, this is a must or console will return the error


FabianLars:
Thanks for contributing!

I think this would cause even more confusion since most tauri users do not use cargo but npm/yarn/pnpm. The new updater guide i'm working on will have a Tabber like the other guides where the user can select the package manager of their choice.

PR #1082, opened 2023-01-26
feat: new search index and auto scraping in CI

FabianLars:


src/theme/SearchBar/style.css
```diff
@@ -72,15 +96,22 @@
...
-  white-space: nowrap;
-  overflow: hidden;
-  text-overflow: ellipsis;
+}
+
+.meilisearch-autocomplete .docs-searchbar-suggestion--subcategory-column {
+  width: 30%;
+  text-align: right;
+  position: relative;
+  padding: 5.3333333333px 10.6666666667px;
```
lorenzolewis, line 105: That's some very precise padding

FabianLars: yeah, i just copy pasted it hoping it would work in our custom-ish css too 😂

FabianLars:
@lorenzolewis any last concerns?

lorenzolewis:
Did you want to update the netlify action along with this PR @FabianLars ?

FabianLars:
i thought i just did 😂

FabianLars:
okay now i was smart enough to actually click on save 💪 thanks for the reminder

lorenzolewis:
Is there an _easy_ way to also include some text from the page here as well? If the results aren't matching against anything in the body and only the header then even just a snippet of the first "content" in the page could be useful context in the results. 

Screenshot 2023-01-27 at 16 10 01

Here's an example of what I mean (and maybe truncating these to a max of 2/3 lines would be a good shout):

Screenshot 2023-01-27 at 16 12 03

FabianLars:
About the first thing: Yes i want to look into it, i _think_ the codeblock prevents it from finding the text = i think it can only find the text if it's immediately below the heading, but need to do a bit of testing to confirm that.

maybe we can use line-clamp for the second one.

lorenzolewis:
Ah, then this is fine since we'll have to rewrite it later for Astro anyway

lorenzolewis (approved):

FabianLars:
i was wrong, meilisearch just doesn't return it unless you typed text that is in that paragraph. i'll take a look at the settings, but this shouldn't need any changes to the tauri-docs repo so we can merge it already.

PR #1083, opened 2023-01-26
extract image ending classes with remark plugin

Aierie:
Some notes:

1. I'm not very sure how to evaluate XSS risk we introduce by moving the ending from `src` to `class`. Stuff in `src` seems to be uri encoded, but not stuff in `class` - my attempts at splitting `class` into 2 attributes using quotes and spaces weren't successful (strings in class are always just strings, OWASP calls attributes like this [Safe Sinks](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md#safe-sinks), so an attacker would first have to split the XSS into a different attribute like onload), but maybe someone more well-versed might be able to?
2. Currently excluding SVGs because of fragment identifiers
3. Not using Typescript because the Astro config file is still in js, though it should be easy to switch to Typescript (https://docs.astro.build/en/guides/configuring-astro/#the-astro-config-file).


remark-plugins/extract-image-class.js
```diff
@@ -0,0 +1,26 @@
+import { visit } from 'unist-util-visit'
+export function extractImageClass() {
+    return (tree) => {
+      visit(tree, "image", (node) => {
+        let [url, ending] = node.url.split("#");
```
Aierie: Making an assumption here that we have a spec-compliant URL (that should not have multiple # signs)

remark-plugins/extract-image-class.js
```diff
@@ -0,0 +1,26 @@
+import { visit } from 'unist-util-visit'
```
lorenzolewis: It seems like prefixing the plugin name itself with remark is common practice. Could you move this file to be in `src/plugins/remark-extract-image-class.js`?

remark-plugins/extract-image-class.js
lorenzolewis: Maybe change this so that it only splits on the last `#` if multiple are present? I can't think of a scenario where we'd actually have that, but would be nice to make it a little more flexible if we'd like to upstream this in the future.

remark-plugins/extract-image-class.js
```diff
@@ -0,0 +1,26 @@
...
+        let [url, ending] = node.url.split("#");
+
+        if (!ending) {
+          return;
+        }
+
+        // svg fragment identifiers should be left alone
+        // until we're clearer what we want to match
+        // https://css-tricks.com/svg-fragment-identifiers-work/
+        if (url.endsWith(".svg")) {
```
lorenzolewis: I read through the link you put here and I get why you're omitting svgs. But I wonder if the scenario in that article would come up in the context of markdown? I'd probably expect svgs to behave the same as other image types and could actually see that being a use case I'd use on the site.

remark-plugins/extract-image-class.js
```diff
@@ -0,0 +1,26 @@
...
+          return;
+        }
+
+        // see https://github.com/syntax-tree/mdast-util-to-hast/tree/main#fields-on-nodes
+        node.url = url;
+        node.data ??= {};
```
lorenzolewis: TIL about nullish coalescing assignment (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment), cool stuff!

remark-plugins/extract-image-class.js
```diff
@@ -0,0 +1,26 @@
+import { visit } from 'unist-util-visit'
+export function extractImageClass() {
```
lorenzolewis: Since this is the only export then does it make sense to do `export default` instead?

lorenzolewis (commented):
Overall really good. I'm assuming you've tested it out a bit. There's some places of this in use in the current repo (https://github.com/tauri-apps/tauri-docs/blob/feat/docs-2.0/old-docs/getting-started/setup/html-css-js.md#L24)

I'll put a few more notes in a separate reply to address the points in your original comment.

lorenzolewis:
> 1. I'm not very sure how to evaluate XSS risk we introduce by moving the ending from `src` to `class`. Stuff in `src` seems to be uri encoded, but not stuff in `class` - my attempts at splitting `class` into 2 attributes using quotes and spaces weren't successful (strings in class are always just strings, OWASP calls attributes like this [Safe Sinks](https://github.com/OWASP/CheatSheetSeries/blob/master/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.md#safe-sinks), so an attacker would first have to split the XSS into a different attribute like onload), but maybe someone more well-versed might be able to?

@FabianLars do you have any thoughts on this and if you think it's something needed? I have a feeling we don't need to do anything around this, we assume any input MD/MDX is trusted. 

> 2. Currently excluding SVGs because of fragment identifiers

I put a comment on this specific section of code.

> 3. Not using Typescript because the Astro config file is still in js, though it should be easy to switch to Typescript (https://docs.astro.build/en/guides/configuring-astro/#the-astro-config-file).

If you'd like to go ahead and switch the config to TS in this PR I'd be fine with that. Would be easier to do it now and get that nice type safety than later down the road.

FabianLars:
> `@FabianLars` do you have any thoughts on this and if you think it's something needed? I have a feeling we don't need to do anything around this, we assume any input MD/MDX is trusted as you said.

Should be fine as-is, especially since we (can) trust the MD/MDX input, _but_ if you want to be on the safe side i guess you could filter `ending` for `'`, `"` and `\` 🤷 

p.s. voting in favor of early TS introduction 👍

remark-plugins/extract-image-class.js
Aierie: I'm consciously trying to avoid default exports, since they generally have worse ergonomics than named exports:

https://basarat.gitbook.io/typescript/main-1/defaultisbad

That said, I'm happy to change it if they're the convention here!

remark-plugins/extract-image-class.js
Aierie: I can't tell which way authors would think about this, that said it seems easy enough to change in the future, so I can leave the comment about fragment identifiers in and remove the guard

remark-plugins/extract-image-class.js
Aierie: Sure, will do that

remark-plugins/extract-image-class.js
Aierie: I'll do that and leave a note about the URL spec and multiple #s... (EDIT) or actually the [html validator](https://validator.w3.org/nu/#textarea) along with a minimal example of invalid url could be better:

```
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <meta http-equiv="X-UA-Compatible" content="ie=edge">
    <title>HTML 5 Boilerplate</title>
    <link rel="stylesheet" href="style.css">
  </head>
  <body>
    <img src="/invalid#two#hashes" alt="bad image with invalid uri">
  </body>
</html>
```

remark-plugins/extract-image-class.js
lorenzolewis: I'm fine with either. Not sure what the "standard" is for remark plugins specifically, but I'd probably follow what other plugins do so we have a similar DX

Aierie:
And yep I've tested it out a bit!

remark-plugins/extract-image-class.js
FabianLars: Some very valid reasons in that link imo.
Other plugins seem to mostly use default exports tho :/ but many of them don't use esm either so idk how much of an inspiration they should be for us.

Aierie:
That should be all the fixes! @lorenzolewis

lorenzolewis (approved):
Looks good to me! Really good and detailed comments, so appreciate that a lot! Requested a review from @FabianLars as well for a third pair of eyes.

FabianLars (approved):

lorenzolewis:
Amazing, thanks so much @Aierie! <3

PR #1088, opened 2023-01-30
Update android path on macOS when setting up

wusyong:


docs/guides/getting-started/prerequisites/_fragments/_android-studio-setup.mdx
```diff
@@ -33,6 +34,9 @@ Now you need to set the `ANDROID_HOME` and `NDK_HOME` environment variables:
   {props.platform == 'linux'
     ? 'export ANDROID_HOME="$HOME/Android/Sdk"\nexport NDK_HOME="$ANDROID_HOME/ndk/25.0.8775105"'
     : ''}
+  {props.platform == 'macos'
```
wusyong, line 37: I'm not sure if this is the right way to set macos platform

FabianLars: should be correct

PR #1098, opened 2023-02-03
feat: switch to `meilisearch-docsearch`

amrbashir:


lorenzolewis:
Put a comment in Discord, but on the component footer how about saying "Powered by" instead of "Search by" since we want to make sure we don't try to represent Meilisearch ourselves

amrbashir:
Should be good to go now

FabianLars (approved):

amrbashir:
alright, that was hopefully the last bug and last version today

lorenzolewis:
Don't hate me, but one visual "glitch". Can the sizes of the magnifying glass and the spinner where the query being performed be set to the same width? They jump around when you're typing (at least on Safari) 

https://user-images.githubusercontent.com/15347255/217293513-0c1f16c3-476b-4f6e-ba88-523cbdb8a9dd.mov

amrbashir:
done!

lorenzolewis:
Perfect! Ready to merge?

amrbashir:
if the glitch is gone, then yes!

PR #1103, opened 2023-02-05
Update sidecar.md

JohnScience:
I've added a less environment-dependent way for obtaining the host triple, which uses the simple library I've made.

In case of security concerns, I'm willing to delegate ownership over [`rustc-host`](https://crates.io/crates/rustc-host) crate and its GitHub repo.

### DIFF
--- a/docs/guides/building/sidecar.md
+++ b/docs/guides/building/sidecar.md
@@ -39,11 +39,13 @@ Here is a sample to illustrate the configuration. This is not a complete `tauri.
 }
 ```
 
-A binary with the same name and a `-$TARGET_TRIPLE` suffix must exist on the specified path. For instance, `"externalBin": ["binaries/my-sidecar"]` requires a `src-tauri/binaries/my-sidecar-x86_64-unknown-linux-gnu` executable on Linux. You can find the current platform's target triple by running the following command:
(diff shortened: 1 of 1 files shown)


JohnScience:
It would probably be a good idea to make this part:

---
With grep and cut available (i.e. on most Linux distributions), you can find the current platform's target triple by running the following command:

```shell
rustc -Vv | grep host | cut -f2 -d' '
```

In Rust, you can also obtain the host triple using [rustc_host::from_cli()](https://docs.rs/rustc-host/latest/rustc_host/fn.from_cli.html) from [rustc-host](https://crates.io/crates/rustc-host) crate.

---

be styled like this (<https://tauri.app/v1/guides/getting-started/setup/html-css-js>) element/view/component:

screenshot

However, I'm not sure how to do it.

I checked out <https://github.com/tauri-apps/tauri-docs/tree/dev/docs/guides/getting-started> and I was a bit confused by `setup` directory. The PR is good to do. I can make an iterative improvement later.

FabianLars:
Thank you for contributing!

To be completely honest with you, we weren't quite comfortable with recommending an uncontrolled external binary just for getting the target triple.
Instead we opted for a PowerShell example, so that we now have examples for Linux&macOS, Windows, and code (nodejs in this case)

PR #1104, opened 2023-02-05
Update prerequisites.md

fei4xu:
Be more accurate and consistent with note at https://next--tauri.netlify.app/next/guides/building/windows/ (or https://tauri.app/v1/guides/building/windows)

### DIFF
--- a/docs/guides/getting-started/prerequisites.md
+++ b/docs/guides/getting-started/prerequisites.md
@@ -29,7 +29,7 @@ You will need to install Microsoft Visual Studio C++ build tools. The easiest wa
 
 :::note
 
-WebView2 is pre-installed in Windows 11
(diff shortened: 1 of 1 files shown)


FabianLars:
Thank you for contributing!

The difference between windows 11 and windows 10, is that there is no actual guarantee that it exists on windows 10 since it's bound to an os update _and_ can be easily manually uninstalled in contrast to win11 where you can't remove it via the default dialogs.
So i'm a bit unsure about the wording here 🤔

fei4xu:
@FabianLars I understand your concern.

The reason why I still created the PR is that the same wording is already used at another page of the doc (https://tauri.app/v1/guides/building/windows), I actually copied the wording from there
image

To make it even more accurate, I insist we mention win10 will have webview2 installed by the system and add another wording of your explanation.

simonhyll (approved):

PR #1105, opened 2023-02-06
DropdownMenu component

Aierie:
DropdownMenu is toggled via click/enter/space when there is JS, and via hover and foucus when there isn't. This follows the guide in https://css-tricks.com/in-praise-of-the-unambiguous-click-menu/.


src/components/Layout/DropdownMenu.astro
```diff
@@ -0,0 +1,122 @@
...
+
+let id = uid();
+---
+
+<div 
+  data-navigation-dropdown-menu 
+  data-unactivated
+  data-inline-menu={Astro.props.inlineMenu}
+> 
+  <button
```
lorenzolewis: Should use `label` with `tabindex="0"` instead of `button` due to this Safari bug https://bugs.webkit.org/show_bug.cgi?id=22261

lorenzolewis:
Per the convo in Discord, I just branched to serve as a bit of a working tree for now: https://github.com/tauri-apps/tauri-docs/tree/feat/dropdown-and-menu

I put an empty Dropdown component here: https://github.com/tauri-apps/tauri-docs/blob/feat/dropdown-and-menu/src/components/Dropdown.astro

I think this would contain the same JS logic you have in place, just minus any styling other than applying classes specified in `global.css` 

I put these together to show the functional CSS for the dropdown: https://github.com/tauri-apps/tauri-docs/blob/feat/dropdown-and-menu/src/styles/global.css#L155. That honestly might be pretty close to all the CSS that's needed, and then just the extra JS functionality you've made.

Now, to separate `.menu` into its own thing. I keep going back to DaisyUI and how they have everything interact, I think it's pretty elegant. So I'd like to have something similar to this: https://daisyui.com/components/menu/. You can see it used with their dropdown component (which the CSS there will look very familiar ;) )

Here's some unstyled and only layout-focused CSS for menu https://github.com/saadeghi/daisyui/blob/master/src/components/unstyled/menu.css

I put together a rough mock of how `<Dropdown />` and `.menu` could work together here: https://github.com/tauri-apps/tauri-docs/blob/feat/dropdown-and-menu/src/pages/components.astro#L21

Hope that all makes sense and give me a shout if not! I'm more than happy to setup the styled CSS for menu (padding, background, etc.) if you'd like. More than happy for other directions by any means! <3

lorenzolewis:
Copy/pasting my comments from Discord to keep it inline here as well

> Could we use that PR to focus on making the <Dropdown /> component functionality in the direction you were headed in? But instead of trying to combine all the menu visual stylings with that how about we leave the styling up to the <NavBar />, <NavBarLocaleTheme /> and <NavBarLinks /> components to handle instead (both the button and the dropdown itself) and then they can use the new <Dropdown /> component for pulling in the default functional CSS (:focus and :focus-within so that everything still functions without JS) and then the progressive-JS that you've already coded?
For :hover how about making that as a prop that can be used to enable/disable hover?

src/components/Layout/Dropdown.astro
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  .dropdown:focus-within > .dropdown-toggle:not([aria-expanded]) + .dropdown-content,
+  .dropdown:hover > .dropdown-toggle:not([aria-expanded]) + .dropdown-content {
+    @apply block;
+  }
+
+  /* With JS */
+  .dropdown-toggle[aria-expanded=true] {
+    @apply rounded-b-none;
```
Aierie: Button style for expanded dropdown

src/components/Layout/Dropdown.astro
```diff
@@ -0,0 +1,137 @@
...
+    @apply hidden absolute;
+  }
+  .dropdown-inline .dropdown-content {
+    @apply static;
+  }
+
+  /* No JS */
+  .dropdown:hover > .dropdown-toggle:not([aria-expanded]),
+  .dropdown:focus-within > .dropdown-toggle:not([aria-expanded]) {
+    @apply rounded-b-none;
```
Aierie: Button style for expanded dropdown when trigger is hover or there is no JS

src/styles/global.css
```diff
@@ -171,53 +171,29 @@
...
-  .menu.menu-hover > li:hover:has(> ul),
-  .menu > li:focus:has(> ul) > *:not(ul),
-  .menu > li:focus-within:has(> ul) > *:not(ul) {
-    @apply rounded-b-none;
+  /************************/
+  /* Navbar dropdown menu */
+  /************************/
+  .navbar-dropdown-menu:hover, 
+  .navbar-dropdown-menu:focus-within, 
+  .navbar-dropdown-menu [aria-expanded=true] {
```
Aierie: Button style for expanded click dropdowns

Aierie (commented):
@lorenzolewis I think this is getting close to what we discussed, though I'm unsure how to really split the button styles nicely; the button really should live in the Dropdown so that we can be sure that the JS can target things correctly, but its styles are related to its container's... any ideas?

src/components/Layout/Dropdown.astro
```diff
@@ -0,0 +1,137 @@
+---
+interface Props {
+  inlineMenu?: boolean;
+  label?: string;
```
lorenzolewis: I'd call this `ariaLabel` to make it more clear what it's used for

src/components/Layout/Dropdown.astro
```diff
@@ -0,0 +1,137 @@
...
+
+let id = uid();
+let { inlineMenu, label, class: className, trigger } = Astro.props; 
+---
+
+<div
+  data-trigger={trigger ?? "click"}
+  class:list={["dropdown", "js-dropdown", { "dropdown-inline js-dropdown-inline": inlineMenu }, className]}
+> 
+  <button
```
lorenzolewis: Should this be a label because of that Safari bug? https://bugs.webkit.org/show_bug.cgi?id=22261

lorenzolewis:
> @lorenzolewis I think this is getting close to what we discussed, though I'm unsure how to really split the button styles nicely; the button really should live in the Dropdown so that we can be sure that the JS can target things correctly, but its styles are related to its container's... any ideas?

You're talking about the toggle/button here, right? You have these states that you need to be able to control/toggle from within the components:

- Normal
- Hover
- Active (content expanded)

But that makes it a bit difficult if we want to keep the component truly "visual" styling-free. You're already using `class` as a prop. Maybe expand on that idea and you could pass specific styles as props:

```ts
interface Props {
   inlineMenu?: boolean;
   label?: string;
   // Is class needed here? See note below
   class?: string;
   classToggleHover: string;
   classToggleActive: string;
   trigger?: "click" | "hover";
 }
```

Right now you have `class` being applied to the container, but is there a reason you'd need to externally style the container instead of just the button/toggle and content?

src/components/Layout/DropdownMenu.astro
Aierie: `label` doesn't seem great because it provides a worse experience on other browsers if we don't add code to mark it as a button to assistive tech and handle key events in addition to clicks. I can't really find a recommendation anywhere to use `label`s to replace buttons in Safari, I'm not sure why DaisyUI chose it.

Aierie:
> You're talking about the toggle/button here, right? You have these states that you need to be able to control/toggle from within the components:
> 
> Normal
Hover
Active (content expanded)

I think passing more classes as props is okay, but it won't work for the hover and focus styles that we use when there is no JavaScript. That sounds like a decent compromise though - we don't need a perfect experience without JS, people can still reach the menu items.

src/components/Layout/DropdownMenu.astro
lorenzolewis: I saw a separate website a few days back that also referenced that same bug but can't find it for now. Let's go with `button` for now but put a note with a link to that WebKit bug in code for future reference.

lorenzolewis:
> I think passing more classes as props is okay, but it won't work for the hover and focus styles that we use when there is no JavaScript.

You're talking about making the passed classes apply for `:hover` and `:focus` that are controlled within the component I think? 

The logic that would have to happen would be like this:

1. Pass the state classes for the toggle to the component via props (without any pseudo styles like `:hover` and `:focus` as that is controlled internally by the component)
2. The component would then need to somehow apply those classes into the CSS-only fallback state styles for the toggle (this is where `:hover` and `:focus` would get applied)
3. If JS is present remove the CSS-only styles for the toggle (`:hover` and `:focus`) in favour of the JS logic you've made

I think you're talking about handling point #2, right? I didn't see how you'd even handle `:hover` and the link with Astro's `class:list` directive unless I overlooked something (might be something I can ask Astro core if you're thinking down the same path).

Also thinking through the case if you have a more complicated toggle (a div with flex with multiple child elements for example which what a button with an icon could look like) then you also _may_ need to style those deeply nested child elements as well (text colours on hover come to mind). 

So this might be you need to apply a whole set of styles either scoped within the component or globally but you still want to retain control of toggling those sets of styles within the component. 

---

Crazy, over-complicated ideas may follow below, take with a large grain of salt 😅

If we think about passing a block of CSS styles instead of just a list of classes then I wonder if the below would work:

```astro
---
interface Props {
   toggleHoverRawCss: string;
   toggleActiveRawCss: string;
 }

const { toggleHoverRawCss, toggleHoverActiveRawCss } = Astro.props

const rawToggleCss = functionThatTransformsCssString();
// ^^ JS code that will parse through passed raw styles and transform it into
// the appropriate CSS-only classes such as :hover and :focus/:focus-within 
---

<style is:inline set:html={rawToggleCss}></style>

<!-- Rest of Dropdown component -->
```

Could be crazy and over-complicated, let me know what you think.

Reference: https://docs.astro.build/en/guides/styling/#raw-css-imports

Edit: This would probably also not play well with Tailwind and you'd probably have to safelist classes https://tailwindcss.com/docs/content-configuration#safelisting-classes

lorenzolewis:
Btw, if moving to `button` instead, could you pull out the tabIndex from here since that was used as a workaround with the label:

https://github.com/tauri-apps/tauri-docs/blob/ffdb0f09fc2403de835832572913436364c06e90/src/layouts/BaseLayout.astro#L54

Aierie:
I'll make the following changes:

1. Change prop from `label` -> `ariaLabel`
2. Remove `tabindex=-1` from `body`
3. Add a comment about Safari focus issues
4. Use `@apply` to use `.btn` styles on dropdown toggle directly, instead of relying on `.menu` descendant styles

lorenzolewis:
Sounds good, thanks for the discussion on Discord, @Aierie !

lorenzolewis:
@Aierie following up to see if there's anything I can do to help with this one

Aierie:
Hey @lorenzolewis ! Sorry, have been a bit busy. I've updated the PR with the stuff we discussed. Is the CI `update-docs` step expected to fail?

FabianLars:
> Is the CI update-docs step expected to fail?

Yep, at least in the docs-2.0 branch. https://github.com/tauri-apps/tauri-docs/pull/1122

Since the workflow is no-op on this branch anyway we can ignore it for now.

src/components/Layout/Dropdown.astro
```diff
@@ -0,0 +1,163 @@
...
+  <div 
+    id={id} 
+    class="dropdown-content z-nav-dropdown" 
+  >
+    <slot name="content"/>
+  </div>
+</div>
+
+<style>
+  @import '../../styles/global.css';
```
lorenzolewis: @Aierie I don't think this import is needed since the base layout pulls it in, right? I tested it locally and seemed to work alright.

lorenzolewis (commented):
Looks amazing! Thank you so much, even fixed a couple of other things I was going to take a look at after. 

One comment I left on a CSS import, not sure if it's needed. But overall seems to work really well on mobile Safari (tested touch), macOS Safari (with and w/o JS enabled).

lorenzolewis:
Btw, updated the base 2.0 branch so that it won't run the `update-docs` workflow on PRs now

src/components/Layout/Dropdown.astro
```diff
@@ -0,0 +1,163 @@
+---
```
lorenzolewis: Oh, one thing I just noticed, let's move this file up a folder into the `src/components` directory directly since it's not exclusive to layouts?

src/components/Layout/Dropdown.astro
Aierie: When I run `pnpm dev` without the `@import` statement, the build completes and I can view the site. However, I get this error in the build logs:

```
 error   The `btn` class does not exist. If `btn` is a custom class, make sure it is defined within a `@layer` directive.
  File:
    /Users/michael/Personal/Code/tauri-docs/src/components/Layout/Dropdown.astro:66:5
  Code:
    64 |  
    65 |    .dropdown-toggle {
    66 |      @apply btn btn-inline text-left;
       |       ^
    67 |    }
    68 |
  Stacktrace:
CssSyntaxError: [postcss] /Users/michael/Personal/Code/tauri-docs/src/components/Layout/Dropdown.astro?astro&type=style&lang.css:11:5: The `btn` class does not exist. If `btn` is a custom class, make sure it is defined within a `@layer` directive.
```

src/components/Layout/Dropdown.astro
Aierie: Done!

src/components/Layout/Dropdown.astro
lorenzolewis: I think it's because of this: https://docs.astro.build/en/guides/integrations-guide/tailwind/#class-does-not-exist-with-apply-directives

But you also get a warning with the import on build due to the font being reimported I think:

```
./files/rubik-cyrillic-ext-400-normal.woff2 referenced in /Users/lorenzolewis/Developer/tauri-docs/src/components/Layout/Dropdown.astro?astro&type=style&index=0&lang.css didn't resolve at build time, it will remain unchanged to be resolved at runtime
```

If you remove the `.dropdown-toggle` styles and move `btn btn-inline text-left` to be inline directly in the `button` component then that should fix it.

src/components/Dropdown.astro
```diff
@@ -0,0 +1,163 @@
...
+    https://bugs.webkit.org/show_bug.cgi?id=22261
+
+    As a result, this dropdown had issues closing because 
+    focusout events were not being fired
+    and that's why we have an "outside click" listener
+  -->
+  <button 
+    aria-controls={id} 
+    aria-label={ariaLabel} 
+    class="dropdown-toggle js-dropdown-toggle"
```
lorenzolewis:
```suggestion
    class="btn btn-inline text-left dropdown-toggle js-dropdown-toggle" 
```

src/components/Dropdown.astro
```diff
@@ -0,0 +1,163 @@
...
+    @apply hidden absolute bg-base-200 rounded-b rounded-tr z-dropdown;
+  }
+  .dropdown-inline .dropdown-content {
+    @apply static;
+  }
+
+  .dropdown-toggle {
+    @apply btn btn-inline text-left;
+  }
+
```
lorenzolewis:
```suggestion

```

src/components/Dropdown.astro
```diff
@@ -0,0 +1,163 @@
...
+  <div 
+    id={id} 
+    class="dropdown-content z-nav-dropdown" 
+  >
+    <slot name="content"/>
+  </div>
+</div>
+
+<style>
+  @import '../styles/global.css';
```
lorenzolewis:
```suggestion
```

lorenzolewis:
The changes I just suggested _should_ resolve the need for importing the CSS file.

Aierie:
Applied the fixes, works without the CSS import now!

lorenzolewis:
Amazing! Thank you so much! I'll take a look at the other PR a bit later this week in more detail.

PR #1106, opened 2023-02-07
Add Guix setup instructions

declantsien:
This PR add Tauri development environment setup instructions using [Gnu Guix](https://en.wikipedia.org/wiki/GNU_Guix).

GNU Guix is a functional cross-platform package manager and a tool to instantiate and manage Unix-like operating systems, based on the Nix package manager. Configuration and package recipes are written in Guile Scheme.

Later we can proceed to distribute Tauri app on Guix..

PR #1107, opened 2023-02-07
 Add Guix setup instructions for next branch

declantsien:
There is #1106 which is using webkit2gtk-4.0.

I saw #1097 mentioned webkit2gtk-4.1.
Instead of waiting for #1106 get merged and ported back to next branch and the send a PR to reflect webkit2gtk-4.1.
I decided to send this separated PR for next branch.


lorenzolewis:
We're going to make a cut of the docs in `dev` and re-do what's in the `next` branch soon. With that in mind, I'm going to close this one out since #1106 will automatically get included in that (noting the version difference).

PR #1108, opened 2023-02-07
Add i18n for search

FabianLars:
most beautiful json file ever

for the astro docs i'll likely create a json generator (with a default `en` json file for easier local dev ig)


FabianLars:
let's see how it goes 🤷

PR #1117, opened 2023-02-09
Update staging index on push. swap index when done

FabianLars:


FabianLars:
You may not have seen it, but it worked - source: trust me bro

PR #1120, opened 2023-02-12
Added missing step to custom titlebar example

Nik4ant:
Custom titlebar example didn't mention that you need to add window permissions in `tauri.conf.json` to make it work


FabianLars (approved):
thank you

PR #1121, opened 2023-02-12
:memo: Adds dependency update commands to configuration

ekwoka:
Currently, information covering how to make a new Tauri Mobile app completely skips over the process of updating the dependencies to the alpha track. This won't matter much once 2.0.0 actually releases, but right now it can be a blocking step to getting started trying it out.

> side note: I could not get the next branch to run locally for a myriad reasons that weren't issues on the dev/release branches

### DIFF
--- a/docs/mobile/development/configuration.mdx
+++ b/docs/mobile/development/configuration.mdx
@@ -7,9 +7,9 @@ import TabItem from '@theme/TabItem'
 
 # Configuration
 
-Currently, Tauri Mobile support is only available on the `2.0.0` alpha tracks, so the main tauri build tools will not produce a project that supports it. You'll need to update a few dependencies to get setup
+Currently, Tauri Mobile support is only available on the `2.0.0` alpha tracks, so the main tauri build tools will not produce a project that supports it. You'll need to update a few dependencies to get started.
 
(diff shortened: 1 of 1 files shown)


docs/mobile/development/configuration.mdx
```diff
@@ -7,6 +7,51 @@ import TabItem from '@theme/TabItem'
...
+
+  </TabItem>
+</Tabs>
+
+To update the Cargo dependencies, run the following in the `src-tauri` folder:
+
+```shell
+cargo add tauri@2.0.0-alpha.0
+cargo add tauri-build@2.0.0-alpha.0 --build
+cargo install tauri-cli --version "^2.0.0-alpha"
```
FabianLars: imo this should either be part of the tabber, or separate with a note that it's not required for everyone, only those that use `cargo tauri x` - what do you think?

ekwoka: Is this step not still required for someone that has made a project with create-tauri-app today? Don't they have to do both the javascript and cargo updates?

FabianLars: oh damn, the preview window is not as clear as i expected it to be. my comment was only about the tauri-cli installation, not tauri and tauri-build, sorry.

ekwoka: ah, yes, understood

FabianLars (approved):

PR #1143, opened 2023-03-06
fix scope example

sameoldlab:
Changed `scope` example to provide an array, as described in the docs and how it works.


FabianLars:
Thanks for contributing! The API docs are auto-generated from the upstream tauri repo. So if you don't mind please update it here https://github.com/tauri-apps/tauri/blob/dev/tooling/api/src/shell.ts too.

FabianLars (approved):

sameoldlab:
done https://github.com/tauri-apps/tauri/pull/6401

PR #1146, opened 2023-03-07
Rewrite updater guide

FabianLars:
Still missing the custom dialog part, but i thought i push what i have so far,

(p.s. i scratched my prior 100 from-scratch rewrites and opted for a simpler approach, re-using more of the old stuff to hopefully keep my sanity here)


docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,126 @@
...
-"dialog" if present must be a boolean. By default, it's set to true. If enabled, [events](#events) are turned off as the updater handles everything. If you need the custom events, you MUST turn off the built-in dialog.
-
-"pubkey" must be a valid public-key generated with Tauri CLI. See [Signing updates](#signing-updates).
+# Updater
 
-### Update Requests
+Tauri offers a built-in updater for the MSI (Windows), AppImage (Linux) and App bundle (macOS) distribution formats.
 
-Tauri is indifferent to the request the client application provides for update checking.
+## Configuration
```
amrbashir: I don't think this heading is needed at all especially since there is another one further down called "Tauri Configuration". I think the heading could be removed but keep the content under it.

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,126 @@
...
 
-It may also include other identifying criteria, such as operating system version, to allow the server to deliver as fine-grained an update as you would like.
+To sign your updates you need two things:
 
-How you include the version identifier, or other criteria is specific to the server from which you request updates. A common approach is to use query parameters, [Configuration](#configuration) shows an example.
+- The _public key_, which will be added to your `tauri.conf.json` file later, to validate update artifacts before the installation.
+- The _private key_, which is used to sign your update artifacts and should NEVER be shared with anyone. Also, if you loose this key, you will NOT be able to publish new updates to your current user base. It is crucial to store it in a safe place where you can always access it.
 
-### Built-in dialog
+To generate the keys, you need to use the Tauri CLI (use `$HOME/.tauri/myapp.key` if you're on Windows):
```
amrbashir: I think the `use <blah> on Windows` should be moved under the command not before it.

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,126 @@
...
-By default, the updater uses a built-in dialog API from Tauri.
+<Command name="signer generate -w ~/.tauri/myapp.key" />
 
-![New Update](https://i.imgur.com/UMilB5A.png)
+To make Tauri pick up these keys should be set as environment variables when building your app:
 
-The dialog release notes are represented by the update `note` provided by the [server](#server-support).
-If the user accepts, the update is downloaded and installed. Afterward, the user is prompted to restart the application.
+- `TAURI_PRIVATE_KEY`: Path or content of your private key
+- `TAURI_KEY_PASSWORD`: Your private key password (optional)
```
amrbashir: should be moved to the updater artifacts section

docs/guides/distribution/updater.md
```diff
@@ -224,113 +130,103 @@ The alternate update technique uses a plain JSON file, storing your update metad
...
-    └── app.x64.msi.zip (update bundle)
-    └── app.x64.msi.zip.sig
-```
+### Custom stuff
 
-### Linux
+:::caution
 
-On Linux, we create a .tar.gz from the AppImage.
+You need to disable the built-in dialog in your [Tauri configuration](#configuration) to enable the JavaScript APIs and updater events!
```
amrbashir:
```suggestion
You need to disable the built-in dialog in your [Tauri configuration](#tauri-configuration) to enable the JavaScript APIs and updater events!
```

docs/guides/distribution/updater.md
```diff
@@ -224,113 +130,103 @@ The alternate update technique uses a plain JSON file, storing your update metad
...
+If the user accepts, the update is downloaded and installed. Afterwards, the user is prompted to restart the application.
 
-```
-target/release/bundle
-└── msi
-    └── app.x64.msi
-    └── app.x64.msi.zip (update bundle)
-    └── app.x64.msi.zip.sig
-```
+### Custom stuff
```
amrbashir:
```suggestion
### Custom Dialog and Logic
```

amrbashir (changes requested):

docs/guides/distribution/updater.md
FabianLars: whoops

docs/guides/distribution/updater.md
FabianLars: Yeah you're right. This heading is indeed a bit much.

amrbashir (approved):

FabianLars:
@lorenzolewis Do you wanna make a quick wording check or do you trust me? 😆 (didn't get any negative feedback from the community so i guess content-wise it's fine to merge and we can iterate from there if needed)

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
+# Updater
 
-Tauri is indifferent to the request the client application provides for update checking.
+Tauri offers a built-in updater for the MSI (Windows), AppImage (Linux) and App bundle (macOS) distribution formats.
 
-`Accept: application/json` is added to the request headers because Tauri is responsible for parsing the response.
+Once your Tauri project is ready, you can configure Tauri's updater to enable auto updating for your users.
 
-For the requirements imposed on the responses and the body format of an update, response see [Server Support](#server-support).
+## Signing updates
```
lorenzolewis: Do we have any consistency between headings being "Capital Case" vs. "Sentence case"? Maybe that's something we could add to the styles section here: https://github.com/tauri-apps/tauri-docs/blob/dev/.github/CONTRIBUTING.md

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
 
-Your update request must _at least_ include a version identifier so that the server can determine whether an update for this specific version is required.
+Tauri's updater has a built-in signature to ensure updates are safe to be installed. This safety measurement can not be turned off.
```
lorenzolewis:
```suggestion
Tauri's updater has a built-in signature to ensure that updates are safe to be installed. This safety measurement cannot be turned off.
```
What's the _why_ behind why you can't turn it off? For security? A technical limitation? I'd expand on this (but very slightly, just added to the sentence).

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
 
-It may also include other identifying criteria, such as operating system version, to allow the server to deliver as fine-grained an update as you would like.
+To sign your updates you need two things:
 
-How you include the version identifier, or other criteria is specific to the server from which you request updates. A common approach is to use query parameters, [Configuration](#configuration) shows an example.
+- The _public key_, which will be added to your `tauri.conf.json` file later, to validate update artifacts before the installation.
+- The _private key_, which is used to sign your update artifacts and should NEVER be shared with anyone. Also, if you loose this key, you will NOT be able to publish new updates to your current user base. It is crucial to store it in a safe place where you can always access it.
```
lorenzolewis:
```suggestion
1. The _public key_, which will be added to your `tauri.conf.json` file later, to validate update artifacts before the installation.
2. The _private key_, which is used to sign your update artifacts and should NEVER be shared with anyone. Also, if you loose this key, you will NOT be able to publish new updates to your current user base. It is crucial to store it in a safe place where you can always access it.
```

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
 
-### Built-in dialog
+To generate the keys, you need to use the Tauri CLI:
```
lorenzolewis:
```suggestion
To generate the keys on Linux and macOS you can use the Tauri CLI:
```

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
 
-By default, the updater uses a built-in dialog API from Tauri.
+<Command name="signer generate -w ~/.tauri/myapp.key" />
 
-![New Update](https://i.imgur.com/UMilB5A.png)
+If you are on Windows, you should use `$HOME/.tauri/myapp.key` instead.
```
lorenzolewis: I'd explicitly write the out Windows variant in another Command

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
-The dialog release notes are represented by the update `note` provided by the [server](#server-support).
-If the user accepts, the update is downloaded and installed. Afterward, the user is prompted to restart the application.
+## Tauri configuration
 
-### Javascript API
-
-:::caution
-You need to _disable built-in dialog_ in your [tauri configuration](#configuration); Otherwise, the javascript API will NOT work.
-:::
+Now you need to configure Tauri's updater. To do this, add this to your Tauri config file:
```
lorenzolewis:
```suggestion
Now you need to configure Tauri's updater. To do this, add this to your Tauri config:
```
I think I'd add some links to the config API docs here (for now just the main page maybe?)

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
-Emitted data:
-
-```
-version    Version announced by the server
-date       Date announced by the server
-body       Note announced by the server
-```
-
-#### Rust
+The required keys are `"active"`, `"endpoints"` and `"pubkey"` to enable the updater. `"dialog"` is optional and will default to `true` if not set.
```
lorenzolewis, line 52: There's a LOT of detail for what each of these keys are, the values, etc. I wonder if this information is better to have in the reference docs and link to it there? Helps with future maintainability as well

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
-Event: `tauri://update-install`
+## Update artifacts
 
-#### Rust
+Tauri's bundler will automatically generate and sign update artifacts once the updater is correctly configured and enabled.
 
-```rust
-window.emit("tauri://update-install".to_string(), None);
-```
+Before building your app, you need to set environment variables for the private key and password:
```
lorenzolewis, line 79: I'd add a code snippet here of how you set the environment variables

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
-PENDING is emitted when the download is started and DONE when the installation is complete. You can then ask to restart the application.
+- **macOS**: On macOS, Tauri will create a `.tar.gz` archive from the application bundle inside the `target/release/bundle/macos/` folder:
 
-ERROR is emitted when there is an error with the updater. We suggest listening to this event even if the dialog is enabled.
+  - `myapp.app` - the standard app bundle.
+  - `myapp.app.tar.gz` - the updater bundle.
+  - `myapp.app.tar.gz.sig` - the signature of the update bundle.
 
-#### Rust
+- **Linux**: On Linux, Tauri will create a `.tar.gz` archive from the AppImage inside the `target/release/bundle/appimage/` folder:
```
lorenzolewis, line 107: Generally I've been trying to sort Linux, macOS and Windows alphabetically to get a bit more consistency

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
-listen('tauri://update-status', function (res) {
-  console.log('New status: ', res)
-})
-```
+The signature can be uploaded and shared safely as long as your private key is secure.
 
 ## Server Support
 
-Your server should determine whether an update is required based on the [Update Request](#update-requests) your client issues.
+Tauri's updater supports 2 kinds of servers. You can host a static JSON file on services like S3 or GitHub gists, or host a dynamic update server. The static JSON file is easier to use while an update server will give you finer control over the update mechanism.
```
lorenzolewis: I'd list out the two kinds it supports and the pros/cons a bit more clearly.
```suggestion
Tauri's updater supports two ways of announcing update data:

 - Static JSON file (to use on services like S3 or GitHub gists)
 - Dynamic update server

The static JSON file is easier to use while a dynamic update server will give you finer control over the update mechanism.
```

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
+
+### Dynamic Update Server
 
-If an update is required, your server should respond with a status code of [200 OK] and include the [update JSON](#update-server-json-format) in the body.
+With this approach, Tauri will follow the update server's instructions. To disable the internal version check you can [overwrite Tauri's version comparison] to always install the version sent by the server. This could be useful if you need to roll back your app version quickly.
 
-If no update is required your server must respond with a status code of [204 No Content].
+Your server can use variables defined in the `endpoint` url above to determine if an update is required. If you need more data, you can include additional [request headers in Rust] to your liking.
```
lorenzolewis: Is there a place in the API docs to link to the endpoint object?

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
 
-### Update Server JSON Format
+Your server should respond with a status code of [`204 No Content`] if there is no update available.
 
-When an update is available, Tauri expects the following schema in response to the update request provided:
+If an update is required, your server should respond with a status code of [`200 OK`] and a JSON response in this format:
```
lorenzolewis: Is this expected version documented anywhere else? Or is this the only place?

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
-
-"signature" is the content of the `.sig` file which was generated by Tauri's CLI. See [Signing Updates](#signing-updates) for instructions on how to set up the required keys.
+- `"version"` must be a valid semver, with or without a leading `v`, meaning that both `1.0.0` and `v1.0.0` are valid.
+- `"url"` must be a valid url to the update bundle.
+- `"signature"` must be the **content** of the generated `.sig` file. The signature may change each time you run `tauri build` so make sure to always update it.
+- `"notes"`: Here you can add notes about the update, like release notes. Tauri's default dialog will present this to the user when it asks if it's allowed to update.
+- `"pub_date"` must be formatted according to [RFC 3339] if present.
 
-### Update File JSON Format
+### Static JSON File
```
lorenzolewis, line 135: If this is the easier of the two I'd move this section above the Dynamic one

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
 
-The alternate update technique uses a plain JSON file, storing your update metadata on S3, gist, or another static file store. Tauri checks against the version field, and if the version of the running process is smaller than the reported one of the JSON and the platform is available, it triggers an update. The format of this file is detailed below:
+With this approach, Tauri will always request the same JSON file and determine if the app needs to be updated by comparing the version field of the response with the current version. Tauri will expect a response in this format:
```
lorenzolewis:
```suggestion
With this approach, Tauri will always request the same JSON file and determine if the app needs to be updated by comparing the version field of the response with the requesting app's current version. Tauri will expect a response in this format:
```

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,128 @@
...
 
 ```json
```
lorenzolewis, line 139: Again, is this format anywhere else or only here?

docs/guides/distribution/updater.md
```diff
@@ -224,113 +132,103 @@ The alternate update technique uses a plain JSON file, storing your update metad
...
+### Built-in dialog
 
-```
-target/release/bundle
-└── macos
-    └── app.app
-    └── app.app.tar.gz (update bundle)
-    └── app.app.tar.gz.sig
-```
+By default, the updater uses a built-in dialog API from Tauri. The dialog will only check for a new update when the app was just launched or when you manually [emit](../../api/js/event.md#emit) the `"tauri://update"` event.
```
lorenzolewis: Maybe link to the dialog API?

docs/guides/distribution/updater.md
```diff
@@ -224,113 +132,103 @@ The alternate update technique uses a plain JSON file, storing your update metad
...
 
-### Windows
+![New Update](https://i.imgur.com/UMilB5A.png)
```
lorenzolewis: Host images in the repo instead of imgur please :)

docs/guides/distribution/updater.md
```diff
@@ -224,113 +132,103 @@ The alternate update technique uses a plain JSON file, storing your update metad
...
-    └── app.x64.msi.zip (update bundle)
-    └── app.x64.msi.zip.sig
-```
+### Custom Dialog and Logic
 
-### Linux
+:::caution
 
-On Linux, we create a .tar.gz from the AppImage.
+You need to disable the built-in dialog in your [Tauri configuration](#tauri-configuration) to enable the JavaScript APIs and updater events!
```
lorenzolewis, line 219: Is there a more specific place to point to here?

docs/guides/distribution/updater.md
```diff
@@ -224,113 +132,103 @@ The alternate update technique uses a plain JSON file, storing your update metad
...
-    └── app.AppImage.tar.gz (update bundle)
-    └── app.AppImage.tar.gz.sig
-```
+:::
 
-## Signing updates
+#### Rust
 
-We offer a built-in signature to ensure your update is safe to be installed.
+Please see the updater module documentation at [docs.rs] for the Rust API.
```
lorenzolewis, line 225: I'd link to the specific module

docs/guides/distribution/updater.md
```diff
@@ -224,113 +132,103 @@ The alternate update technique uses a plain JSON file, storing your update metad
...
 
-To sign your updates, you need two things.
+#### JavaScript
 
-The _Public-key_ (pubkey) should be added inside your `tauri.conf.json` to validate the update archive before installing.
+For the complete API docs see [here](../../api/js/updater.md). An example using the JS APIs could look like this:
```
lorenzolewis:
```suggestion
For the complete API docs see [here](../../api/js/updater.md). An example using the JavaScript API looks like this:
```

docs/guides/distribution/updater.md
```diff
@@ -224,113 +132,103 @@ The alternate update technique uses a plain JSON file, storing your update metad
...
+  if (shouldUpdate) {
+    // You could show a dialog asking the user if they want to install the update here.
+    console.log(
+      `Installing update ${manifest?.version}, ${manifest?.date}, ${manifest?.body}`
+    )
 
-```
-Generate keypair to sign files
+    // Install the update. This will also restart the app on Windows!
+    await installUpdate()
```
lorenzolewis, line 254: Technical question, but why is this different behaviour on platforms? If there's no reason then I think this should be opened as an issue in Tauri

lorenzolewis (changes requested):
You asked for a review ;) Overall it's a nice upgrade from the previous one!

docs/guides/distribution/updater.md
FabianLars: I think we're try to stick to `Capital Case`, just missed that there. But we never "officially" stated a guideline for that i think.

lorenzolewis: Would you mind sticking that in the style section in this PR? I'm thinking about pulling that into a separate, more detailed doc eventually but will be nice to start capturing things there

docs/guides/distribution/updater.md
FabianLars: Well, the actual reason was to force users to actually use the signature (used to be optional) :D Since it wasn't used as much as we liked and because we thought setting it up is fairly simple and quick to do we changed it to be required.

Not sure what to write in the docs though 🤔

docs/guides/distribution/updater.md
FabianLars: sure

docs/guides/distribution/updater.md
FabianLars: On windows the app needs to be closed to be able to install the new version with the MSI installer. I'm not sure if that can be worked around.

docs/guides/distribution/updater.md
FabianLars: yeah, same actually. Let's not talk about that i thought M comes before L in the alphabet ever again please 😂

docs/guides/distribution/updater.md
FabianLars: So, we somewhat regularly get asked if we can expose the built-in updater dialog so that devs can re-use that for their own updater flow (p.s. i agree that this is a good idea) and i wonder if linking to the normal dialog api would suggest that it's the same thing? I hope you get what i'm trying to say.

lorenzolewis: Yup makes sense. I'd rephrase to say "Tauri shows a dialog" or something a little more generic to hopefully reduce the confusion?

docs/guides/distribution/updater.md
FabianLars: Yes: https://tauri.app/v1/api/config#updaterconfig.
I don't think removing this section in favor of a link would be beneficial here though to be honest. I added a link to the config object above so hopefully that's enough of a cross ref?

docs/guides/distribution/updater.md
lorenzolewis: If that's the case then I'd just remove the last sentence completely.

docs/guides/distribution/updater.md
lorenzolewis: Doesn't really matter for this PR I guess, was just a weird inconsistency between OSs where `installUpdate()` DOES restart on Windows but then for macOS and Linux you have to do a separate command to get the same functionality 🤷‍♂️

Maybe something to think about for the API re-write

docs/guides/distribution/updater.md
lorenzolewis: Sounds good to me

static/img/guides/distribution/updater/update-available.png
lorenzolewis, line 1: Oh gosh this is an old app 😅 Maybe we could crop it to just the dialog?

FabianLars: haha, sure thing

FabianLars: Is there a way to give it a max size in markdown/docusaurus? It's a bit ugly if it gets scaled up that much

lorenzolewis: Hmmm not that I know of. Could fall back to an HTML image with a max width style property and that should work both here and the new site

docs/guides/distribution/updater.md
FabianLars: yeah, there are a lot of quirks in the updater that need a bit refinement.

docs/guides/distribution/updater.md
FabianLars: We do actually, it links to https://docs.rs/tauri/latest/tauri/updater/index.html

docs/guides/distribution/updater.md
FabianLars: Either the tauri config section in the guide, or the `dialog` property in config.md. In this case i'd be fine with both

docs/guides/distribution/updater.md
FabianLars: hmm, the problem with that is the amount of different ways to do that and platform inconsistencies 🤔

docs/guides/distribution/updater.md
FabianLars: Only here i think. The updater guide is somewhat of the source of truth for the server side stuff.

docs/guides/distribution/updater.md
FabianLars: Same as above.

docs/guides/distribution/updater.md
FabianLars: i added instructions to set the env var for the current session, is that ok? (just pushed it so may need a while to build)

FabianLars:
I just noticed that i didn't mention the windows specific install modes, gonna check if it makes sense to include them tomorrow

lorenzolewis:
@FabianLars is there any review you want on this one? Or still working on it?

FabianLars:
Yep review please. I think i addressed all comments now, either via code changes or by adding an answer to these comments.

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,139 @@
...
+# Updater
 
-Tauri is indifferent to the request the client application provides for update checking.
+Tauri offers a built-in updater for the MSI (Windows), AppImage (Linux) and App bundle (macOS) distribution formats.
 
-`Accept: application/json` is added to the request headers because Tauri is responsible for parsing the response.
+Once your Tauri project is ready, you can configure Tauri's updater to enable auto updating for your users.
 
-For the requirements imposed on the responses and the body format of an update, response see [Server Support](#server-support).
+## Signing updates
```
lorenzolewis:
```suggestion
## Signing Updates
```

Even added it to the contributing section 😂

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,139 @@
...
 
-![New Update](https://i.imgur.com/UMilB5A.png)
+If you are on Windows, you should use `$HOME/.tauri/myapp.key` or a different path of your choice instead:
 
-The dialog release notes are represented by the update `note` provided by the [server](#server-support).
-If the user accepts, the update is downloaded and installed. Afterward, the user is prompted to restart the application.
+<Command name="signer generate -w $HOME/.tauri/myapp.key" />
 
-### Javascript API
+## Tauri configuration
```
lorenzolewis:
```suggestion
## Tauri Configuration
```

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,139 @@
...
-#### Rust
+`"pubkey"` must be a valid public key generated with Tauri's CLI [above](#signing-updates).
 
-```rust
-window.emit("tauri://update".to_string(), None);
-```
+`"dialog"` if present must ba a boolean. By default it's set to true. If enabled, updater [events](#events) will be disabled as the built-in dialog handles everything. If you need custom events, you must turn off the built-in dialog.
 
-#### Javascript
+### Installmode on Windows
```
lorenzolewis:
```suggestion
### `Installmode` on Windows
```

docs/guides/distribution/updater.md
```diff
@@ -2,220 +2,139 @@
...
-```
+On Windows there is an additional config [`"installMode"`](../../api/config.md#updaterwindowsconfig.installmode) to change how the update is installed.
 
-#### Listen to the New Update Available event
+- `"passive"`: The default and general recommendation. It will show a small progress bar when installing the update without requiring any user interaction.
+- `"basicUI"`: There will be a basic user interface shown which requires user interaction to finish the installation.
+- `"quiet"`: There will be no progress feedback to the user. With this mode the installer cannot request admin privileges by itself so it only works in user-wide installations or when your app itself already runs with admin privileges. Generally not recommended.
 
-Event: `tauri://update-available`
+## Update artifacts
```
lorenzolewis:
```suggestion
## Update Artifacts
```

docs/guides/distribution/updater.md
```diff
@@ -224,113 +143,131 @@ The alternate update technique uses a plain JSON file, storing your update metad
...
-target/release/bundle
-└── msi
-    └── app.x64.msi
-    └── app.x64.msi.zip (update bundle)
-    └── app.x64.msi.zip.sig
-```
+## Checking for Updates
 
-### Linux
+### Built-in dialog
```
lorenzolewis:
```suggestion
### Built-In Dialog
```

docs/guides/distribution/updater.md
```diff
@@ -224,113 +143,131 @@ The alternate update technique uses a plain JSON file, storing your update metad
...
-    └── app.AppImage.tar.gz.sig
-```
+<img src={updaterDialogImage} alt="Default updater dialog" style={{maxHeight: "200px", textAlign: 'center'}} />
 
-## Signing updates
+The dialog release notes are represented by the update `notes` provided by the [server](#server-support).
+If the user accepts, the update is downloaded and installed. Afterwards, the user is prompted to restart the application.
 
-We offer a built-in signature to ensure your update is safe to be installed.
+### Custom Dialog and Logic
```
lorenzolewis:
```suggestion
### Custom Dialog
```

lorenzolewis (commented):
Those dang headers

docs/guides/distribution/updater.md
FabianLars: fml

FabianLars: eeehhhh, i mean, i obviously just wanted to check how thorough you review my PRs 😂

lorenzolewis: Haha you caught me on a good day 😂

lorenzolewis (approved):

PR #1155, opened 2023-03-10
Added information about the use of Cargo.toml if the version field

simonhyll:
Saw this functionality mentioned in [#4643](https://github.com/tauri-apps/tauri/issues/4643) and figured I'd amend it.


FabianLars:
This MD file is auto generated from the main repo so if we want this change to persist between updates you should open a PR upstream too :)

lorenzolewis:
Going to close this one out. @simonhyll I believe this is the file upstream if you want to open a PR against it with this change: https://github.com/tauri-apps/tauri/blob/e8e214b72ede33ca2cc09fa41a8baf56fa7e8b8e/core/tauri-utils/src/config.rs#L2870

PR #1158, opened 2023-03-13
chore: update deps in next

amrbashir:
I am not sure what is the opensuse equivalent for webkit2gtk-4.1 so I left it as is for now.


FabianLars (approved):
thank you :)

PR #1153, opened 2023-03-09
doc: update install command to use secure curl, with fish-compatible pipe

coolaj86:
and updated the blog too

PR #1162, opened 2023-03-15
Revert "doc: update install command to use secure curl, with fish-compatible pipe (#1153)"

FabianLars:
This reverts commit 6e1deeeae9e6847ca97356f1e8f107d79ec0f284 because the new instructions don't work on macOS, we should look for a new instruction layout in the new docs to reintroduce this change for Linux users if possible.

### DIFF
--- a/blog/2023-03-01-create-tauri-app-version-3-released.mdx
+++ b/blog/2023-03-01-create-tauri-app-version-3-released.mdx
@@ -35,7 +35,7 @@ cargo install create-tauri-app
 cargo create-tauri-app --alpha
 
 # Bash
-curl https://create.tauri.app/sh | sh -s -- --alpha
+sh <(curl https://create.tauri.app/sh) --alpha
 
 # Powershell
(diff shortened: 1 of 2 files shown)


lorenzolewis:
Looks good. I made a tracking issue upstream to try to take a deeper look at this: https://github.com/tauri-apps/create-tauri-app/issues/375

CC: @coolaj86 since they made the original PR

lorenzolewis (approved):

coolaj86:
What part doesn't work on macOS? Is it just BSD curl not supporting some of those options?

Looks like yes. Let's just remove that part and keep the posix-style shell exec.

lorenzolewis:
I'm not too sure what the root error is for it (not sure if we had a chance to dig in tbh)

lorenzolewis:
But feel free to take a look at https://github.com/tauri-apps/create-tauri-app/issues/375 if that's something you'd like to help us out with <3

PR #1163, opened 2023-03-16
Update command.md about async

hackers267:
Add a note about async command which takes arguments

### DIFF
--- a/docs/guides/features/command.md
+++ b/docs/guides/features/command.md
@@ -172,11 +172,10 @@ Since invoking the command from JS already returns a promise, it works just like
 invoke('my_custom_command').then(() => console.log('Completed!'))
 ```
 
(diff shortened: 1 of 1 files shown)


FabianLars:
Thanks for contributing!

Technically this is only required if we're talking about borrowed arguments (including tauri::State) or like arguments with a non-static lifetime. @lorenzolewis Do you think changing the wording to `may require a Result` (together with the error message at the end) would convey that better?

lorenzolewis:
Hmm, so I'm thinking of the "happy path" here. I think it's best to explain this directly to a beginner where they might not have concepts of lifetimes, etc. It's easier to tell them to use `Result<>` everywhere (whether it's needed or not). Then use one of these boxes to highlight those cases where it isn't required (for those more advanced readers to reference).

FabianLars:
Superseded by #1214, thank you again though!!

PR #1165, opened 2023-03-16
Update system-tray.md

GolDNenex:
I've changed the example for "Keep the app running in the background after closing all windows". The old one was not usable because the way of detecting event have changed since.

### DIFF
--- a/docs/guides/features/system-tray.md
+++ b/docs/guides/features/system-tray.md
@@ -193,10 +193,31 @@ Note that you need to add `icon-ico` or `icon-png` feature flag to the tauri dep
 app.tray_handle().set_icon(tauri::Icon::Raw(include_bytes!("../path/to/myicon.ico").to_vec())).unwrap();
 ```
 
-### Keep the app running in the background after closing all windows
+### Preventing the app from closing when they are no windows left
 
-By default, tauri closes the application when the last window is closed.
-If your app should run in the background, you can call `api.prevent_close()` like so:
+By default, tauri closes the application when the last window is closed. We can simply call `api.prevent_close()` to prevent this.
+
+Depending of your need you can use one of the two following exemples:
+
+#### Keeping the backend running in the background after closing all windows
+
+If your backend should run in the background, you can call `api.prevent_close()` like so:
+
+```rust
+tauri::Builder::default()
+  .build(tauri::generate_context!())
+  .expect("error while building tauri application")
+  .run(|_app_handle, event| match event {
+    tauri::RunEvent::ExitRequested { api, .. } => {
+      api.prevent_exit();
+    }
+    _ => {}
+  });
+```
+
+#### Keeping the app running in the background after closing all windows
+
+In case you also need to keep the frontend running in the background, this can be achieved like this:
 
 ```rust
 tauri::Builder::default().on_window_event(|event| match event.event() {
--- a/docs/guides/features/system-tray.md
+++ b/docs/guides/features/system-tray.md
@@ -197,7 +197,7 @@ app.tray_handle().set_icon(tauri::Icon::Raw(include_bytes!("../path/to/myicon.ic
 
 By default, Tauri closes the application when the last window is closed. You can simply call `api.prevent_close()` to prevent this.
 
-Depending of your need you can use one of the two following exemples:
+Depending on your needs you can use one of the two following options:
 
 #### Keeping the Backend Running in the Background
 
--- a/docs/guides/features/system-tray.md
+++ b/docs/guides/features/system-tray.md
@@ -195,7 +195,7 @@ app.tray_handle().set_icon(tauri::Icon::Raw(include_bytes!("../path/to/myicon.ic
 
 ### Preventing the App from Closing
 
-By default, tauri closes the application when the last window is closed. We can simply call `api.prevent_close()` to prevent this.
+By default, Tauri closes the application when the last window is closed. You can simply call `api.prevent_close()` to prevent this.
 
 Depending of your need you can use one of the two following exemples:
 
--- a/docs/guides/features/system-tray.md
+++ b/docs/guides/features/system-tray.md
@@ -199,15 +199,15 @@ By default, tauri closes the application when the last window is closed.
 If your app should run in the background, you can call `api.prevent_close()` like so:
 
 ```rust
-tauri::Builder::default()
-  .build(tauri::generate_context!())
-  .expect("error while building tauri application")
-  .run(|_app_handle, event| match event {
-    tauri::RunEvent::ExitRequested { api, .. } => {
-      api.prevent_exit();
-    }
-    _ => {}
-  });
+tauri::Builder::default().on_window_event(|event| match event.event() {
+  tauri::WindowEvent::CloseRequested { api, .. } => {
+    event.window().hide().unwrap();
+    api.prevent_close();
+  }
+  _ => {}
+})
+.run(tauri::generate_context!())
+.expect("error while running tauri application");
 ```
 
 [template image]: https://developer.apple.com/documentation/appkit/nsimage/1520017-template?language=objc
--- a/docs/guides/features/system-tray.md
+++ b/docs/guides/features/system-tray.md
@@ -199,7 +199,7 @@ By default, tauri closes the application when the last window is closed. We can 
 
 Depending of your need you can use one of the two following exemples:
 
-#### Keeping the backend running in the background after closing all windows
+#### Keeping the Backend Running in the Background
 
 If your backend should run in the background, you can call `api.prevent_close()` like so:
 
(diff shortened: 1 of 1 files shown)


FabianLars:
Thanks for contributing!

The "old" example is still correct. It just has a different goal. The current example only keeps the rust runtime going, while your code also keeps the window open (so that you can keep using JS).
Since this section is explicitly about keeping the app running when all windows _closed_ i think the current docs are more correct, _but_ this page is honestly in need of some kind of rewrite and then it should definitely show your variant too!

GolDNenex:
Oh, i haven't think about the difference between close and hide because a user close the windows for hiding it. My bad, its really a different behavior. Since i was following the SystemTrayMenu example (with close and hide/show), for some reason that was logical at that time in my head.

I'm relatively new to rust and Tauri. Trying to use the "old" example throw me a error:
image

After some reseach, from what i understand we supposed to use "on_window_event". Did i miss something ?

FabianLars:
I think the issue is that you reuse the `run` method that was in the project template. instead you need to use `.build` (which takes the generate_context!() input) and then add a different kind of `run` behind that. 

sorry that explanation was really bad, just compare the old example with your code again 😅

GolDNenex:
Sorry, yesterday when I was trying the example I was stupid.

Just after sending the message, i notice that ". Expect (" Error While Running Tauri Application ");" was not supposed to be here. Removing it, fix the error.

Thanks for your replies @FabianLars !

In my noble opinion, the current method seem more adapted for a general handling of event and in this context we should use "on_window_event" in this part of the docs.

FabianLars:
> In my noble opinion, the current method seem more adapted for a general handling of event and in this context we should use "on_window_event" in this part of the docs. This could be a opportunity for adding a new page for events handling.

Yeah, i'm really unsure about this tbh. It's basically a different use-case, like preventing _window_ from closing vs preventing the _app_ from closing. Both have pro and cons, but with the current layout of this doc page, as in being focused on the rust side, it feels like the window version would be the wrong choice since you'd mainly do that to keep the js runtime running, which obviously wastes resources if you don't actually use it.

So yeah, imo we really need to show both variants, telling users the pros and cons for each approach, and preferably how they can "recover" from them (= hot to show the app again)

----

I'd like to have an opinion from other team and/or community members on this one though so i'll leave this PR open/unmerged for a while if that's okay :)

lorenzolewis:
I'd agree with showing both options. But I'd be sure to include the points that @FabianLars mentioned (one only keeping the Rust runtime running, the other also keeping the frontend alive).

Especially with this heading:
"Keep the app running in the background after closing all windows"

The word "_app_" to me means both the Rust backend and the JS frontend. When you cover keeping ONLY the Rust backend running then I'd call that heading "Keep the backend running in the background after closing all windows" (we can work on a more succinct header). 

For the earlier point of close vs. hide, keep in mind those are different things on macOS for example. "Close" exits that window and **destroys** the UI state, whereas "Hide" will hide the window but **preserve** the UI state.

GolDNenex:
Let me know what you think about my change. 
Also, the english is not my first language so don't hesitate to tell me if something is wrong.

Edit: I've used  `####` for the sub-categories like above in the page but they blend to much since they are longer. Not really sure how to improve that. Maybe adding a number a the start ?

docs/guides/features/system-tray.md
```diff
@@ -193,10 +193,31 @@ Note that you need to add `icon-ico` or `icon-png` feature flag to the tauri dep
 app.tray_handle().set_icon(tauri::Icon::Raw(include_bytes!("../path/to/myicon.ico").to_vec())).unwrap();
 ```
 
-### Keep the app running in the background after closing all windows
+### Preventing the app from closing when they are no windows left
```
FabianLars:
```suggestion
### Preventing the App from Closing
```

docs/guides/features/system-tray.md
```diff
@@ -193,10 +193,31 @@ Note that you need to add `icon-ico` or `icon-png` feature flag to the tauri dep
...
 
-By default, tauri closes the application when the last window is closed.
-If your app should run in the background, you can call `api.prevent_close()` like so:
+By default, tauri closes the application when the last window is closed. We can simply call `api.prevent_close()` to prevent this.
+
+Depending of your need you can use one of the two following exemples:
+
+#### Keeping the backend running in the background after closing all windows
```
FabianLars:
```suggestion
#### Keeping the Backend Running in the Background
```

docs/guides/features/system-tray.md
```diff
@@ -193,10 +193,31 @@ Note that you need to add `icon-ico` or `icon-png` feature flag to the tauri dep
...
+  .expect("error while building tauri application")
+  .run(|_app_handle, event| match event {
+    tauri::RunEvent::ExitRequested { api, .. } => {
+      api.prevent_exit();
+    }
+    _ => {}
+  });
+```
+
+#### Keeping the app running in the background after closing all windows
```
FabianLars:
```suggestion
#### Keeping the Frontend Running in the Background
```

FabianLars:
> Also, the english is not my first language so don't hesitate to tell me if something is wrong.

Let us hear @lorenzolewis thoughts :D

docs/guides/features/system-tray.md
```diff
@@ -193,10 +193,15 @@ Note that you need to add `icon-ico` or `icon-png` feature flag to the tauri dep
 app.tray_handle().set_icon(tauri::Icon::Raw(include_bytes!("../path/to/myicon.ico").to_vec())).unwrap();
 ```
 
-### Keep the app running in the background after closing all windows
+### Preventing the App from Closing
 
-By default, tauri closes the application when the last window is closed.
-If your app should run in the background, you can call `api.prevent_close()` like so:
+By default, tauri closes the application when the last window is closed. We can simply call `api.prevent_close()` to prevent this.
```
lorenzolewis:
```suggestion
By default, Tauri closes the application when the last window is closed. You can simply call `api.prevent_close()` to prevent this.
```

docs/guides/features/system-tray.md
```diff
@@ -193,10 +193,15 @@ Note that you need to add `icon-ico` or `icon-png` feature flag to the tauri dep
...
+
+Depending of your need you can use one of the two following exemples:
```
lorenzolewis:
```suggestion
Depending on your needs you can use one of the two following options:
```

docs/guides/features/system-tray.md
```diff
@@ -193,10 +193,15 @@ Note that you need to add `icon-ico` or `icon-png` feature flag to the tauri dep
...
+
+#### Keeping the Backend Running in the Background
```
lorenzolewis:
```suggestion
**Keep the Backend Running in the Background**
```

docs/guides/features/system-tray.md
```diff
@@ -210,5 +215,21 @@ tauri::Builder::default()
   });
 ```
 
+#### Keeping the Frontend Running in the Background
```
lorenzolewis:
```suggestion
**Keep the Frontend Running in the Background**
```

docs/guides/features/system-tray.md
```diff
@@ -210,5 +215,21 @@ tauri::Builder::default()
...
+
+In case you also need to keep the frontend running in the background, this can be achieved like this:
```
lorenzolewis:
```suggestion
If you need to keep the frontend running in the background, this can be achieved like this:
```

lorenzolewis (changes requested):
Overall awesome, just a few very minor tweaks. Thanks a bunch!

docs/guides/features/system-tray.md
lorenzolewis: Changing to a bold instead of heading should hopefully help a bit (but admittedly I didn't preview it live)

FabianLars:
i think i preferred the small headings. i can't see a difference between bold and normal text on firefox@linux :laughing: 
grafik

GolDNenex:
Looks good on my side. Thanks you both for the help.

Edit: I don't see any difference to be honest. But since they are shorter, they do more they're job of spliting.

Edit 2: Look like they are a 1px difference with the font, minus 1px with the bolded one.

lorenzolewis:
I'm fine with whichever goes in. All the fonts and weights will be re-done in the new site anyway. @FabianLars are we fine to go ahead and merge this one in you think?

FabianLars:
i'd prefer the headings since that's what they effectively are here but i don't care too much about it so merge whenever you want ig

FabianLars (approved):

lorenzolewis (approved):

lorenzolewis:
Merged in. Thank you so much @GolDNenex ! <3

PR #1183, opened 2023-03-25
Fix typos and make other small changes

elliotwaite:
If any of these changes are incorrect or unwanted, let me know and I can update the PR.

Some of these changes are specific to the new mobile docs, which is why this is a PR for the "next" branch. But if for some reason this should not be a PR for the "next" branch, let me know.


FabianLars:
First of all, thanks for the awesome PR! A few things:

1) The dev branch would've indeed be the right call (with a second PR targeting the next branch with next specific changes) but we know that this whole setup is super confusing rn
2) Technically we require all commits to be [signed](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits), though i tend to be quite liberal about this for docs changes.
3) the config.md file is auto-generated from upstream, meaning that the change should happen here instead: <https://github.com/tauri-apps/tauri/blob/dev/core/tauri-utils/src/config.rs#L509-L517> (also targeting dev)
4) There may be one or two merge conflicts if i update next with all the changes from dev (thinking about the updater guide)

Since this change is quite big i'm fine with cherry-picking it onto dev myself if you can rebase it onto the latest next commit (`git merge next` works too, no need for an actual rebase)

docs/guides/testing/webdriver/example/selenium.md
```diff
@@ -245,7 +245,7 @@ $ Mocha
 Done in 0.93s.
 ```
 
-We can see that our `Hello Tauri` sweet we created with `decribe` had all 3 items we created with `it` pass their
+We can see that our `Hello Tauri` test suite we created with `describe` had all 3 items we created with `it` pass their
```
FabianLars, line 248: i love that one 😂

elliotwaite:
Okay, thanks for being okay with cherry-picking it. I'll use the dev branch for general changes next time.

I just merged the next branch onto this commit. Let me know if there is anything else I need to do. I can figure out the signing if needed.

FabianLars (approved):

FabianLars:
(@ Lorenzo just a quick scan is enough, don't want to take too much time away from the 2.0 branch)

elliotwaite:
And I opened a new PR on the tauri repo for the config.md related changes: https://github.com/tauri-apps/tauri/pull/6561

lorenzolewis:
Thanks for the awesome PR @elliotwaite ! Got this in just in time as I'm going to be taking a cut in a few days to port over to the new website branch.

Looks good to me, @FabianLars I'll give you a shout before I start porting changes over to make sure you've had a chance to cherry pick.

lorenzolewis (approved):

PR #1189, opened 2023-03-29
feat: add android apk/aab signing guide, remove standalone android setup

amrbashir:
This is heavily based on the flutter signing guide but fits perfectly for tauri-apps as well. It is probably fine to merge this now but I think we should wait until https://github.com/tauri-apps/tauri-mobile/pull/124 is merged and new version of the tauri CLI is published.

This PR also removes the standalone setup for Android because it is more error-prone and requires continuous maintenance of the latest CLI urls and versions.


FabianLars:
I feel like we should keep that info around somewhere easily accessible, i think it still has valid use-cases especially in CI. But let's worry about that later ig.

PR #1210, opened 2023-04-16
Docs: Add prerequisites for Gentoo Linux

rivnakm:
webkit-gtk needs to be emerged with a certain slot so adding this would be helpful


lorenzolewis (approved):
Awesome, thank you so much for putting this in! 🙌

PR #1213, opened 2023-04-17
Update next.js gettting-started since Next.js 13.3.0 now supports static export

irg1008:
Since ~~23.0.0~~ 13.3.0, Next.js now supports static exports with `next export`.

Source: https://beta.nextjs.org/docs/app-directory-roadmap#configuration
Details: https://beta.nextjs.org/docs/configuring/static-export

We can now update the docs to reflect this change and let the user now that appDir is not a viable option

### DIFF
--- a/docs/guides/getting-started/setup/next-js.mdx
+++ b/docs/guides/getting-started/setup/next-js.mdx
@@ -61,7 +61,7 @@ pnpm create next-app --use-pnpm --typescript
 
 2. _Experimental `app/` directory_  
    Next.js will ask you if you want to try the new and experimental `app/` directory.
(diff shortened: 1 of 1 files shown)


irg1008:
I mispelled 13.3.0 for 23.0.0

FabianLars:
Thanks for contributing!

I feel like the note should be a bit more specific, making it explicit that 13.3 is required for the app dir, but old-school nextjs apps worked before that. Not the best wording but i mean something like this: `This feature requires at least Next.js version 13.3.0 for static exports required for Tauri`, i hope you understand what i mean and can think of something better than i can 😅

irg1008:
Absolutely, that works for me but I will try to come up with a better phrase!

PR #1214, opened 2023-04-18
Include information about borrowed values in command.md

IntegralPilot:
This pull request updates the part of command.md about asynchronous functions to include an important note about how borrowed values like &str are not allowed at present, and updates the code sample to include an example of using a String instead as a workaround.

Helps resolve `https://github.com/tauri-apps/tauri/issues/6733`.

### DIFF
--- a/docs/guides/features/command.md
+++ b/docs/guides/features/command.md
@@ -190,7 +190,7 @@ When working with borrowed types, you have to make additional changes. These are
 *Example:*
 
 ```rust
-// Declare the constructor using String instead of &str, as &str is borrowed and thus unsupported
+// Declare the async function using String instead of &str, as &str is borrowed and thus unsupported
 #[tauri::command]
 async fn my_custom_command(value: String) {
   // Call another async function and wait for it to finish
--- a/docs/guides/features/command.md
+++ b/docs/guides/features/command.md
@@ -176,15 +176,21 @@ Commands without the _async_ keyword are executed on the main thread unless defi
 
 :::note
 
-You need to be careful when creating asynchronous functions using Tauri. Currently, you cannot include types that contain the & symbol (indicating they are borrowed) in the constructor of an asynchronous function. A common example of a type like this is _&str_.
+You need to be careful when creating asynchronous functions using Tauri. Currently, you cannot include types that are borrowed, such as those with the & symbol, in the constructor of an asynchronous function. Some common examples of types like this are `&str` and `State<'_, Data>`.
 
 :::
 
 
-If your command needs to run asynchronously, simply declare it as `async`, and convert all borrowed types to their non-borrowed counterparts:
+If your command needs to run asynchronously, simply declare it as `async`, and resolve any issues with borrowed types.
+
+When working with borrowed types, there are two main options.
+
+**Option 1**: Convert the type, such as `&str` to a similar type that is not borrowed, such as `String`. This won't work for all types.
+
+*Example:*
 
 ```rust
-// declare the constructor using String instead of &str, as &str is borrowed and thus unsupported
+// Declare the constructor using String instead of &str, as &str is borrowed and thus unsupported
 #[tauri::command]
 async fn my_custom_command(value: String) {
   // Call another async function and wait for it to finish
@@ -193,7 +199,23 @@ async fn my_custom_command(value: String) {
 }
 ```
 
-Since invoking the command from JS already returns a promise, it works just like any other command:
+**Option 2**: Wrap the return type in a Result. This one is a bit harder to implement, but works for all types.
+
+If you have no return type, like in our function, use the return type `Result<(), ()>`. Otherwise, use the return type `Result<bool, ()>`, replacing `bool` with whatever type you wish to return.
+
+*Example:*
+
+```rust
+// Return a Result<(), ()> to bypass the borrowing issue
+#[tauri::command]
+async fn my_custom_command(value: &str) -> Result<(), ()> {
+  // Call another async function and wait for it to finish
+  let result = some_async_function(value).await;
+  println!("Result: {}", result);
+}
+```
+
+**Invoking from JS:** Since invoking the command from JS already returns a promise, it works just like any other command:
 
 ```js
 invoke('my_custom_command', {value: "Hello, Async!"}).then(() => console.log('Completed!'))
--- a/docs/guides/features/command.md
+++ b/docs/guides/features/command.md
@@ -164,6 +164,8 @@ It also gives you full control over the way your error type gets serialized. In 
 
 ## Async Commands
 
+Asynchornous functions are benefical in Tauri to perform heavy work in a manner that doesn't result in UI freezes or slowdowns.
+
 :::note
 
 Async commands are executed on a separate thread using [`async_runtime::spawn`].
@@ -171,13 +173,23 @@ Commands without the _async_ keyword are executed on the main thread unless defi
 
 :::
 
+
+
+:::note
+
+You need to be careful when creating asynchornous functions using Tauri. Currently, you cannot include types that contain the & symbol (indicating they are borrowed) in the constructor of an asynchornous function. A common example of a type like this is _&str_.
+
+:::
+
+
 If your command needs to run asynchronously, simply declare it as `async`:
 
 ```rust
+// declare the constructor using String instead of &str, as &str is unsupported
 #[tauri::command]
-async fn my_custom_command() {
+async fn my_custom_command(value: String) {
   // Call another async function and wait for it to finish
-  let result = some_async_function().await;
+  let result = some_async_function(value).await;
   println!("Result: {}", result);
 }
 ```
@@ -185,7 +197,7 @@ async fn my_custom_command() {
 Since invoking the command from JS already returns a promise, it works just like any other command:
 
 ```js
-invoke('my_custom_command').then(() => console.log('Completed!'))
+invoke('my_custom_command', {value: "Hello, Async!"}).then(() => console.log('Completed!'))
 ```
 
 ## Accessing the Window in Commands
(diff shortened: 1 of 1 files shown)


FabianLars:
fwiw here's a similar PR: <https://github.com/tauri-apps/tauri-docs/pull/1163> - i'll take a look as soon as i can.

IntegralPilot:
oops... made a typo in the word Asynchronous, trying to figure out how to add a new commit in the same PR

*edit: figured out, all is fixed now*

IntegralPilot:
Fixed grammatical mistakes! Also changed the page title for command.md from _Calling Rust from the frontend_ to _Communication between Rust and the Frontend_ as it appears more relevant.

IntegralPilot:
Okay! Just read another issue on Tauri main, and it turns out this effects more than just types with the & like my documentation suggests. I'm going to fix that now, and maybe include an example of the result wrapping...

*done! also prettified some of my markdown formatting, i'll just check after the build, but I think this is nearly ready to merge!*

IntegralPilot:
Build checked, I'm really happy with my changes, I think this PR is finalised from my end now. 

Hopefully, when reviewed and merged, this will help save people some debugging time on the borrowing-async issue.

docs/guides/features/command.md
```diff
@@ -1,4 +1,4 @@
-# Calling Rust from the frontend
+# Communication between Rust and the Frontend
```
FabianLars:
```suggestion
# Calling Rust from the frontend
```

Please revert this for now as it doesn't look good in the sidebar since it's the only item taking up multiple lines:
grafik

We could revisit this in the new docs in the future.

docs/guides/features/command.md
```diff
@@ -164,28 +164,61 @@ It also gives you full control over the way your error type gets serialized. In
...
 Async commands are executed on a separate thread using [`async_runtime::spawn`].
 Commands without the _async_ keyword are executed on the main thread unless defined with _#[tauri::command(async)]_.
 
 :::
 
-If your command needs to run asynchronously, simply declare it as `async`:
+
+:::note
+
+You need to be careful when creating asynchronous functions using Tauri. Currently, you cannot include types that are borrowed, such as those with the & symbol, in the constructor of an asynchronous function. Some common examples of types like this are `&str` and `State<'_, Data>`.
```
FabianLars:
```suggestion
You need to be careful when creating asynchronous functions using Tauri. Currently, you cannot include borrowed arguments in the signature of an asynchronous function. Some common examples of types like this are `&str` and `State<'_, Data>`.
```

My main issue with this one is `such as those with the & symbol` and `constructor`. "constructor" is just wrong in this context afaik, and "such as those with the & symbol" sounds a bit too weird imo, even though i see how that would help beginners understand what we're talking about.

docs/guides/features/command.md
```diff
@@ -164,28 +164,61 @@ It also gives you full control over the way your error type gets serialized. In
...
+
+:::
+
+
+If your command needs to run asynchronously, simply declare it as `async`, and resolve any issues with borrowed types.
+
+When working with borrowed types, there are two main options.
```
FabianLars:
```suggestion
If your command needs to run asynchronously, simply declare it as `async`.

When working with borrowed types, you have to make additional changes. These are your two main options:
```

Not quite happy with my suggestion but i hope you're getting the idea.

docs/guides/features/command.md
```diff
@@ -164,28 +164,61 @@ It also gives you full control over the way your error type gets serialized. In
...
+
+**Option 1**: Convert the type, such as `&str` to a similar type that is not borrowed, such as `String`. This won't work for all types.
```
FabianLars:
```suggestion
**Option 1**: Convert the type, such as `&str` to a similar type that is not borrowed, such as `String`. This may not work for all types.
```

docs/guides/features/command.md
```diff
@@ -164,28 +164,61 @@ It also gives you full control over the way your error type gets serialized. In
...
+
+*Example:*
+
+```rust
+// Declare the constructor using String instead of &str, as &str is borrowed and thus unsupported
```
FabianLars:
```suggestion
// Declare the async function using String instead of &str, as &str is borrowed and thus unsupported
```

docs/guides/features/command.md
```diff
@@ -164,28 +164,61 @@ It also gives you full control over the way your error type gets serialized. In
...
+#[tauri::command]
+async fn my_custom_command(value: String) {
+  // Call another async function and wait for it to finish
+  let result = some_async_function(value).await;
+  println!("Result: {}", result);
+}
+```
+
+**Option 2**: Wrap the return type in a Result. This one is a bit harder to implement, but works for all types.
```
FabianLars:
```suggestion
**Option 2**: Wrap the return type in a Result. This one is a bit harder to implement, but should work for all types.
```

docs/guides/features/command.md
```diff
@@ -164,28 +164,61 @@ It also gives you full control over the way your error type gets serialized. In
...
+
+If you have no return type, like in our function, use the return type `Result<(), ()>`. Otherwise, use the return type `Result<bool, ()>`, replacing `bool` with whatever type you wish to return.
```
FabianLars: I think we need to be a bit more explicit that this is only if you don't want to return any actual errors from the command.

docs/guides/features/command.md
```diff
@@ -164,28 +164,61 @@ It also gives you full control over the way your error type gets serialized. In
...
 :::note
 
 Async commands are executed on a separate thread using [`async_runtime::spawn`].
 Commands without the _async_ keyword are executed on the main thread unless defined with _#[tauri::command(async)]_.
 
 :::
 
-If your command needs to run asynchronously, simply declare it as `async`:
+
+:::note
```
FabianLars:
```suggestion
:::warning
```

Let it take the user's attention :D

docs/guides/features/command.md
```diff
@@ -164,28 +164,61 @@ It also gives you full control over the way your error type gets serialized. In
...
   println!("Result: {}", result);
 }
 ```
 
-Since invoking the command from JS already returns a promise, it works just like any other command:
+**Invoking from JS:** Since invoking the command from JS already returns a promise, it works just like any other command:
 
 ```js
-invoke('my_custom_command').then(() => console.log('Completed!'))
+invoke('my_custom_command', {value: "Hello, Async!"}).then(() => console.log('Completed!'))
```
FabianLars:
```suggestion
invoke('my_custom_command', { value: "Hello, Async!" }).then(() => console.log('Completed!'))
```
I know that's really picky but i have to 😂

docs/guides/features/command.md
```diff
@@ -164,28 +164,61 @@ It also gives you full control over the way your error type gets serialized. In
...
+async fn my_custom_command(value: &str) -> Result<(), ()> {
   // Call another async function and wait for it to finish
-  let result = some_async_function().await;
+  let result = some_async_function(value).await;
   println!("Result: {}", result);
 }
 ```
 
-Since invoking the command from JS already returns a promise, it works just like any other command:
+**Invoking from JS:** Since invoking the command from JS already returns a promise, it works just like any other command:
```
FabianLars:
```suggestion
#### Invoking from JS

Since invoking the command from JS already returns a promise, it works just like any other command:
```

FabianLars (changes requested):
Thanks for your contribution!

While this is imo is almost good to merge as-is i still went ahead and added a few pretty picky comments on stuff i'd like to talk about. Please take a look at them and see what makes sense to you and what doesn't (not all of my suggestions must be applied!)

docs/guides/features/command.md
IntegralPilot: Yep, didn't notice that, thanks. Changing title back now...

docs/guides/features/command.md
IntegralPilot: Definitely, sounds a lot better than fluid. I think `constructor` is the name for function signatures in Kotlin, got my language lingo a bit mixed up there, thanks.

IntegralPilot: Commited!

docs/guides/features/command.md
IntegralPilot: Committed!

docs/guides/features/command.md
IntegralPilot: I think the suggestion is great, makes it more fluid. :D

IntegralPilot: Committed!

docs/guides/features/command.md
IntegralPilot: I think may not works better, thanks. Committed!

docs/guides/features/command.md
IntegralPilot: Again, got my Kotlin and my Rust lingo mixed up, constructor definitely doesn't fit well there, thanks. Committed!

docs/guides/features/command.md
IntegralPilot: Again, should work sounds a lot better. Committed! Thanks

docs/guides/features/command.md
IntegralPilot: :D

docs/guides/features/command.md
IntegralPilot: That markdown looks better, thanks.

docs/guides/features/command.md
IntegralPilot: It does look nicer...

docs/guides/features/command.md
IntegralPilot: Hmm... I've fixed that just now, but I think my English is a bit clunky.

IntegralPilot:
Thanks for all your help! I've applied all your changes, I think they were great. I've tried to explain the return types a bit better, but I think my English was a bit clunky.

FabianLars (approved):
Thanks again!!

IntegralPilot:
Wow, just read the latest commits from you before merge, and they are great!

Thank you so much for helping to make this PR so much better, its being lots of fun contributing back to one of the open source projects I love.

PR #1224, opened 2023-04-28
Update mocking.md: fixed sample code to avoid runtime type error

shotanue:
Updated sample code in documentation to avoid runtime type error 'Cannot set property' due to read-only window.crypto.

`window.crypto` appears to be read-only, so running this code throws a TypeError at runtime, as shown below:

スクリーンショット 2023-04-29 3 11 01
https://www.typescriptlang.org/play?#code/PTAEEsFsAcHsCcAuoDe8CGA7AJrSAxcAGyIGUBPTAYwF9QAzePUAcivnOkVhYCheQoAAKIAzgFpwAc0wIAprwDu4HLEUA6dp26gAvKl6hQgkROmz4Co1LmIASllyQAauiIBXOaIBcoABQAlHoAfKg0vOFAA

While I handling to this issue in my project, I found `Object.defineProperty` can prevent this runtime error. How about this fix for the sample code?

### DIFF
--- a/docs/guides/testing/mocking.md
+++ b/docs/guides/testing/mocking.md
@@ -31,12 +31,14 @@ import { invoke } from "@tauri-apps/api/tauri";
 
 // jsdom doesn't come with a WebCrypto implementation
 beforeAll(() => {
-  //@ts-ignore
-  window.crypto = {
-    getRandomValues: function (buffer) {
-      return randomFillSync(buffer);
-    },
-  };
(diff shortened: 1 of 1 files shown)


FabianLars:
I'd imagine this is because of the browser environment. I've not seen this issue in actual test runners yet. That said, i'm ok with merging it. Thanks for contributing :)

FabianLars (approved):

Issue #1226, opened 2023-04-29
write guides for nsis


FabianLars:
https://github.com/tauri-apps/tauri/pull/5788

PR #1228, opened 2023-05-02
Add some initial NSIS documentation

FabianLars:
ref: #1226 

Edit: We're waiting for the 1.3 release to merge this so that we don't push docs with broken links :)

### DIFF
--- a/docs/guides/building/windows.md
+++ b/docs/guides/building/windows.md
@@ -264,7 +264,7 @@ Note that `ComponentGroup`, `Component`, `FeatureGroup`, `Feature` and `Merge` e
 
 ## Internationalization
 
-The NSIS Installer is a multi-language installer, which means you always have a single installer which contains all the selected translations. You can specify which versions to include using the [`tauri.bundle.windows.nsis.languages`](../../api/config.md#nsisconfig.languages) property.
+The NSIS Installer is a multi-language installer, which means you always have a single installer which contains all the selected translations. You can specify which versions to include using the [`tauri.bundle.windows.nsis.languages`](../../api/config.md#nsisconfig.languages) property. A list of languages supported by NSIS is available in [the NSIS GitHub project]. There are a few [Tauri-specific translations] required, so if you see untranslated texts feel free to open a feature request in [Tauri's main repo].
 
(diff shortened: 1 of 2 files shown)


docs/guides/building/windows.md
```diff
@@ -250,9 +264,11 @@ Note that `ComponentGroup`, `Component`, `FeatureGroup`, `Feature` and `Merge` e
 
 ## Internationalization
 
-The Windows Installer is built using the `en-US` language by default. Internationalization (i18n) can be configured using the [`tauri.bundle.windows.wix.language`] property, defining the languages Tauri should build an installer against. You can find the language names to use in the Language-Culture column on [Microsoft's website][localizing the error and actiontext tables].
+The NSIS Installer is a multi-language installer, which means you always have a single installer which contains all the selected translations. You can specify which versions to include using the [`tauri.bundle.windows.nsis.languages`](../../api/config.md#nsisconfig.languages) property.
```
amrbashir: we should add a note here about missing languages and redirect users to open a PR

amrbashir (approved):

docs/guides/building/windows.md
FabianLars: True, wanted to keep that really short for now and just link to the config prop but i'll at least at that one. On that topic, i think we linked the wrong one in the config docs, we're [linking to the nsis repo](https://github.com/tauri-apps/tauri/blob/dev/core/tauri-utils/src/config.rs#L460) instead of our own (https://github.com/tauri-apps/tauri/tree/dev/tooling/bundler/src/bundle/windows/templates/nsis-languages). Or was that intentional and i'm misunderstanding the intention here?

amrbashir: yeah it is intentional, the link has a list of the languages that NSIS has built-in, so their app should usually be translated correctly, it will only have some untranslated text for our custom upgrading dialog.

amrbashir: so maybe state that these languages exist but if you see any untranslated text, please open an issue

FabianLars: Does this work? (preview still building...)

amrbashir: sounds good 👍

PR #1232, opened 2023-05-06
Add experimental cross-platform guide for building windows apps

FabianLars:


docs/guides/building/cross-platform.md
```diff
@@ -149,10 +151,104 @@ env:
...
+cargo install xwin
+```
+
+Then you can use the `xwin` CLI to install the needed files to a location of your choice. Remember the location, we will need it in the next step. In this guide we will create a `xwin` directory in the Home directory.
+
+```sh
+xwin --accept-license splat --disable-symlinks --output ~/xwin
+```
+
+If you want to read Microsoft's license upon installation, you can run the same command without the `--accept-license` flag.
```
amrbashir: Could we change this the other way around, eg if you want to accept all licenses at once, use this flag.

docs/guides/building/cross-platform.md
```diff
@@ -149,10 +151,104 @@ env:
...
+
+```toml title=.cargo/config.toml
+[target.x86_64-pc-windows-msvc]
+linker = "lld"
+rustflags = [
+  "-Lnative=/home/username/xwin/crt/lib/x86_64",
+  "-Lnative=/home/username/xwin/sdk/lib/um/x86_64",
+  "-Lnative=/home/username/xwin/sdk/lib/ucrt/x86_64"
+]
+```
```
amrbashir: this will break building on a native Windows machine, no? maybe change this to a build alias that has these flag specified on the command line?

amrbashir (changes requested):

docs/guides/building/cross-platform.md
FabianLars: Actually, let's completely remove it, it's just one simple yes/no prompt.

docs/guides/building/cross-platform.md
FabianLars: Well, it will only work on the system this file is defined on, even if you're moving from Linux to Linux (unless you use the same username lol).
But a more opt-in like approach would be good, i agree. I don't really know what you mean with build alias though, do you have a link at hand? I can only think of custom profiles right now which don't support custom rustflags.

amrbashir: > Well, it will only work on the system this file is defined on, even if you're moving from Linux to Linux (unless you use the same username lol).

yeah you're right, but I was more concerned about building for Windows on Windows.

We should instead mention that this file should be changed for every system and thus shouldn't be checked into git


cargo aliases: https://doc.rust-lang.org/cargo/reference/config.html#alias

FabianLars: Ah that alias thing. How would you combine that with the npm cli though?

Anyway, we can also just figure that out later, i tried to discourage this approach in the docs anyway so if users _do_ use it it's fine if it's not perfect imo.

I'll add a note about the git check-in thingy though, good call.

amrbashir: cool

amrbashir (approved):

PR #1233, opened 2023-05-07
Add Qwik getting started setup

vanvuongngo:
Hi there,

Thanks for your great project.
I would like to add the [Qwik](https://qwik.builder.io) framework.


FabianLars:
Thank you very much for contributing! Qwik is one of the more often requested frameworks :)

Can we maybe change the app structure to match the other guides? So that it looks more like this:
```
project-name/
  dist/
  src/
  src-tauri/
  package.json
```
instead of
```
random-folder/
  ui/
    dist/
    src/
    package.json
  src-tauri/
```

----

Also, no idea why the tabber isn't working, i'll take a look at that locally after the change above :)

vanvuongngo:
Hi @FabianLars, thanks a lot for your review and it's a great idea to streamline it. I will have a look.

vanvuongngo:
@FabianLars, bitte nochmals review :)

vanvuongngo:
The tabber is fixed. Good catch and I forgot the imports.

FabianLars (approved):
Abused this PR a bit to fix a few things in the other guides x)

Anyway, thank you!

PR #1252, opened 2023-05-17
Expand hyperlink around list item

i-c-b:
Expands hyperlink so that list item can be clicked. Previously, only text could be clicked.


simonhyll (approved):

lorenzolewis:
Thanks for being awesome and contributing this in so quickly @i-c-b !

PR #1260, opened 2023-05-23
Feat/versioned paths

lorenzolewis:
Currently has an issue where the locale isn't being picked up during production build. Most likely need to change to built-in Astro routing to handle the `[lang]` part of generated slugs.


FabianLars:
i know you're out of office rn but is there anything specific you want feedback on here?

Also, idk if i'm the only one but what do you think about going with `v1` instead of `1` again? 😂

lorenzolewis:
@FabianLars if you just want to check that all the routes between langs map as expected that'd be awesome <3

For the URL it seems like most sites just use the version number and don't include a `v` in the URL (looking at docs.rs specifically) so I'd prefer to keep a similar convention maybe?

I'm also going to merge this one in to get it deployed on the beta domain so feel free to give me a shout on Discord for anything.

FabianLars:
> For the URL it seems like most sites just use the version number and don't include a v in the URL (looking at docs.rs specifically) so I'd prefer to keep a similar convention maybe?

Yeah idk, a _single_ digit looks a bit weird to me, but let's stick with it for now.

PR #1265, opened 2023-05-27
Fix `cargo-xwin` installing command may not work for Linux

FlysoftBeta:
Linking issue: https://github.com/tauri-apps/tauri/issues/7004


lucasfernog:
@FabianLars should we actually add the error message here?

FabianLars:
yes I think we should

FlysoftBeta:
The error message is now added :)

FabianLars (commented):
Thank you!

FabianLars (approved):

PR #1275, opened 2023-05-31
Update math calc on eco impact table

sam-crabnebula:
Updated to correct rounded estimates on memory size based on app downloads for the ecological impact table


FabianLars (approved):
That went unnoticed for quite a long time, wow. Thank you!

PR #1279, opened 2023-06-08
[fix] Arch dependencies on prerequisites docs page 

CompeyDev:
Compiling Tauri apps fail on Arch, when following the current dependencies guide within the documentation. 

This PR fixes that by replacing the faulty `webkit2gtk` with the expected `webkit2gtk-4.1` within the pacman install command.

### DIFF
--- a/docs/guides/getting-started/prerequisites.md
+++ b/docs/guides/getting-started/prerequisites.md
@@ -119,7 +119,7 @@ sudo apt install libwebkit2gtk-4.0-dev \
 ```sh
 sudo pacman -Syu
 sudo pacman -S --needed \
-    webkit2gtk \
+    webkit2gtk-4.1 \
     base-devel \
     curl \
     wget \


CompeyDev:
**UPDATE**: Apparently, both `webkit2gtk` and `webkit2gtk-4.1` are required by `javascriptcore-rs-sys`.

Could someone please confirm that this is the case?

FabianLars:
`webkit2gtk-4.1` should only be required for tauri's next branch / v2 alphas. v1 still needs 4.0 which is still called just `webkit2gtk` according to their package search: https://archlinux.org/packages/extra/x86_64/webkit2gtk/

Sooo as far as i can see the docs are still correct since they document v1 (the `next` docs are already updated to 4.1)

> UPDATE: Apparently, both webkit2gtk and webkit2gtk-4.1 are required by javascriptcore-rs-sys.

If you use multiple versions of that crate it could be possible, yes. If there indeed multiple versions in the crate graph then there's likely something wrong with your dependencies.

PR #1283, opened 2023-06-12
Add completions guide to cli doc

FabianLars:
### wait for the 1.4 release to merge this!

Relevant preview will be here instead: <https://deploy-preview-1281--tauri.netlify.app/> (via https://github.com/tauri-apps/tauri-docs/pull/1281)

### DIFF
--- a/docs/.templates/cli.md
+++ b/docs/.templates/cli.md
@@ -84,7 +84,7 @@ Get the Bash completions and move to a known folder:
 <Tabs groupId="package-manager">
   <TabItem value="npm">
 
-```sh
+```shell
 npm run tauri completions -- --shell bash > tauri.sh
 mv tauri.sh /usr/local/etc/bash_completion.d/tauri.bash
 ```
@@ -92,7 +92,7 @@ mv tauri.sh /usr/local/etc/bash_completion.d/tauri.bash
   </TabItem>
   <TabItem value="Yarn">
 
-```sh
+```shell
 yarn tauri completions --shell bash > tauri.sh
 mv tauri.sh /usr/local/etc/bash_completion.d/tauri.bash
 ```
@@ -100,7 +100,7 @@ mv tauri.sh /usr/local/etc/bash_completion.d/tauri.bash
   </TabItem>
   <TabItem value="pnpm">
 
-```sh
+```shell
 pnpm tauri completions --shell bash > tauri.sh
 mv tauri.sh /usr/local/etc/bash_completion.d/tauri.bash
 ```
@@ -108,7 +108,7 @@ mv tauri.sh /usr/local/etc/bash_completion.d/tauri.bash
   </TabItem>
     <TabItem value="Cargo">
 
-```sh
+```shell
 cargo tauri completions --shell bash > tauri.sh
 mv tauri.sh /usr/local/etc/bash_completion.d/tauri.bash
 ```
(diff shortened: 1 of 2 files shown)


docs/.templates/cli.md
```diff
@@ -63,7 +65,147 @@ If you have entered a command to the `build.beforeBuildCommand` property, this o
...
-[Tauri Icon Guide](../guides/features/icons.md)
+For more information, check out the complete [Tauri Icon Guide](../guides/features/icons.md).
+
+## `completions`
+
+<Command name="completions" />
+
+{completions}
+
+The Tauri CLI can generate shell completions for Bash, Zsh, PowerShell and Fish. To write the completions script to a file, run tauri completions --shell [bash|zsh|powershell|fish] > outputFile.
```
lorenzolewis: You mention Fish but I'm not seeing a section below for it

lorenzolewis:
```suggestion
The Tauri CLI can generate shell completions for Bash, Zsh, PowerShell and Fish.
```

docs/.templates/cli.md
```diff
@@ -63,7 +65,147 @@ If you have entered a command to the `build.beforeBuildCommand` property, this o
...
+Here are some instructions to configure Bash, Zsh and PowerShell. If you face an issue, please follow your shell's instructions instead. Note that it is recommended to check the generated completions script before executing it for security reasons.
+
+### Bash
+
+Get the Bash completions and move to a known folder:
+
+<Tabs groupId="package-manager">
+  <TabItem value="npm">
+
+```sh
```
lorenzolewis: I think using `shell` instead of `sh` on these will make the leading `$` render, right?

docs/.templates/cli.md
FabianLars: probably because nobody knows how to do that with fish 😅 i think Lucas added that because clippy_complete says it supports it.

Gonna see if i find something

docs/.templates/cli.md
FabianLars: they even were `shell` when i copied the tab but i thought the word `shell` is kinda ugly so i replaced it. looks like you really can't trust my judgement anymore after a 3-4 day break 😂 

fixed it

docs/.templates/cli.md
lucasfernog: I didn't even test completions on Fish tbh (but it looks alright)... I still think a Fish is that thing that lives in the ocean

FabianLars: Looking at this https://fishshell.com/docs/current/completions.html#where-to-put-completions it sounds like we should go for `~/.local/share/fish/vendor_completions.d/`. Do we still want to use `~/.config/fish/completions/` to be safe (also note the part further down `If you are unsure, your completions probably belong in ~/.config/fish/completions.`)

FabianLars: so far i only make fish show the completions once (using tauri.fish), but i think that was a bug somehow 😅The issue is that the file name must be the same as the command to make fish pick it up so i'm a bit lost right now.

Looks like we'll have to name the file cargo.fish pnpm.fish etc which is super annoying. I guess i'll change the command to append to the file instead of moving/overwriting but that would break completion files from one of the other sources from the link above because of this [issue](https://togithub.com/fish-shell/fish-shell/issues/6814); what a mess.

Unless someone has an idea rn, I tend to just leave that unresolved for now and open a "help wanted" issue to see if someone with more fish experience can help out because i can't find anything helpful on the internet here.

lucasfernog: In this case I would leave the issue open and wait for a Fish user to try use the completions (is there any???)

FabianLars: (~~Yeah, the one who requested fish support in clap_complete~~😂)

FabianLars: https://github.com/tauri-apps/tauri-docs/issues/1288

PR #1290, opened 2023-06-12
Update Governance Page

lorenzolewis:
This updates the governance page inline with the new model.

For reference a version of the SVG is in Figma here: https://www.figma.com/file/CKTllB7R52D080yDHjno92/Tauri-Brand-Guidelines?type=design&node-id=2213%3A3263&t=NhCNeUhHF4X7uUWI-1


src/pages/about/governance.md
```diff
@@ -1,44 +1,51 @@
----
-id: governance
----
-
 # Tauri Governance
 
-## Sustainability
+One of the main goals of the organizational structure of Tauri is to guarantee the sustainability, health and well-being of its contributors. [The Tauri Programme within the Commons Conservancy](https://dracc.commonsconservancy.org/0035/) was established to facilitate an open, transparent and efficient governance process throughout the future development of the Tauri and its auxiliary materials.
```
Beanow:
```suggestion
One of the main goals of the organizational structure of Tauri is to guarantee we stay true to our open source values and do so sustainably while respecting the health and well-being of contributors. [The Tauri Programme within the Commons Conservancy](https://dracc.commonsconservancy.org/0035/) was established to commit to those values, and facilitate an open, transparent and efficient governance process throughout the future development of the Tauri and its auxiliary materials.
```

The main part I feel is inaccurate is that it currently reads like the programme was established for it's governance utility. To me governance is the *how*. The *why* I think is because we wanted a permanent commitment to core values (open source, sustainability, for the greater good, equal opportunity, ...).

Still I think it's worth making special mention of sustainability and well-being as many "design choices" in the governance model are specifically for this. Such as why we have appointing seasons, or prefer 2-3 Domain Leads over 1.

src/pages/about/governance.md
```diff
@@ -1,44 +1,51 @@
...
 
-One of the main goals of the organizational structure of Tauri is to guarantee the sustainability of Tauri and the health and well-being of its contributors. The world of Open Source is fraught with peril and discord, and we have taken measures to ensure the longevity of Tauri. This document explains how we go about doing so.
+## Tauri Working Group
 
-## Organizational Structure
+The Tauri Working Group is the collective framework created to enable this governance process. It's composed of the following components:
```
Beanow, line 7: Did we ever find a great description for the working group? I think the most *truthful* is that it's where regular contributors more deeply involved with Tauri collaborate.

src/pages/about/governance.md
```diff
@@ -1,44 +1,51 @@
...
+### Working Group Members
 
-Visit the [Governance and Guideance repository](https://github.com/tauri-apps/governance-and-guidance) to get more information.
+All of the individuals that make up the Tauri Working Group.
 
-## Code of Conduct
+### Tauri Board & Board Directors
 
-Everyone participating in the Tauri community is expected to follow the [Code of Conduct](https://github.com/tauri-apps/governance-and-guidance/blob/main/CODE_OF_CONDUCT.md).
+The collective Board is the central decision making body for the Tauri Programme. Although the Tauri Board is ultimately responsible for decisions, they may delegate to Domain Leads as they see fit.
```
Beanow: Who they might delegate to isn't restricted to Domain Leads. For example, Domain Leads may in turn delegate to Teams. So for this document I would describe it in terms of what they *typically* are concerned with.
- Major decisions (potentially impacting the future of Tauri)
- Concerned with overall project goals

src/pages/about/governance.md
```diff
@@ -1,44 +1,51 @@
...
-## Licensing
+Teams are small groups of contributors that support or maintain specific areas of the Tauri Programme. They are a means for Tauri to execute its longer-term tasks and goals, especially when ad-hoc contributions cannot achieve the same results.
 
-We, the contributors to Tauri Apps, use the MIT and Apache licenses for all code content. Images and bodies of text, unless otherwise noted are CC-BY-ND-NC.
 
-## Trademark
+## Get Involved
 
-It is a permissible use of the name "Tauri App" or the Tauri logo to show that a project uses Tauri. "Tauri Studio" is reserved for use by the organization.
+Anyone is welcome to request joining the Tauri Working Group. You can reach out on the [Tauri Discord](https://discord.com/invite/tauri) or send an email to [board@tauri.app](mailto:board@tauri.app).
```
Beanow: Oh please no :joy:

I feel like:
1. WG access shouldn't be decided *just* by the Board.
2. The WG itself is really fuzzy, so the CTA should concentrate on Teams first.
3. Current WG members are invited / endorsed by someone else, it represents a trust-level that I think we shouldn't change here.

Beanow (changes requested):

src/pages/about/governance.md
lorenzolewis: @Beanow I tweaked the grammer on that suggestion.

I think I'd use any of the "why's" in a blog post about these changes instead of here (and reference that blog post on this page). To Perry's point we should think of what's the point of this page, it's to serve as a quick tl;dr reference, the blog post can be a deep-dive into the _why_.

src/pages/about/governance.md
lorenzolewis: Eh, this is one that I think is slightly evolving over time and will probably continue to evolve before it settles a bit more. So I'm not sure if we want to spend too much time on it in this current moment.

src/pages/about/governance.md
lorenzolewis: Good point on this, do you have some specific phrasing in mind to use?

src/pages/about/governance.md
lorenzolewis: Phrasing might have been a bit off (it was end of the week for me 😅) but I meant it to be:

1) If you don't use Discord you can email board as an alternative means
2) You mean to illustrate the interest-specific places for someone to get involved in? I'd start off with domains since those are more mature in practice than teams atm
3) Hmm, I see your point. But I'm also wondering how we have someone express interest in joining without them having "buddied up" with someone if that makes sense

Lmk what you think about those and can work to tweak it a bit.

src/pages/about/governance.md
Beanow: Yep I got stuck on this for like an hour the other day. Somewhat leaning towards, *it's where we coordinate for long term goals.* But indeed. I'd rather have a meh definition right now, and concentrate on getting straight inaccurate things corrected.

(Code Teams are superseded, and there's no WG Media, sort of thing :joy:)

src/pages/about/governance.md
Beanow: Yeah I understand having the board address as fallback. What I mean is historically WG membership has been invite based. You were invited and people didn't apply or request access.

I'm **all for changing that**. But this isn't the PR to make that change imo. Because of that trust level it's a bigger change. So I would scrap this CTA entirely *for now*. Until we've looked closer at what the new approach should be.

lorenzolewis: Ah gotcha, let me revise that

lorenzolewis: Just removed that paragraph/sentence

Beanow (approved):
Nice TL;DR for the current structure. Expecting we may rephrase parts, but I prefer that over leaving the outdated information :smile:

PR #1292, opened 2023-06-13
Create CODEOWNERS file

FabianLars:


FabianLars:
ig we'll have to merge it to see if it works

FabianLars:
actually lemme comment out devops as long as we don't actually have that team

PR #1301, opened 2023-06-19
feat(config): document for enum variant properties, closes #1300

### LINKED ISSUE #1300: [docs] Missing `require_literal_leading_dot` documentation
The newly (`1.4.x`) introduced configuration switch `require_literal_leading_dot` which configures part of the `fs` scope is not properly documented yet.

Looping in @lucasfernog as discussed.

lucasfernog:


docs/api/config.md
```diff
@@ -495,7 +558,17 @@ Each pattern can start with a variable that resolves to a system base directory.
 Can be any of the following types:
 
 - `string[]`: A list of paths that are allowed by this scope.
-- `object`: A complete scope configuration.
+- `{ "allow": [string], "deny": [string], "requireLiteralLeadingDot": boolean? }`: A complete scope configuration.
+
+	
+	| Name | Type | Default | Description |
+	| ---- | ---- | ------- | ----------- |
+	| <div className="anchor-with-padding" id="fsallowlistscope.allow">`allow`<a class="hash-link" href="#fsallowlistscope.allow"></a></div> | [string] | [] | A list of paths that are allowed by this scope. |
```
amrbashir: the type format is incorrect, it should be

```suggestion
	| <div className="anchor-with-padding" id="fsallowlistscope.allow">`allow`<a class="hash-link" href="#fsallowlistscope.allow"></a></div> | string[] | [] | A list of paths that are allowed by this scope. |
```

docs/api/config.md
```diff
@@ -495,7 +558,17 @@ Each pattern can start with a variable that resolves to a system base directory.
 Can be any of the following types:
 
 - `string[]`: A list of paths that are allowed by this scope.
-- `object`: A complete scope configuration.
+- `{ "allow": [string], "deny": [string], "requireLiteralLeadingDot": boolean? }`: A complete scope configuration.
```
amrbashir: do we want to embed definition here instead of just `object`? because it is epxlained belowe

amrbashir (changes requested):
there is other two issues I noticed that are not related to this PR but I will mention them in case you want to fix'em:
1. there is a lot of keys that has `array?` or `object?` as types and not actual type
image

2. Also if you look at the above image, you will notice that the question mark indicating optional value is not inside the backticks, it should 
```
`string?`
```
not 
```
`string`?
```

docs/api/config.md
lucasfernog: There's some enum variants with multiple objects so it feels weird to have several `- object:` lines IMO. But longer definitions are indeed weird.

docs/api/config.md
lucasfernog: I've fixed it via the generator script.

docs/api/config.md
lucasfernog: What do you think @amrbashir @lorenzolewis

amrbashir: alright, lets keep it like this for now

amrbashir (approved):

FabianLars (approved):
Thanks for fixing it :)

PR #1302, opened 2023-06-21
Typo on build argument

goenning:
Just tried running it locally and `bundle` is not a valid


FabianLars (approved):
That's what happens if you always use the `-b` shortcut 😂 Thanks!

PR #1307, opened 2023-06-21
fix(config): prepare generate script for v2 config schema

lucasfernog:


FabianLars:
For reference (had to search for it first too 🙃 ) this is needed for example `Color` which is now
```
Type: `[{"type":"integer","format":"uint8","minimum":0},{"type":"integer","format":"uint8","minimum":0},{"type":"integer","format":"uint8","minimum":0},{"type":"integer","format":"uint8","minimum":0}][]`
```
instead of
```
Type: ``integer` _(format: `uint8`)_[]`
```

FabianLars (approved):
thanks

PR #1324, opened 2023-07-13
Rewrite Windows prereq guide

FabianLars:
@lorenzolewis Please destroy my self-esteem by checking my wording here.

I rewrote this so many times that i completely lost track of what i wanted it to look like and now i dislike everything 🙃


lorenzolewis (changes requested):
I would still keep the formatting of the larger "1: Build tools, 2: WebView2, 3: Rust" steps like we had before (and still have for macOS and Linux). I think that makes it clear that there are 3 distinct dependencies to look through at least. But I'd add this good information you have around the MS C++ build tools. 

So overall love it, I'd just tweak the style to match the other OSes like the live version is.

FabianLars:
Do you mean for both auto and manual installation? Because that's a bit tricky since in the automatic one the build tools and rust installations are basically one step. I thought about splitting them up ala "Once you hit the main rustup menu go to step X" but thought that would interrupt the flow and would be hard to follow.

At first the manual install section was basically like the old one but because the automatic one also needed the webview one i felt it'd be a bit nicer to have them before or after vs&rust (so the sections match each other) at which point i thought it'd be better to just move that into a general step especially since most win10+ systems should have it installed now.

FabianLars:
idk, i really feel like this is too much text, especially compared to linux & macos. I think sticking more to the old layout and just somehow mention it would still be better. Either way i think i'll leave this as-is and instead do it the right way in the new docs.

lorenzolewis:
@FabianLars do you want to keep this branch around for reference or should we delete it and yeet it out the window?

FabianLars:
deleted it. if we do end up needing it after all, github let's us restore it anyway, and i have the images (plus one) saved locally too.

PR #1328, opened 2023-07-17
Update quickstart: SvelteKit

atilafassina:
Hello!
First-time contributor here, so apologies if I put something out of place.

I was just following the SvelteKit QuickStart and noticed that some of the code snippets were out of date and that the suggested Port was not the one being used by the app.

So here is my attempt of providing a fix.

## This PR:
- Replaces `localhost:5173` with `localhost:5175` as its the default for SvelteKit
- Replaces `svelte-preprocess` with `@sveltejs/kit/vite`
~- switches the `invoke` import from `@tauri-apps/api/tauri` to `@tauri-apps/api`~  **reverted**

## Testing
Following the step should end up with the exact same app as the images.

### DIFF
--- a/docs/guides/getting-started/setup/sveltekit.mdx
+++ b/docs/guides/getting-started/setup/sveltekit.mdx
@@ -141,7 +141,7 @@ Furthermore, if you prefer a Single-Page Application (SPA) mode over SSG, you ca
     __html: 'Use <code>../build</code> for this value.',
   }}
   devPathExplination={{
-    __html: 'Use <code>http://localhost:5175</code> for this value.',
+    __html: 'Use <code>http://localhost:5173</code> for this value.',
   }}
   beforeDevCommandExplination={{
     __html:
@@ -162,7 +162,7 @@ Now that we have scaffolded our frontend and initialized the Rust project the cr
     "beforeBuildCommand": "npm run build",
     // This command will execute when you run `tauri dev`
     "beforeDevCommand": "npm run dev",
-    "devPath": "http://localhost:5175",
+    "devPath": "http://localhost:5173",
     "distDir": "../build"
   },
 ```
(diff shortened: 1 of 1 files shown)


FabianLars:
Thanks for contributing! ❤️ 

~~Apart from complaining about sveltekit which i like to do,~~ could you also revert this change please?
> switches the invoke import from @tauri-apps/api/tauri to @tauri-apps/api

I'm actually surprised you didn't see any errors but importing from the package root often causes issues in ssr-first frameworks. Either way importing it from the module should be a tiny bit more future proof too (in case we remove the re-export)

FabianLars:
Also, i just created a new sveltekit app and it still uses vite's default port 5173 here

atilafassina:
> Apart from complaining about sveltekit which i like to do, could you also revert this change please?

🤣


> Also, i just created a new sveltekit app and it still uses vite's default port 5173 here

Weird.
I actually create 3 different apps to test. port 5173 doesn't error, but the app window is blank for me.
I used `pnpm create svelte` and `pnpm create svelte@latest` to generate that.
Should I revert this one as well?


> I'm actually surprised you didn't see any errors but importing from the package root often causes issues in ssr-first frameworks. Either way importing it from the module should be a tiny bit more future proof too (in case we remove the re-export)

Not gonna lie. I was a bit confused by this one too. I just followed VSCode's autocomplete and my console was clean, so I assumed it was a good to go. I can dig deeper into it later, but for now I'll just revert 😊

FabianLars:
> I actually create 3 different apps to test. port 5173 doesn't error, but the app window is blank for me.
> I used pnpm create svelte and pnpm create svelte@latest to generate that.

Maybe there's a zombie nodejs process still binding the port? i also used `pnpm create svelte@latest`

> Should I revert this one as well?

Yeah, i guess so. For what it's worth, their docs also mention 5173 (not that i really trust their docs) and i sincerely hope they wouldn't change the default port in a non-major release 😅 

> I just followed VSCode's autocomplete

vscode is lazy and uses the first thing it finds :D

FabianLars (approved):

PR #1330, opened 2023-07-21
feat(core): allow window label pattern for remote domain access scope

james138138:
The code change is https://github.com/tauri-apps/tauri/pull/7468


FabianLars:
The config doc page is auto generated, so once your tauri PR lands it will be added to the docs automatically. Thanks though!

PR #1333, opened 2023-07-27
Add a guide for cross-compilation for ARM devices

Benji377:
Dear Tauri community,

Some time ago, I encountered challenges while trying to cross-compile my Tauri app for the Raspberry Pi. Fortunately, with the invaluable help of @FabianLars on Discord, I received a small guide on how to achieve this. After further research, I stumbled upon a GitHub Action that streamlines the cross-compilation process.

Seeing that many others were also facing similar difficulties and seeking guidance on cross-compilation in the same Discord support thread, I decided to create a comprehensive guide on how to cross-compile Tauri applications for ARM-based devices, with both manual and automated approaches.

The guide covers prerequisites, setting up your environment, and the step-by-step cross-compilation process. Additionally, it includes a GitHub Action for automatic cross-compilation. I have tested the guide on Debian 11 WSL and validated the resulting `.deb` package on a Raspberry Pi 3B+.

I believe this guide would be a valuable addition to the Tauri documentation, as it addresses a common pain point for developers targeting ARM devices. As someone who mainly translates documentation, this is my first attempt at writing Tauri documentation, and I would be grateful for any feedback or suggestions to improve it.


docs/guides/building/arm-cross-compile.md
```diff
@@ -0,0 +1,162 @@
...
+### Prerequisites
+
+:::warning Important Note
+
+Ubuntu 20.04 WSL has issues with webkit dependencies. It's recommended to use Debian instead.
+
+:::
+
+- For Windows, install Debian on WSL using the [wsl setup guide].
+- On Linux, ensure your machine has the same GLIBC version as the target device. Check the version using: `ldd --version`.
```
FabianLars:
```suggestion
- On Linux, ensure your build machine has an older GLIBC version than the target device. Check the version using: `ldd --version`.
```

It must be the same version or an older version, having troubles making it sound like correct english so i just wrote "older". If you can come up with something better feel free to change it.

docs/guides/building/arm-cross-compile.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Manual Compilation
+
+Manual compilation is suitable when you don't need to compile your application frequently and prefer a one-time setup. Follow the steps below:
+
+### Prerequisites
+
+:::warning Important Note
+
+Ubuntu 20.04 WSL has issues with webkit dependencies. It's recommended to use Debian instead.
```
FabianLars: Just FYI, before merging i'll test 18.04 and 20.04 on my system because we tend to default to ubuntu everywhere else.

docs/guides/building/arm-cross-compile.md
```diff
@@ -0,0 +1,162 @@
...
+
+This guide explains how to cross-compile your Tauri application for ARM-based devices, such as the Raspberry Pi. Compiling directly on the device may not be feasible due to limited RAM, so we'll explore two methods: manual compilation using Linux or WSL on Windows and automatic cross-compilation using a GitHub Action.
+
+## Manual Compilation
+
+Manual compilation is suitable when you don't need to compile your application frequently and prefer a one-time setup. Follow the steps below:
+
+### Prerequisites
+
+:::warning Important Note
```
FabianLars:
```suggestion
:::warning
```

docs/guides/building/arm-cross-compile.md
```diff
@@ -0,0 +1,162 @@
...
+4. Install [Tauri using Cargo]: `cargo install tauri-cli`.
+5. Install necessary tools for your frontend framework, like npm and Next.js.
+
+:::info OpenSSL Workaround for WSL on Windows
+
+If using WSL on Windows and encountering OpenSSL-related errors, add the following line to `Cargo.toml` in `<project-root>/src-tauri/Cargo.toml` within the `[dependencies]` section:
+```toml
+openssl-sys = {version = "0.9.66", features = ["vendored"]}
+```
+:::
```
FabianLars: to be honest i feel like this whole section should be removed

docs/guides/building/arm-cross-compile.md
```diff
@@ -0,0 +1,162 @@
...
+
+### Cross-Compiling
```
FabianLars: We need to explain arm 7 vs 8/64 in this section imo (and add instructions for the latter). Both are equivally often requested, arm 7 for (older) Pis and arm64 for more desktop-like targets.

docs/guides/building/arm-cross-compile.md
```diff
@@ -0,0 +1,162 @@
...
+
+5. Update the package info: `sudo apt-get update && sudo apt-get upgrade -y`.
+6. Install webkitgtk for ARM: `sudo apt install libwebkit2gtk-4.0-dev:armhf`.
+7. Set the PKG_CONFIG_SYSROOT_DIR for the appropriate arch: `export PKG_CONFIG_SYSROOT_DIR=/usr/arm-linux-gnueabihf/`.
+8. Build the app by specifying the desired ARM version: `cargo tauri build --target armv7-unknown-linux-gnueabihf`.
+
+### Final Notes
+
+- The static website can be found in the `out` folder.
+- Rust may try to run your app after compiling it, which will fail on non-ARM platforms. Ignore this error, as the .deb package will be correctly built.
```
FabianLars: This section can be removed imo. The `out` folder is completely relevant and for the second point see my comment on the other Final Notes section

docs/guides/building/arm-cross-compile.md
```diff
@@ -0,0 +1,162 @@
...
+
+## Automatic Cross-Compilation using a GitHub Action
+
+For automated ARM executable builds on GitHub, we'll use the [arm-runner-action] created by [Paul Guyot].
+
+### Setup
```
FabianLars: I can't look into this section too deeply today but 2 things users might end up having problems with are that it doesn't use tauri-action, understandibly so since it's incompatible with the arm action, and secondly that it uploads it as an artifact and not to the release what we do in the other guides.

Because of that i'd almost prefer mirroring the local approach 🤔 not sure though.
Also maybeee we should pull this, and the actions from the cross-platform and macos guides into a new GitHub Action build page but we can do that in a separate PR or just in the new docs.

docs/guides/building/arm-cross-compile.md
```diff
@@ -0,0 +1,162 @@
...
+To avoid this issue and ensure a smooth cross-compilation process, you can take the following steps:
+
+1. Configure `tauri.conf.json`: Inside the `src-tauri` folder, you can customize the `tauri.conf.json` file to specify the target platforms. Remove `"appimage"` from the `"targets"` array and keep only the desired targets for your ARM-based device. For example:
+   ```json
+   "targets": ["deb", "nsis", "msi", "app", "dmg", "updater"],
+   ```
+
+By adjusting the `tauri.conf.json` file, you instruct Tauri to exclude the "appimage" target, which is irrelevant for ARM devices. This way, Rust won't attempt to execute the non-ARM-compatible "appimage" target during the cross-compilation process, preventing any failures and ensuring the successful creation of the `.deb` package for your ARM-based device.
+
+With these modifications, you can confidently use the GitHub Action to automatically build an ARM-compatible `.deb` package for your Tauri application. Happy coding!
```
FabianLars: I will copy my comment from discord:

> Also, the Final Notes section: The fix itself is correct, but the problem you describe isn't. It's not trying to run the app at the end or something. This is solely the appimage bundler not supporting arm targets resulting in the download of the necessary bundler tools failing
and nothing of it is related to rust, that part is already done at that point
And lastly on that topic, i think it makes more sense to completely remove that section and add a warning ala "appimages currently unsupported, disable it via" at the top similar to the ubuntu2004 warning

FabianLars (changes requested):
Thank you!! I do have a few requests as you see but nothing too bad i hope 😅 

Also, as mentioned on Discord i feel like it would be better if we can add it to the "Linux Bundle" page but i'm open for discussion.

docs/guides/building/arm-cross-compile.md
Benji377: I remember trying with Ubuntu 20.04, and I had problems with it. Here: https://discord.com/channels/616186924390023171/1056210277894463571/1057382381268508702

docs/guides/building/arm-cross-compile.md
Benji377: It would be great if you could use the tauri action instead. It would be great, especially because you would have:
1. Everything in one action
2. Automatic Release draft creation, which is a feature I really enjoy.

I need to admit that I was not able to set up this action alone, the creator of the Action, Paul, helped me out a lot and fixed most of the issues I encountered. You are right, this might fit better somewhere else. Also, something I forgot to mention is that the Action takes about 3 hours to create an executable on Raspirus. My project is not too big, but also not very efficient, so maybe it's a bit less for others. Still, it's a very long time and I usually let it run before going to sleep.

docs/guides/building/arm-cross-compile.md
Benji377: Done

docs/guides/building/arm-cross-compile.md
Benji377: Removed as requested

docs/guides/building/arm-cross-compile.md
Benji377: I unfortunately couldn't find a better word to describe it. I came up with:
On Linux, the build machine requires GLIBC version equal to or older than the target device.

Which is basically what you suggested

docs/guides/building/arm-cross-compile.md
Benji377: I removed the entire section as requested and added a warning at the top.

docs/guides/building/arm-cross-compile.md
Benji377: I added the same warning twice. Once in the Prerequisites section and once at the start of the Action

docs/guides/building/arm-cross-compile.md
Benji377: Should I add a second command showing the difference to each step, or a separate step-by-step guide entirely? Should I also discuss the background to both architectures and what is compatible with which?
I also don't think I can properly test it, as I only have Raspberry Pi 3 and 4.

docs/guides/building/arm-cross-compile.md
FabianLars: Yeah, i remembered that thread too, but iirc it worked for me on ubuntu 1804 a while back and in the context of that thread i meant to try out 20.04 myself too but never got to it :/

docs/guides/building/arm-cross-compile.md
FabianLars: works for me for now 👍

docs/guides/building/arm-cross-compile.md
FabianLars: > Should I add a second command showing the difference to each step, or a separate step-by-step guide entirely?

a seperate guide sounds too much, i'd either just mention the differences, or if that's not clear enough in your opinion, make use of the tabs component.

> I also discuss the background to both architectures and what is compatible with which?

You _can_ but it's not a requirement in my eyes. (keep it short though if possible :D )

> I also don't think I can properly test it, as I only have Raspberry Pi 3 and 4.

no worries, i didn't hear any complains about my arm64 instructions yet, and either way we can just fix the docs later if it ends up being wrong in some way.

Benji377: I added the separate commands for armv8, let me know if it's ok

docs/guides/building/arm-cross-compile.md
Benji377: Maybe I should remove the action entirely. It's a bit weird, as you said

docs/guides/building/arm-cross-compile.md
FabianLars: cool, looks good!

FabianLars (approved):
sry for the delay i was out of office. I think this is good enough as-is and i don't wanna keep v1 PRs open for too long since we're shifting our development to the new docs website.

Thank you!

PR #1335, opened 2023-07-27
Update Troubleshooting part of prerequisites.md

Phloemus:
changes: in Troubleshooting, precising that the terminal should be closed an an other one should be used in order to check if Rust is installed correctly with the rustc --version command


FabianLars:
Thanks for contributing! Question, isn't it enough to mention that in all the install instructions already? :thinking:

Phloemus:
I thought, it was important to start a new terminal to run the command. I tested again and it turns out to work by simply launching an other shell.. so yes, nothing to change. Sorry 😅

PR #1337, opened 2023-07-28
Update or fix Debian dependencies

JonathanLindsey:
I was getting errors while building tauri's helloworld example on Ubuntu 22.10. These changes allowed me to build.

Please verify these are the *right* changes. I have not tested on vanilla Debian if that matters.


FabianLars:
> Please verify these are the right changes

Well, they are correct _and_ incorrect :D - These are the dependencies you need for Tauri v2, for v1 the current docs are still correct. Since we already list the new deps in the [v2 docs](https://next--tauri.netlify.app/next/guides/getting-started/prerequisites/linux) i'll close this PR. Thank you nonetheless.

PR #1339, opened 2023-07-31
Fixed link to tauri-bundler

Benji377:
The url to the tauri-bundler was broken, as it pointed to a non-existing v1.4 release. As @lorenzolewis suggested, the url has been updated to point to the v1.3 release instead


lorenzolewis (approved):
Thank you so much @Benji377 <3

PR #1341, opened 2023-08-01
Translation:"guide" part for zh-cn

Embers-of-the-Fire:
Translate and fix expressions in `docs/guides` for Chinese.


lorenzolewis:
Hey @Embers-of-the-Fire thank you so so much for this! 

Like we mentioned on Discord, we've currently frozen any translations that are happening while we're re-writing our docs system and preparing for Tauri 2.0. You can see the branch here that will be the future setup: 

https://github.com/tauri-apps/tauri-docs/tree/starlight

I'm going to go ahead and close this for now but keep an eye out on Discord where we'll announce a call for translators when we're ready

PR #1354, opened 2023-08-03
Integrate Tauri & Integration Guide

lorenzolewis:
Notes to address before merging:

- [ ] Remove `rm -rf ./dist/node_modules` from `astro:build` once https://github.com/withastro/astro/issues/7933 is fixed
- [ ] Tauri Config: `tauri.identifier` change the default behaviour so that either a) it allows `com.tauri.dev` or b) derives something like `com.example.project-name`
- [ ] `tauri:init`: Update `beforeBuildCommand`, `beforeDevCommand`, `devPath`, and `distDir` and include sensible defaults (such as detecting package manager), updated copy (with an example) and ghost text so you can type to replace what the prompt has or `<enter>` to accept what's there
- [ ] Default `tauri.conf.toml` file: Clean it up to only include items that are configured to make it more approachable
- [ ] Ask if mobile would like to be initialised as part of `tauri init` flow


lorenzolewis:
@simonhyll this might be relevant to you if you're still playing around with combining Tauri and Astro together: https://github.com/withastro/astro/issues/7933 Should be published when https://github.com/withastro/astro/pull/8096 is merged.

For some reasons a `node_modules` folder slips into the `dist` folder and then Tauri will error out during build, thinking the user pointed to the wrong folder. This should resolve that and clean up the `node_modules` folder during build.

lorenzolewis:
Going to de-prioritize this one until we get more guides written (and closer to beta)

Issue #1363, opened 2023-08-05
Create Tauri 2.0 CLI Reference


lorenzolewis:
In the 1.0 documentation we use this to document the Tauri CLI: https://github.com/tauri-apps/tauri-docs/blob/dev/.scripts/generate-cli-doc.js

This relies on having upstream compiled which is a bit annoying from a CI/CD perspective. Is there a way we can write this information in a JSON format upstream so that we can just pull it directly and format it here? Or this might be something updated so infrequently that we opt for manually writing it instead.

Edit: I'd actually prefer starting simple and just making a `.md` file specifically for this since it doesn't change much at all over time.

FabianLars:
I copied the v1 approach into [here](https://github.com/tauri-apps/tauri-docs/blob/docs/cli/src/content/docs/2/reference/cli.mdx) so we can have something to look for this question.

I'm a bit unsure how we want to handle sub commands. I don't think the approach used in v1 is the best fit but i can't think of anything much better. For example `tauri android` looks fine as is imo, so one section for all sub commands but the plugins commands i think are better suited to have a section each beacuse a) they are all pretty different to a point i feel like it's weird that they are in one, like `plugin add` vs all the rest basically (the "rest" is about modifying the current plugin project while plugin add is about adding a plugin to the current non-plugin project) and b) because some of these subcommands have subcommands

Also i'm a bit unsure about the wanted level of verbosity, do we want to keep it fairly simple as it is now, though i'd like to convert it into tables like here https://vitejs.dev/guide/cli.html, or do we want it a bit more text-like and split-up like how astro does it https://docs.astro.build/en/reference/cli-reference/

----

I think i shouldn't have started with this and let you handle it 😂

lorenzolewis:
Going to pick up this convo over on the PR @FabianLars

lorenzolewis:
Closed by #1365 , not worried about automation right at this point with how infrequently this changes.

PR #1365, opened 2023-08-07
Docs/cli

### LINKED ISSUE #1366: Create a "Commands" Astro component
Create a component similar to this one we had in Docusaurus but for Astro: https://github.com/tauri-apps/tauri-docs/blob/dev/src/theme/Command.js

Let's keep this simple and allow passing parameters for specifically-labeled props. Maybe an interface like this:

```ts
interface Props {
  npm?: string;
  yarn?: string;
  pnpm?: string;
  cargo?: string;
}
```

We would pass the full command to the component and it would conditionally render a Starlight `Tabs` component: https://starlight.astro.build/guides/components/#tabs

lorenzolewis:
Just getting this into a draft PR for those fancy netlify previews

#1363

### DIFF
--- a/src/content/docs/2/reference/cli.mdx
+++ b/src/content/docs/2/reference/cli.mdx
@@ -33,8 +33,10 @@ pnpm add -D @tauri-apps/cli@next
   	<TabItem label="Cargo">
 
 ```sh
-cargo install tauri-cli@next
+cargo install tauri-cli --version "^2.0.0-alpha"
 ```
+
+{/* TODO: 2.0 */}
 
   </TabItem>
 </Tabs>
--- a/src/content/docs/2/reference/cli.mdx
+++ b/src/content/docs/2/reference/cli.mdx
@@ -42,9 +42,12 @@ For CLI commands related to developing plugins visit the [Develop a Tauri Plugin
 
 ## `build`
 
-```sh
-cargo tauri build
-```
+<CommandTabs
+	npm="npm tauri build"
+	yarn="yarn tauri build"
+	pnpm="pnpm tauri build"
+	cargo="cargo tauri build"
+/>
 
 ```
 Tauri build
@@ -84,9 +87,12 @@ Options:
 
 ## `dev`
 
-```sh
-cargo tauri dev
-```
+<CommandTabs
+	npm="npm tauri dev"
+	yarn="yarn tauri dev"
+	pnpm="pnpm tauri dev"
+	cargo="cargo tauri dev"
+/>
 
 ```
 Tauri dev
@@ -118,9 +124,12 @@ If you have entered a command to the `build.beforeDevCommand` property, this one
 
 ## `icon`
 
-```sh
-cargo tauri icon
-```
+<CommandTabs
+	npm="npm tauri icon"
+	yarn="yarn tauri icon"
+	pnpm="pnpm tauri icon"
+	cargo="cargo tauri icon"
+/>
 
 ```
 Generates various icons for all major platforms
@@ -141,9 +150,12 @@ Options:
 
 ## `info`
 
-```sh
-cargo tauri info
-```
+<CommandTabs
+	npm="npm tauri info"
+	yarn="yarn tauri info"
+	pnpm="pnpm tauri info"
+	cargo="cargo tauri info"
+/>
 
 ```
 Shows information about Tauri dependencies and project configuration
@@ -161,9 +173,12 @@ Options:
 
 ## `init`
 
-```sh
-cargo tauri init
-```
+<CommandTabs
+	npm="npm tauri init"
+	yarn="yarn tauri init"
+	pnpm="pnpm tauri init"
+	cargo="cargo tauri init"
+/>
 
 ```
 Initializes a Tauri project
@@ -203,9 +218,12 @@ Options:
 
 ## `plugin add`
 
-```sh
-cargo tauri plugin add
-```
+<CommandTabs
+	npm="npm tauri plugin add"
+	yarn="yarn tauri plugin add"
+	pnpm="pnpm tauri plugin add"
+	cargo="cargo tauri plugin add"
+/>
 
 ```
 Installs a plugin on the project
@@ -226,9 +244,12 @@ Options:
 
 ## `signer`
 
-```sh
-cargo tauri signer
-```
+<CommandTabs
+	npm="npm tauri signer"
+	yarn="yarn tauri signer"
+	pnpm="pnpm tauri signer"
+	cargo="cargo tauri signer"
+/>
 
 ```
 Tauri updater signer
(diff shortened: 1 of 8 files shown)


lorenzolewis:
Continuing the conversation from #1363 ...

> I copied the v1 approach into [here](https://github.com/tauri-apps/tauri-docs/blob/docs/cli/src/content/docs/2/reference/cli.mdx) so we can have something to look for this question.

Awesome, thank you! <3

> I'm a bit unsure how we want to handle sub commands. I don't think the approach used in v1 is the best fit but i can't think of anything much better. For example tauri android looks fine as is imo, so one section for all sub commands but the plugins commands i think are better suited to have a section each beacuse a) they are all pretty different to a point i feel like it's weird that they are in one, like plugin add vs all the rest basically (the "rest" is about modifying the current plugin project while plugin add is about adding a plugin to the current non-plugin project) and b) because some of these subcommands have subcommands

Maybe this is an upstream question... but why do we have `tauri plugin ios` vs `tauri ios` plus `tauri plugin android` vs `tauri android`? I can see this being super confusing from a DX perspective. 

> Also i'm a bit unsure about the wanted level of verbosity, do we want to keep it fairly simple as it is now, though i'd like to convert it into tables like here https://vitejs.dev/guide/cli.html, or do we want it a bit more text-like and split-up like how astro does it https://docs.astro.build/en/reference/cli-reference/

I think I prefer the current, non-table format. We can work on adjusting the formatting (because right now it's a ton of code blocks and inline code which can be tough to glance through), but I think this is a really solid foundation.

---

Two more points:

- One place I'd consider a table is at the top of a breakdown of each command and then the one-line description of what it's for/does.
- On the completions I think it'd be better to not break out the Bash, Zsh, etc. sections and instead any of that relevant helper text should be in the CLI itself. Makes it to where if upstream changes anything there that we don't get doc mis-matches either.

Thanks for being amazing @FabianLars <3

lorenzolewis:
Also, in addition to the table I mentioned, I think a quick tabbed code block of how to install the CLI would be amazing up at the top.

lorenzolewis:
One final thing: upstream is working on some fancy stuff that we can do in code blocks, so that might help the walls of code blocks be a bit more formatted once it's in: https://github.com/withastro/starlight/issues/461#issuecomment-1666077060

(I really should just let my replies sit for a few minutes before I hit comment so I don't post back to back replies 😅 )

src/content/docs/2/reference/cli.mdx
```diff
@@ -0,0 +1,508 @@
...
+
+```sh
+pnpm add -D @tauri-apps/cli@next
+```
+
+  </TabItem>
+  	<TabItem label="Cargo">
+
+```sh
+cargo install tauri-cli@next
```
lorenzolewis: This is my syntax, is there an equivalent to `next` in Cargo? I only see the alpha numbers in the blog posts so far.

src/content/docs/2/reference/cli.mdx
```diff
@@ -0,0 +1,508 @@
...
+      --no-dev-server             Disable the dev server for static files
+      --port <PORT>               Specify port for the dev server for static files. Defaults to 1430 Can also be set using `TAURI_DEV_SERVER_PORT` env var
+      --force-ip-prompt           Force prompting for an IP to use to connect to the dev server on mobile
+  -h, --help                      Print help
+  -V, --version                   Print version
+```
+
+This command will open the WebView in development mode. It makes use of the `build.devPath` property from your `tauri.conf.json` file.
+
+If you have entered a command to the `build.beforeDevCommand` property, this one will be executed before the dev command.
```
lorenzolewis, line 123: Add directly to the CLI helper text upstream instead of here

src/content/docs/2/reference/cli.mdx
```diff
@@ -0,0 +1,508 @@
...
+Usage: cargo tauri info [OPTIONS]
+
+Options:
+      --interactive  Interactive mode to apply automatic fixes
+  -v, --verbose...   Enables verbose logging
+  -h, --help         Print help
+  -V, --version      Print version
+```
+
+`tauri info` shows a concise list of information about the environment, Rust, Node.js and their versions as well as a few relevant project configurations.
```
lorenzolewis, line 172: Add directly to the CLI helper text upstream instead of here

src/content/docs/2/reference/cli.mdx
```diff
@@ -0,0 +1,508 @@
...
+  </TabItem>
+  <TabItem label="Cargo">
+
+```powershell
+cargo tauri completions --shell powershell > ((Split-Path -Path $profile)+"\_tauri.ps1")
+Add-Content -Path $profile -Value '& "$PSScriptRoot\_tauri.ps1"'
+```
+
+  </TabItem>
+</Tabs>
```
lorenzolewis, line 424: Add directly to the CLI helper text upstream instead of here

src/content/docs/2/reference/cli.mdx
FabianLars: nope. only stuff like `cargo install tauri-cli --version 2.0.0-alpha` exists but no extra tags like in npm world afaik

lorenzolewis: Alright, so then this just might be a necessary evil until we switch 2.0 over to release then. I'll just add one of those fancy comments to help us remember when that release time comes.

lorenzolewis:
Notes that I'll open a tracking issue upstream for:

- Competitions instructions should be inline in the CLI and not something in the docs
- `plugin add` should be renamed to `add`
- Any additional information (like in the `info` and `dev` commands) should be added directly the the CLI upstream

If/when those are resolved then we can come back and revise this.

Does everything else look alright to a second pair of eyes?

FabianLars:
Do we want to make all section use tabs for commands in this pr or later when https://github.com/tauri-apps/tauri-docs/issues/1366 is resolved?

otherwise looks good ig

lorenzolewis:
Ahh good point, I read that but missed it on the current pass. That component should be _super_ simple to make so it might be better to just whip it up and do it right the first time.

lorenzolewis:
Closes #1366

src/content/docs/2/reference/cli.mdx
lorenzolewis: Tracking in https://github.com/tauri-apps/tauri-docs/issues/1373

src/content/docs/2/reference/cli.mdx
lorenzolewis: Tracking in https://github.com/tauri-apps/tauri-docs/issues/1373

src/content/docs/2/reference/cli.mdx
lorenzolewis: Tracking in https://github.com/tauri-apps/tauri-docs/issues/1373

lorenzolewis:
Looks good now, going to merge it in 💪

PR #1367, opened 2023-08-07
feat: dev container

### LINKED ISSUE #1362: Create DevContainer Configuration for Repo
Would like to setup a nice dev container config for the `next` branch so that GitHub Codespaces get setup correctly.

simonhyll:
Resolves #1362 

I added some extra extensions that are installed automatically when you open the dev container.

Simply start it and run `pnpm i && pnpm dev`


.devcontainer/Dockerfile
lorenzolewis, line 1: Isn't there a JS base or something we can use? I'd really love to keep this all as simple as possible and rely on templates/images already out there if at all possible (I can just see this becoming out of date so easily with time and never getting touched).

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
+{
+  "name": "Tauri Docs",
+  "runArgs": [],
```
lorenzolewis, line 3: Not sure if the devcontainer stuff is smarter, but I had to set the gitpod one to specifically call `pnpm i` and `pnpm dev` otherwise it defaulted to `npm`. Also had to forward port 3000 (astro default for now until next major version).

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,28 @@
...
+  "mounts": [],
+  "build": {
+    "dockerfile": "Dockerfile",
+    "args": {}
+  },
+  "customizations": {
+    "settings": {
+      "extensions.installRecommended": true
+    },
+    "extensions": [
```
lorenzolewis, line 14: If these extensions are in the `.vscode` folder/config do they get picked up by the dev container? I think it'd make more sense to include them all there if so (that way people developing outside of a dev container still get the recommended "ideal" setup).

lorenzolewis (changes requested):
Thank you so much for putting this in! I haven't tested the full setup yet but put a few comments for feedback and to get your thoughts.

.devcontainer/devcontainer.json
simonhyll: Haven't tested that 🤔 I'll give it a try. Otherwise a functionality difference could be that what we have in .devcontainer gets installed for everyone like it or not (like the Astro extension) while the rest are just recommended. Or we could simply put all extensions in both, it's not *that* big a deal I feel to have just the list of extensions in two places since preferably the list of extensions should be kept fairly limited so the environment doesn't have a million tools only a handful of people even know what they do

.devcontainer/devcontainer.json
simonhyll: I can take a look at making it Gitpod compatible, didn't think that's something we wanted 🙂

.devcontainer/Dockerfile
simonhyll: Is it the `curl -sL "https://deb.nodesource.com/setup_${NODE_VERSION}.x" | bash -E -` part you're worried about? Because that has been the same since like, idk, 2015 or something. Using a JS base image could skip that part I guess but I'm not sure which JS image to use

.devcontainer/devcontainer.json
simonhyll: Alright I've pushed some changes now that fixes Gitpod compatibility 🙂 

We should add a link to this page for people that wanna use Gitpod so your browser doesn't block popping up the documentation site tab:
https://www.gitpod.io/docs/configure/user-settings/browser-settings

.devcontainer/devcontainer.json
simonhyll: I put the list of extensions in both files now, it's a short list, I think it's fine

.devcontainer/devcontainer.json
lorenzolewis: Do you want to add that to the contributing guide? Just super simple to say "learn more about developing on GitPod here" or something like that.

.devcontainer/devcontainer.json
lorenzolewis: So Codespaces (and maybe Gitpod) don't read from `.vscode` and need them separately defined, right?

.devcontainer/Dockerfile
lorenzolewis: This is what I've seen and had in mind: https://github.com/withastro/starlight/blob/main/.devcontainer/Dockerfile

I just really want something we can point to, let it deal with all the OS setup, deps, etc. so that we don't have to maintain it. The KISS method ;)

lorenzolewis:
Pulling this from the [Discord conversation](https://discord.com/channels/616186924390023171/662624589037436928/1138477121031917569):

> Btw @Lorenzo speaking of Gitpod, how is contributing meant to work with that exactly? I mean when I wanna contribute I fork the repo, but now that Gitpod link in the README isn't valid anymore because it points to the main repo instead of my fork. One alternative could be to open gitpod based on the main repo, but now I need to be a smart Git user in order to change the upstream branch to my fork of the docs. Like, if everyone has write access to the main repo it's easy, then it's just open gitpod and start making new branches, but if we rely on forking then making PR's then it's kind of a pain

The flow I have in mind is that a person _starts_ from our repo, they don't have a fork. GitPod, Codespaces, etc. are for those people who don't care about forking, cloning, setting up on their local machine, etc. It's for people who want to contribute a typo fix, a translation, etc. and not worry about all of that. 

At least in VSCode locally, if I try to "push" to a repo that I don't have write access to it will suggest to me to fork it and open a PR which has been a flow that's worked for me (might need to confirm what that experience is through Gitpod/codespaces to confirm). 

> For that reason I'd suggest we set up protected branches and give the general public write access to branches starting with contrib/, then you can skip forking altogether, you just open Gitpod, run git checkout -b contrib/my-feature and bam, you're off to the races

I think my reply above addresses this a bit, but going to a put a few thoughts in general on this topic. In theory I _love_ the idea of making it more collaborative and accessible for people to work together, but this could potentially have a lot of security things we have to think through (already had a bit of a scare with some PRs taking advantage of some actions a few months back). 

Plus, contributors have the check box to "allow maintainers to edit this pull request" which I think will cover any "collaboration" I'd expect tbh. If we hit cases in the future where non-maintainers want to collaborate with each other then we can cross that bridge when we get there.

.devcontainer/Dockerfile
simonhyll: I switched to using `mcr.microsoft.com/devcontainers/javascript-node` to skip installing Node.js. The rest in the image however I really don't see a good way around, nor a need to offload to someone else, the rest is just general utilities and convenience stuff.

.devcontainer/devcontainer.json
simonhyll: I can write a short section about Gitpod development in the contributing guide, sure, just gotta figure it out myself fully first x)

.devcontainer/devcontainer.json
simonhyll: Not sure, they should afaik, but they might just not honor the request to install them by default so people have to manually do it from the recommended tab, not sure

.devcontainer/devcontainer.json
lorenzolewis: That'd be awesome! I need to play around with this a little bit more myself as well, so doesn't have to be anything too formal as I'm sure we'll adjust it as we learn over time.

lorenzolewis:
Going to do a bit of testing on this one later on. I'm still eye-ing that docker file a bit tbh and if it needs to have that many "convenience" things. But that one might be better to talk through a bit on Discord later.

One concern I have with this is commit signing. This doesn't necessarily change the difficulties we have with commit signing today. I'll maybe start a separate discussion on this to get some insight and feedback from the security team.

simonhyll:
> that many "convenience" things
It's mainly just these:
- Tabbing: I mean, you *can* get rid of it, but it's really nice for navigating the filesystem when you're a terminal user like myself
- Adding sudo: Just lets you run `apt install` and such in case something is missing in the devcontainer that people wanna use
- Apt clean: makes the image a bit smaller
- Package managers: We can remove the yarn and npm parts since we use pnpm. The line for pnpm is effectively the same as in starlight
- All the apt install stuff: Sets up `git` and `git-lfs` primarily which we should be using more than we seem to be, and installs some utilities for `apt` so that people don't run into issues when using it

The arguments at the top *can* be removed if you want to make it shorter but they're really not something you need to maintain, they just effectively expose those variables so that you can set them in `devcontainer.json` instead, in case someone needs to override them for some reason.

All in all I'd say that yes, we can take the image down to like 2-3 lines long, but the extras I've added don't actually involve any added maintenance, it's mainly just best practises for docker images and convenience stuff.

simonhyll:
> One concern I have with this is commit signing
This is something I had issues with too. In order for commit signing to work you really do need to one way or another get your credentials into Gitpod or wherever you add it, and I'm personally not at all comfortable with putting my GPG key in Gitpod. Sure I could make a new SSH key for contributing, but if that gets compromised I'll have to remove it from my account nonetheless resulting in all my signed commits no longer being signed.

Like, I don't see any solution for using Gitpod that doesn't involve putting keys in a place I'm not comfortable with, other than making unsigned commits first to your own fork of the project, then retroactively signing them, but that's not all that beginned friendly, which I feel is what Gitpod was mainly intended to solve.

lorenzolewis:
The points make sense. I'll trust your guidance on this one as long as we keep maintainability front of mind.

On commit signing, I pinged security in Discord for their input but will also do it here.

@tauri-apps/wg-security Can you all give some input on yes/no to NOT require commit signing in the docs repo? I know that _ideally_ we'd want it, but I'm really thinking of if there are any actual risks here that outweigh a (very high imo) barrier to entry for new people contributing.

tweidinger:
We enforce commit signing for several reasons from my perspective but we don't have a central document describing the reasons (or at least I don't know of it). In general signing is just attribution of changes to someone.

For us this means whoever commits to our projects is actually identified properly through git commit signatures and not just through Github itself.
So what this protects against is that someone impersonates Lorenzo and is committing some hidden malware into the website/docs and we are all angry at him for doing such mean things. We would still have to merge these changes ;)

Signing does not protect us against missing reviews of the PR and merging malicious code. So the thing we are using commit signing for is for simple attribution of whoever wrote the code. 

Having enforced reviews, ideally by 2 (core-)people, I think this is much more worth than enforcing commit signing for documentation changes from external contributors. However, signing should be the default and encouraged wherever possible. Unsigned code will then stick out and reviewers will have a more critical perspective when checking out the code.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,27 @@
...
```
tweidinger, line 14: Not a big fan of pushing all those third party extensions in the container by default. Are they needed for usage or do they provide convenience from your personal perspective?

simonhyll: The Astro one at least is basically needed for usage. Several of the others are less about being "needed" and more about providing developers with a similar setup of tools related to things like linting and task management. Technically all extensions are always optional, Notepad is the best IDE after all x)

Also it's reeeally annoying to have to manually install all recommended extensions when you load up the repo, this is the only way to make it as quick and easy to just boot up and start editing with. Remember that if it's ultimate customization you want a devcontainer may not be the best to use in the first place, and the main purpose of this devcontainer is to get people up and running quickly.

That's my take on it at least, but I'm not impossible to convince otherwise. I'm not super emotionally attached to the devcontainer.

lorenzolewis: I'd rather trim it down to only the Astro extension. The idea of having a dev container in the first place is to get something spun up super quick and easy and we should keep the plugins to a minimum to achieve that. 

If we slim it down to only Astro's plugin then I think `extensions.installRecommended` is fine.

.vscode/extensions.json
```diff
@@ -1,4 +1,14 @@
...
+      "dbaeumer.vscode-eslint", // ESLint
+      "waderyan.gitblame", // Shows blame in source code
+      "GitHub.vscode-pull-request-github", // Pull requests
+      "codezombiech.gitignore", // .gitignore
+      "numso.prettier-standard-vscode", // Prettier
+      "Gruntfuggly.todo-tree", // To do
+      "MarkosTh09.color-picker", // Color picker
+      "redhat.vscode-yaml", // Yaml
+      "astro-build.astro-vscode" // Astro
+  ],
```
lorenzolewis, line 12: Some comment on this, let's only do the Astro plugin

.devcontainer/Dockerfile
lorenzolewis, line 1: I want to clean this up to be as minimal as possible. Looking at the container image itself at https://github.com/devcontainers/images/tree/main/src/javascript-node, the only thing I think we need that isn't available here is PNPM. Can we run `RUN npm install -g pnpm` so that we don't need to deal with `apt` at all?

I think the only thing we need in here is:

1. `FROM` to specify the image and variant
2. The `RUN` command above to get PNPM

I don't think anything else is needed to get a basic setup going for a user (including the `EXPOSE`)

lorenzolewis (changes requested):
Let's get this one stripped down to the most basic version possible. I want to rely on images as much as possible in here and not have the possibility of something getting out of date and broken for something as non-critical as this.

PR #1374, opened 2023-08-09
Fixes for non-main ref in translation status page

lorenzolewis:
@dklassic would you be able to test the contributing page now to make sure all the links work as expected? We won't really be able to test the "Translations that need reviews" section until we cross that path, but I think this should resolve all the broken links I hit yesterday during the call.

Might still be a few weird things around CI, but I'll fix that separately because it won't impact the i18n flow

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -29,6 +29,7 @@ export default defineConfig({
 				twitter: 'https://twitter.com/TauriApps',
 				mastodon: 'https://fosstodon.org/@TauriApps',
 			},
+			// TODO: Be sure this is updated when the branch is switched
 			editLink: {
 				baseUrl: 'https://github.com/tauri-apps/tauri-docs/edit/next',
 			},
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -30,7 +30,7 @@ export default defineConfig({
 				mastodon: 'https://fosstodon.org/@TauriApps',
 			},
 			editLink: {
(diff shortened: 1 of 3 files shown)


lorenzolewis:
(kicks Netlify's CI): here's the rendered dashboard btw: https://deploy-preview-1374--tauri-docs-starlight.netlify.app/contribute/translate-status

FabianLars:
ngl that dashboard is fancy af

dklassic:
> @dklassic would you be able to test the contributing page now to make sure all the links work as expected? We won't really be able to test the "Translations that need reviews" section until we cross that path, but I think this should resolve all the broken links I hit yesterday during the call.
> 
> Might still be a few weird things around CI, but I'll fix that separately because it won't impact the i18n flow

All links from Contribution Page seems fine, but I believe "Edit page" on individual pages still isn't referenced properly, not sure if you're fixing it with this pr as well. If not, than all seems fine :)

dklassic (approved):
nvm didn't notice the issue title. All seems to be working, let's goooo

lorenzolewis:
> ngl that dashboard is fancy af

That's why I "borrowed" it ;) I've been thinking about asking if they want that turned into an official package/integration upstream, but I feel like they might not because it's for a very specific i18n flow.

lorenzolewis:
Fixed the edit link, thanks for jogging my memory @dklassic 😅

PR #1375, opened 2023-08-10
feat: added 3d hero logo

simonhyll:
I made an animated 3D version of the logo with mouse interactions. Still want to tweak the appearance a bit, thinking of adding colors to it. Up for discussion. Still need to clean up the code a bit.

I've cleaned up the code now and added comments, so I think it's ready to be shipped. Minor restyling as well from the gif I sent, 1px instead of 2px for the lines, and less padding.

Animation


simonhyll:
@lorenzolewis What did we say about this one? Are we merging it as-is for the time being and then in a later PR switching to Three.js? If you feel it can be merged feel free to push the button 🙂

lorenzolewis:
I put it on you and @AlveLarsson to have a 3D framework war and see which framework comes out on top 😆 

But this is the criteria I'd noted down to check and detail out:

- What's the change in bundle size?
- Which framework balances future maintainability for contributors with minimal bundle size best?
- Check how this performs across different browsers and OSs (I'm very curious on how this performs inside WebKit on Linux inside a Tauri app because I know there have been performance issues there)

After we have those decided/answered then would love to get this one merged in and out in the wild. It's a good chance to have it tested and see if there are any weird bugs before diving into the fully 3D layered Tauri sandwich idea @AlveLarsson has.

I'd also ask for @AlveLarsson 's input on the style of the 3D model, if we want to go with the current wireframe or if we want to instead go for one that's closer to the Tauri logo (filled and with the yellow/blue colours, keeping in mind contrast for a light theme when we build it).

dreyfus92:
I don't think that the report presented by Lighthouse is accurate, there's always an impact of performance by shipping that much JS. I like the animation, but I agree with Lorenzo when it comes to browser compat. This `sc` was taken by running Lighthouse for mobile devices, which is inaccurate cause it will vary depending the device.
<br/>


image

simonhyll:
There is an impact on performance, in my test it went from 99 to 88 in Lighthouse. The solution I have is fairly efficient, and imo 80+ is fine since it's only on the landing page.

If we switch to three.js the bundle size would increase more but the performance and capabilities would go up, and maintainability would be easier since it's a large framework that's been around forever.

I don't currently have the capacity to improve upon this solution. It was mostly a fun side project.

lorenzolewis:
Should we go ahead and close this out? I know even with our current (minimal compared to this) implementation we were getting a few reports of stuttering animations, so would rather error on the side of caution with this (plus free up some of @simonhyll 's time ;) )

simonhyll:
We'll use what we have instead. Was a fun little project though. :)

PR #1376, opened 2023-08-10
feat: added demo component

simonhyll:
Adds a little demo component for conditionally displaying something when the site is being ran as a frontend for the Tauri App.

Demo might not be the best name, something like "TauriOnly" might be better. Plus it's my first Astro component ever made and after making the Hero component I kinda wanna redevelop it.

But, those things aside, it works and does what it should do. The most important thing for me right now is deciding what the component should be called so we can start using it, I can fix the code in another PR.

### DIFF
--- a/src/components/Demo.astro
+++ b/src/components/Demo.astro
@@ -1,22 +1,8 @@
 ---
 import { Card } from '@astrojs/starlight/components';
-const isTauri = process.env.TAURI_DEBUG
+const enabled = process.env.TAURI_DEBUG && Astro.slots.has('default');
 ---
 
-{
-  isTauri ? (
-  <Card title="Demo" icon="document">
-    {
-      Astro.slots.has('default') && (
-        <>
-          <slot />
-        </>
-      )
-    }
-  </Card>
-  ) : (
-  <Card title="Demo" icon="document">
-    <p>You can download the Tauri Documentation app here: <a href="#">INSERT LINK</a></p>
-  </Card>
-  )
-}
+<Card title="Demo" icon="document">
+  {enabled ? <slot /> : <p>Try it in the Tauri Docs app (coming soon)</p>}
+</Card>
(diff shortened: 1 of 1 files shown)


lorenzolewis:
Take a look at this one: https://docs.astro.build/en/core-concepts/astro-syntax/#dynamic-html

Any scripts between the `---` block are executed at build time (thinks explicitly in a `<script>` block are what make it to the browser). So you could set a variable there to determine if you're building for Tauri. Then use that variable to conditionally render using that JSX-like syntax.

Build only results in the HTML you need for that specific build. Should let you simply things a bit I think.

simonhyll:
There we go @lorenzolewis now it's not quite as dirty anymore 😄

lorenzolewis:
Just did a few refinements to de-dup some component stuff and update the copy. Here's a screenshot of it now:

Screenshot 2023-08-16 at 09 39 38

Before merging in I have a few questions on how this would be integrated into a page:

- What's the future behaviour when viewed on the web? Is it going to have a button to just open it in the Tauri app?
- What's the future behaviour when viewed within a Tauri app? Is it just going to contain buttons/controls to execute and adjust the behaviour? Similar to the example API app in the Tauri repo (checkboxes, buttons, etc.)?

image

Depending on what you have in mind for these then I may have some refinement suggestions on the UI.

I think https://github.com/tauri-apps/tauri-docs/pull/1401 could have some impact into this, but for the sake of discussion maybe we could pretend we're in a little bubble and we control all the scripts, styles, components, etc. here and then bring that back to how it relates to #1401

simonhyll:
So yea it's something along those lines I'm thinking it would look like. The main function of the Demo component is just ensuring you get either a standard message stating how you can get the app downloaded when you're in a browser, or show whatever is passed to its default slot if it's running in the app. Any UI related stuff should be taken care of inside each page directly, might need some extra components to make them all look similar with buttons and inputs and such.

lorenzolewis (approved):
Coolio, let's get this one merged in!

PR #1382, opened 2023-08-12
Add set up installations for Enterprise Linux distributions (RHEL)

mzen17:
Minor PR:
Redhat Enterprise Linux 9 and before use a Fedora version <36, (based on Fedora 34), so this is a minor PR that adds clarity for those that are not using Fedora and instead using RHEL/ one of its downstreams.

Basically the Fedora one copied over, but changed the name of C Development tool to "Development Tools" and defaulted to webkit2-3.

### DIFF
--- a/docs/guides/getting-started/prerequisites.md
+++ b/docs/guides/getting-started/prerequisites.md
@@ -151,7 +151,7 @@ sudo dnf group install "C Development Tools and Libraries"
 Note that on Fedora 36 and below the `webkit2gtk4.0-devel` package was called `webkit2gtk3-devel`.
 
   </TabItem>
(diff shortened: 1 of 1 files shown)


FabianLars:
Thanks for contributing!

Honest question, so feel free to say no: Do you think it makes sense to include this in the fedora section similar to the fedora<36 note? - Edit: one could also rename the fedora section to fedora/rhel if that changes anything 🤷

mzen17:
I think it would make sense, as they are just down streams of each other.

I'll go fix it.

lorenzolewis:
Good call to collapse them together @FabianLars

FabianLars (approved):
thank you!

PR #1385, opened 2023-08-13
Update gentoo webkit-gtk package slot: 4 => 4.1

markfswe:
Fix outdated package version (for Gentoo only)

With webkit4.0 installed I get this error:
```
error: could not find system library 'webkit2gtk-4.1' required by the 'webkit2gtk-sys' crate
```


FabianLars:
Thank you!

PR #1395, opened 2023-08-15
Feat/link validator

lorenzolewis:
Currently just keeping as a draft while some upstream issues are worked through.

## Upstream issues:

- [x] https://github.com/HiDeoo/starlight-links-validator/issues/3
- [x] https://github.com/HiDeoo/starlight-links-validator/issues/4
- [ ] typedoc-plugin-markdown isn't building paths correctly when `entryFileName: index.md` is set (including index in filename). Need to open an issue on this
- [x] A few broken links I need to fix upstream inline with Tauri code

There may be additional issues with `typedoc-plugin-markdown` that I noticed in the output. One example is this link:

`/2/reference/core/js/Namespace.event.md#windowlabel`

It should instead be in this format:
`/2/reference/core/js/namespaceevent#windowlabel`

It could be another issue that I haven't had time to fully dig into yet, so I'll see what I can come up with on that.


lorenzolewis:
Got this one closer with the update from `starlight-links-validator@0.2.0` (special thanks to @hideoo for that one)

Before:
`Found 272 invalid links in 23 files.`

After: 
`Found 254 invalid links in 20 files.`

The remaining I think are due to an upstream issue in how links are generated that point to `index.md` files. I'll put an issue together to address that one in the morning.

Also noticed there's a few more typos in upstream APIs in Tauri I'll need to PR to get fixed.

lorenzolewis:
Looking into this more there are 2 issues I've found:

- Using the `propertiesFormat = 'table'` option causes an issue as properties in a table don't have an `id` since they aren't a heading. The same can be seen if referencing enum members and using `enumMembersFormat: 'table'`
- A `@link` inline tag within a summary isn't using the overridden `relativeURL` property. The logic path can be found here: https://github.com/tgreyuk/typedoc-plugin-markdown/blob/292b8b953cfdd667337893f881d26ed67ae16df3/packages/typedoc-plugin-markdown/src/theme/resources/partials/comment.parts.ts#L30. I'm not sure why this isn't the case as any other references seem to be setup the same and correctly using the property.

lorenzolewis:
Not an upstream issue, but https://github.com/HiDeoo/starlight-typedoc/issues/14 covers the same problem that's happening here with `{@link}` not having it's URL generated correctly.

PR #1397, opened 2023-08-15
feat: added `fr` translation for "Why Tauri?" page

dklassic:
> [!NOTE]  
> This is a test commit to see how the pr workflow would look like and should not be merged.
> 
> This Pull Request is powered by GPT-4 to provide a French translation of "Why Tauri?" page.

I've translated "Why Tauri?" page for `fr`.

I'm not sure if the English video "Tauri in 100 Seconds" from the source page should be translated into "Tauri en 100 Secondes" or not. Maybe I should leave it as "Tauri in 100 Seconds (Vidéo en anglais)".

> Need places to mention page (how? by name and/or by path?) and target language.
> Need a place to mention concerns about the translation to ask for feedback.

Extra notes:
- How should the title be formatted? and again, how should the source page be reference in title?
- Should pr be enforced to be written in English? or is it okay to write entirely in target language?
- How to ask for review from locale specific reviewer?
- What is the `description:` for? The source text doesn't seem to have it.
- Is translator allowed to swap out resources to a more locale specific choice? e.g. a French video explaining Tauri in this case.

### DIFF
--- a/src/content/docs/fr/2/guide/index.mdx
+++ b/src/content/docs/fr/2/guide/index.mdx
@@ -1,0 +1,26 @@
+---
+title: Pourquoi Tauri ?
+description:
+---
+
+import { Card } from '@astrojs/starlight/components';
+import Stub from '@components/Stub.astro';
+
+<Stub />
+
+:::caution[Logiciel en Pré-sortie]
+Cette documentation concerne la version en pré-sortie de Tauri 2.0 et est susceptible de changer au fur et à mesure que le développement se poursuit. Si vous recherchez la documentation de Tauri 1.0, visitez https://tauri.app.
+:::
+
+<Card title="Prochaines Étapes" icon="rocket">
+	Préparez-vous à créer un projet Tauri en installant les
+	[prérequis](/2/guide/prerequisites).
+</Card>
+
+<Card title="Ressources Supplémentaires" icon="document">
+
+- 🎥 [Tauri en 100 Secondes](https://www.youtube.com/watch?v=-X8evddpu7M), Fireship (YouTube)
+  {/* - 🎧 [Podcast] */}
+  {/* - 📝 [Article] */}
+
+</Card>
--- a/src/content/docs/fr/2/guide/index.mdx
+++ b/src/content/docs/fr/2/guide/index.mdx
@@ -1,6 +1,5 @@
 ---
 title: Pourquoi Tauri ?
-description:
 ---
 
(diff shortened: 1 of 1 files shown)


src/content/docs/fr/2/guide/index.mdx
```diff
@@ -0,0 +1,26 @@
+---
+title: Pourquoi Tauri ?
+description:
+---
```
lorenzolewis:
```suggestion
---
title: Pourquoi Tauri ?
---
```

lorenzolewis:
`description` is optional, it's mainly used for SEO: https://starlight.astro.build/reference/frontmatter/#description

Just removed it to get the build running for now

lorenzolewis:
So it looks like the status page on this branch gets updated as expected: <https://deploy-preview-1397--tauri-docs-starlight.netlify.app/contribute/translate-status>.

Looks like the "needs review" stuff isn't getting updated appropriately: <https://beta.tauri.app/contribute/translate-status>. I did a "clear cache and build" on netlify to force-update it but I think the code is looking for a specific label or something. I'll dig into that later and see what we need to get it working.

lorenzolewis:
Alright, magical trick is to make sure it has the `i18n` label, you can see it in the dashboard now: <https://beta.tauri.app/contribute/translate-status>.

Next question: how do we automate that 😅 

What Starlight is doing is using the labeler workflow: https://github.com/withastro/starlight/blob/main/.github/labeler.yml and https://github.com/withastro/starlight/blob/main/.github/workflows/pr-labeler.yml

Any issues with that approach or should I build that out?

dklassic:
lgtm!
I guess the only "issue" is that we'll have to manually add in each language, but I guess that's already the case for our language support so why not!

lorenzolewis:
@dklassic do you want to do another PR and test that everything gets caught correctly? Can close this one and re-open it if you like I guess.

I do know that I'll need to think about how we keep the tracker up to date on the live site. Maybe we deploy that as a separate site that's run on a new PR with the right label 🤔. But I'll note that in the meta issue to track and tackle a bit later

dklassic:
> The configuration file (path: .github/labeler.yml) isn't not found locally, fetching via the api

Okay something needs fixing

So the labeler.yml has to be in the pr or it'll break, makes sense.

lorenzolewis:
Should be fixed by merging from next into this branch then re-opening

PR #1401, opened 2023-08-15
feat: added examples-generator

simonhyll:
Started work on something that fetches examples from upstream so we can focus on writing recipes there instead.

Currently I'm calling it `examples-generator` and keeping it as a separate thing, but I would prefer it if we went all in on this and made it the `recipe-generator`. If we decide that recipes are different from examples we could still use this to raise awareness of the examples folder seeing as so many people don't know it's a thing.

For this PR I've manually edited the upstream README's to make them compatible.

TODO:
- [ ] Fix upstream examples, their README's are either bad, missing or causes errors (mainly an image that isn't required or that we could look into pulling as well if that's desired)

### DIFF
--- a/packages/examples-generator/build.ts
+++ b/packages/examples-generator/build.ts
@@ -1,0 +1,18 @@
+import { join, dirname } from 'node:path';
+import { readdirSync, writeFileSync, readFileSync, mkdir } from 'node:fs';
+import { fileURLToPath } from 'node:url';
+
+const rootPath = dirname(dirname(dirname(fileURLToPath(import.meta.url))));
+const examplesPath = join(rootPath, 'packages', 'tauri', 'examples');
+const examplesOutputPath = join(rootPath, 'src', 'content', 'docs', '2', 'example');
+
+async function main() {
+	console.log('Copying examples', examplesPath);
+	for (const example of readdirSync(examplesPath, { withFileTypes: true })) {
+		if (example.name.startsWith('.') || example.isFile()) continue;
+		const sourcePath = join(examplesPath, example.name, 'README.md');
+		const outPath = join(examplesOutputPath, example.name) + '.md';
+		writeFileSync(outPath, readFileSync(sourcePath, 'utf-8'));
+	}
+}
+main();
--- a/packages/examples-generator/package.json
+++ b/packages/examples-generator/package.json
@@ -1,0 +1,15 @@
+{
+  "name": "examples-generator",
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "scripts": {
+    "build": "tsm ./build.ts"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "tsm": "^2.3.0"
+  }
+}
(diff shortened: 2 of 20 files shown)


lorenzolewis:
While this is a good idea to expose these examples more, I don't want to add more cross-repo syncing stuff. I'd much rather do a simple link to the pages where it's relevant for a certain example and possible an index page for them all. I think automating this is going to cause us more headaches than it's worth than to just manually put in some links.

It's similar to why I decided to not pull in the plugin readme and instead just link out. Any broken links in those readme's, unresolved images, broken links, etc. just break the docs build and also add one more thing for upstream to keep in consideration instead of just being able to code cool stuff.

simonhyll:
The main reason I want this sync is because it forces upstream to maintain their examples better. Currently the README's are either missing or incomplete, and they don't get the love they deserve. It would improve their quality and provide Tauri with more integration tests.
cc @FabianLars @lucasfernog @amrbashir for input since this would affect the `tauri` repo.

FabianLars:
> The main reason I want this sync is because it forces upstream to maintain their examples better

I think at this point this statement is almost desilusional. I agree that they need to be better all around but i don't think exposing them on the docs page change anything in this context. Also i'm not sure if they really serve the same purpose as docs example do. Or maybe i think too much about the api example since that's mostly used to test tauri itself, not to showcase it (or so i've heard, i don't use it myself).

Either way i think the docs deserve better and by making the examples primarily for the docs site i think we can make them better than having them be in the repos first and in the docs page as a second thought. (i'm thinking about tightly integrated examples however that would look like)

amrbashir:
I remember a time when I used to direct users to some of the examples in the repo which was good to an extent but as time went by, the examples got a bit complicated (mainly the API example) and had less maintenance efforts which at the moment just receives the bare minimum effort to just make it compile.

The `api` example was actually meant to be a showcase for the tauri APIs so users can go to tauri.app and download a precompiled version and see what tauri is capable of. It can also be used as a test bed for new tauri features.

I am not opposed to this idea but ofc this introduces more maintenance burden on us, so If someone is willing to maintain the examples and keep them beginner-friendly and actually explain one concept and explains it well with a README and some in-code comments, I'd be okay with it, otherwise No.

lorenzolewis:
To @amrbashir 's and @FabianLars 's points, I really don't want to turn those examples there into something that _has_ to be polished to the level we're wanting to achieve in docs. I really want dev to be able to do anything and everything they want with them.

I don't want to discourage this idea and it's a fantastic one, but I'm not sure if we're there quite yet. I'd much rather ask the dev team to spend any time polishing inline API docs, helping out with docs questions as we have them, and to also focus on getting 2.0 shipped than to add another thing on their plate.

Also, just having to maintain these sorts of inter-repo connections sends shivers down my spine 😅 I still want to work on carving out some of the stuff I have just for the API generation to make it less inter-connected (but add that to the pile for now until it breaks)

CC @nothingismagick since I remember you'd advocated on Discord for this for any additional thoughts you'd like to share.

lorenzolewis:
Going to go ahead and close this one now as not planned. Awesome idea but let's get some content written and published first before diving into this.

PR #1403, opened 2023-08-16
Add labeler for i18n

lorenzolewis:
Using https://github.com/tauri-apps/tauri-docs/pull/1397 to test


dklassic (approved):
Let's merge it in and see it in action, I think!

PR #1404, opened 2023-08-16
feat: added `fr` translation for "Why Tauri?" page

dklassic:
> [!NOTE]  
> This is a test commit to see how the pr workflow would look like and should not be merged.
> 
> This Pull Request is powered by GPT-4 to provide a French translation of "Why Tauri?" page.

I've translated "Why Tauri?" page for `fr`.

I'm not sure if the English video "Tauri in 100 Seconds" from the source page should be translated into "Tauri en 100 Secondes" or not. Maybe I should leave it as "Tauri in 100 Seconds (Vidéo en anglais)".

> Need places to mention page (how? by name and/or by path?) and target language.
> Need a place to mention concerns about the translation to ask for feedback.

Extra notes:
- How should the title be formatted? and again, how should the source page be reference in title?
- Should pr be enforced to be written in English? or is it okay to write entirely in target language?
- How to ask for review from locale specific reviewer?
- What is the `description:` for? The source text doesn't seem to have it.
- Is translator allowed to swap out resources to a more locale specific choice? e.g. a French video explaining Tauri in this case.

Reference:
- #1397


lorenzolewis:
Will trigger a netlify refresh on prod now

dklassic:
I messed up something so here's a new pr.
And there's the error again:

> The configuration file (path: .github/labeler.yml) isn't not found locally, fetching via the api

which is weird because the file is in the directory.

lorenzolewis:
Weird, if I look at Starlight they have the same message in their logs so maybe that's not the issue? https://github.com/withastro/starlight/actions/runs/5872476759/job/15924061052

lorenzolewis:
@dklassic can you try this one again? I had a typo in the labeler that I've fixed in `next`. I also added it to `dev` just in case this action tries to read from only the default branch in some way.

But the action may only refer to the point in time where you branched FROM, not sure if it will refer to the latest commits.

dklassic:
> @dklassic can you try this one again? I had a typo in the labeler that I've fixed in `next`. I also added it to `dev` just in case this action tries to read from only the default branch in some way.
> 
> But the action may only refer to the point in time where you branched FROM, not sure if it will refer to the latest commits.

Seems working now! Nice work!

lorenzolewis:
Awesome! I was reading up on the action and it does look like it pulls from the default branch, so something to keep in mind if we hit this in the future before replacing "dev"

lorenzolewis:
Alrighty, did some more automation magic with the actions. @dklassic could we test one more PR? I think it will need to be re-made from the latest in `next` so that it picks up the updated actions, but essentially it should now automate re-deploying the production i18n dashboard if a PR contains the `i18n` label.

Put one in and then check https://beta.tauri.app/contribute/translate-status to see if it's populated there (usually takes ~2 minutes to build and update)

PR #1407, opened 2023-08-16
Update adapter in SvelteKit guide

isaac-mcfadyen:
- When using SvelteKit adapters, it is no longer recommended to install `@sveltejs/adapter-static@next`, which installs version `1.0.0-next.50` of the adapter.
- Instead, `@sveltejs/adapter-static` (without the `@next` part) should be used, which installs version `2.0.3` or greater of the adapter.
- This PR updates the Tauri SvelteKit guide to use `@sveltejs/adapter-static`, removing the `@next` part so that the latest version is used instead of the old `next` release.

Apologies for the extra commit. It seems I couldn't commit directly to `dev`, even on a fork, so I had to commit to a branch, merge it, and then create a PR to the actual repo.

### DIFF
--- a/docs/guides/getting-started/setup/sveltekit.mdx
+++ b/docs/guides/getting-started/setup/sveltekit.mdx
@@ -80,7 +80,7 @@ First, we need to install [`@sveltejs/adapter-static`]:
   <TabItem value="npm">
 
 ```shell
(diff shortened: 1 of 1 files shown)


FabianLars (approved):
thanks :)

PR #1415, opened 2023-08-18
fix: browserlist typo

hellno:


lorenzolewis:
Thanks for this! I'm going to fix the issue causing it not to build really quickly to get all the green checkmarks in place.

lorenzolewis (approved):
Thank you very much!

PR #1425, opened 2023-08-24
docs: ipc blog post

### LINKED ISSUE #1466: Write IPC Blog post
Rework of the page at https://tauri.app/v1/references/architecture/inter-process-communication/, write as a blog post in new docs

simonhyll:
Writing a little blog post about the IPC. #1466


lorenzolewis:
Haven't looked at the styles/content yet, but for the slug I'd like to move to slug-ifying the post title instead of using a date. We had a few instances where we forgot to update the filename and didn't get posted on the date we were targeting for and that was a bit awkward.

For existing blog posts we'd update the slugs based on this and then set up re-directs to fix any broken links.

lorenzolewis:
Just pushed the blogs integration we're using to next, take a look there for the format and how it comes through.

lorenzolewis:
@simonhyll is this one you'd like to keep pushing forward? If we can get the content written up then getting the diagrams to the state we want should be pretty quick.

simonhyll:
I'll take a closer look at this during the weekend

PR #1513, opened 2023-09-05
Fix: Correct missing `--` from `--version` for Cargo Tauri CLI install

i-c-b:


lorenzolewis (approved):
Awesome, thanks for the catch @i-c-b !

PR #1529, opened 2023-09-08
Fix: Use `@tauri-apps/cli@next` for Upgrade guide

carllippert:
#### What kind of changes does this PR include?

- Minor content fixes (broken links, typos, etc.)

#### Description
- fix package upgrade experience

PR #1530, opened 2023-09-09
Feat: Refactor Guides, Features, and other URLs

lorenzolewis:

- Refactor guides and features
- Refactor sidebar to be dynamically generated

#### Description

We previously talked about not versioning guides for Tauri as they should be fairly common across versions, however that conversation slipped my radar when I began building this out. Although there are changes from version 1, a lot of those can be noted inline in a single page if required and don't warrant a completely different set of pages and versions. This makes it a lot more manageable for us to maintain and also allows users on Tauri v1 to still benefit from all the improvements we're putting into docs.

- `/2/guide` -> `/guides`
- "Feature" and "Recipe" entries under `/2/guide` now moved to `/features`
- "Recipes" have been more unified with "Features" as the reader doesn't necessarily care about the difference but instead the functionality (although there can still be a distinction for docs writers) 
- Change "Features & Recipes" sidebar to be dynamically created based on pages in the `/features` directory
- Refactor `Guides.astro` and `Recipes.astro` to be a single component `Features.astro` generated off of the `/features` directory
- Fix links based on refactor (both English and translated docs)
- Add a "Deep Linking" page for the upcoming plugin using the new `badge` option from upstream

Would like to get this merged in ASAP if we agree with the changes to reduce any refactoring and fixing that may be required of doing this later.


ahqsoftwares:
The preview is lagging in my mobile phone for some reason

lorenzolewis:
Is this on any particular page? Is it the same for the same production page at beta.tauri.app?

ahqsoftwares:
> Is this on any particular page? Is it the same for the same production page at beta.tauri.app?

Only tested on home page rn, doesn't occur in beta.tauri.app (it occurs but not as much compared to here, I think animations is the cause)

ahqsoftwares:
After the animation everything is fine again

lorenzolewis:
Which browser/device are you using?

ahqsoftwares:
Android 12 ONEUI 4.1 Firefox Browser

lorenzolewis:
This shouldn't be related to these changes. The additional "lag" may be because of the Netlify overlay on the bottom. Do you have the same behaviour on https://deploy-preview-1529--tauri-docs-starlight.netlify.app/ from #1529 ?

ahqsoftwares:
> This shouldn't be related to these changes. The additional "lag" may be because of the Netlify overlay on the bottom. Do you have the same behaviour on https://deploy-preview-1529--tauri-docs-starlight.netlify.app/ from #1529 ?

I hadn't tried on my phone that deploy preview I had use browser devtools to simulate mobile, now as it turns out it lags in every preview where the hero icon animates

Issue #1344, opened 2023-08-01
Roadmap to Documentation for Tauri 2.0


lorenzolewis:
As we're prepping for Tauri 2.0 we're also switching over our documentation platform. Now is an awesome time to get involved in Tauri documentation!

Check the milestone for the overall status to having 2.0 docs ready: https://github.com/tauri-apps/tauri-docs/milestone/4

## Writing Contributions

Below is the outline for the new docs. If there is no name next to one feel free to claim it so we can mark your name and then open a PR so that we can start crafting the doc together.

- [ ] Quick Start
  - [x] What is Tauri? @lorenzolewis 
  - [x] Prerequisites
  - [ ] Create a Project
    - [ ] Using CTA
    - [ ] Manually using Tauri CLI
    - [ ] Add mobile to an existing project
    - [x] ...Frontend framework-specific guides
  - [ ] Upgrade and Migration
    - [x] From Tauri v1 @lucasfernog https://github.com/tauri-apps/tauri-docs/pull/1340
    - [ ] Electron to Tauri
    - [ ] ...any other framework suggestions?
  -  [ ] Core Concepts
      - [ ] SPAs, SSGs, etc.
      - [ ] ...what else?
    - [x] Plugins @lucasfernog 
      - [x] What is a Plugin? @lucasfernog 
      - [x] Developing a Plugin @lucasfernog 
      - [x] List of plugins (This page should cover that: https://beta.tauri.app/features/) @lucasfernog 
  - [ ] Troubleshooting
- [ ] Guides
  - [x] #1468
  - [x] #1469
  - [x] #1470
  - [x] #1471
- [ ] Features & Recipes (all listed under https://beta.tauri.app/2/guide/list)
  - [x] #1473
  - [x] #1474
  - [x] #1475
  - [x] #1476
  - [x] #1477
  - [x] #1478
  - [x] #1479
  - [x] #1480
  - [x] #1481
  - [x] #1482
  - [ ] #1483
  - [x] #1484
  - [x] #1485
  - [x] #1486
  - [x] #1487
  - [x] #1488
  - [x] #1489
  - [x] #1490
  - [x] #1491
  - [x] #1492
  - [x] #1493
  - [x] #1494
  - [x] #1495
  - [x] #1496
  - [x] #1497
  - [x] #1547
  - [x] #1548
  - [ ] #1949 
  - [x] #1948 
- [x] References
  - [x] https://github.com/tauri-apps/tauri-docs/issues/1363
[...]

PR #1536, opened 2023-09-10
Use internal-ip@7.0.0 to avoid a serious bug

UnsolvedCypher:
- Update to version of a referenced library

#### Description
The docs currently suggest using the latest version of the `internal-ip` library. However, version 8.0.0 has a [serious bug](https://github.com/sindresorhus/internal-ip/issues/48) and I've updated the docs to recommend 7.0.0.


simonhyll:
Now the question is: how fast do we think it'll be resolved. Because merging this also means we gotta put a reminder somewhere about removing it once they patch the bug. I like providing users with the best available information, the question is just if we want to make this the modus operandi for the project in general or not, because there is no way we're gonna be able to keep up with all these kind of bugs for all dependencies everywhere

Seeing as there is already a PR for a fix, I'm not sure this is worth bringing up, but then again, it's been waiting for merging for a while now 🤔 https://github.com/sindresorhus/internal-ip/pull/49

lorenzolewis:
One way I've been tracking some of these things (although mainly for internal issues/PRs) is to link to the issue/PR inline as a comment in the copy:

```mdx
{ /* Link to issue */ }
```

and then adding it to https://github.com/tauri-apps/tauri-docs/issues/1344

FabianLars (approved):
doesn't look like it will be resolved upstream anytime soon so let's just merge this since users are actively hitting this issue.

p.s. thanks for the contribution :)

Edit: I added it to 1344

PR #1544, opened 2023-09-13
feat: added text about writing code blocks

simonhyll:

- New or updated content

#### Description

Added a section about writing code blocks as per discussions on Discord.


lorenzolewis (changes requested):
Thanks for this! Although it has a lot of useful information on why we got to this, I think this should really be very prescriptive in "here's how we write code blocks". 

Take a look at the [translating guide](https://github.com/tauri-apps/tauri-docs/blob/next/.github/TRANSLATING.md) and see how it's very to-the-point for someone to be able to dive in and contribute without needing to read through all the context of _how_ you we got to a decision.

Hopefully that makes sense. The goal is to make it as straight to the point as possible so that writers can know the expectation and run with it.

simonhyll:
Yea I agree it's too long, just couldn't think of a more concise way of getting the point across than with examples >.< I can take a look at it again but no promises I'll get it done anytime soon, got a lot of my plate already. Might have to let my favourite employee ChatGPT take a look at it lol

lorenzolewis:
No prob. Just getting it all down is the hard bit, word smithing I can try to take a look at a bit later in my evening if I have time and make a few suggestions.

.github/CONTRIBUTING.md
```diff
@@ -96,6 +96,33 @@ Not yet created, will be added in the future.
 
 Topics that are around understanding something can be written as a blog post (we welcome submissions to the Tauri blog from anyone). Blog posts are a good option because they help the reader understand that information is accurate at the time of writing (and of course can always be updated later if it becomes stale). Blog posts follow the goals of [explanation in Diátaxis](https://diataxis.fr/explanation.html), but not all blog posts necessarily fit this specific format.
 
+### Code Blocks
+
+In general when you provide people with code samples you want to give them the highest code quality possible. However, this isn't always aligned with what's best for the user when it comes to showing off a new API. For code blocks meant to convey simple copy-paste friendly snippets of code that aim to show off a certain API we have concluded that it's better for the user if we don't necessarily aim for the highest code quality, but rather the smallest and easiest code to follow with helpful panic messages.
```
dreyfus92, line 100:
```suggestion
In general, when providing code samples, it's essential to offer users the highest possible code quality. However, this isn't always aligned with the best approach for showcasing a new API. For code blocks meant to convey simple, copy-paste-friendly code snippets demonstrating a particular API, we've found that it's better for the user experience to prioritize minimalism and clarity over high code quality, even if that means using helpful panic messages.
```

.github/CONTRIBUTING.md
```diff
@@ -96,6 +96,33 @@ Not yet created, will be added in the future.
...
+
+Take the following two code samples, the first one is code from a previous version of the Tauri documentation.
```
dreyfus92, line 102:
```suggestion
Consider the following two code samples, with the first one being from a previous version of the Tauri docs:
```

.github/CONTRIBUTING.md
```diff
@@ -96,6 +96,33 @@ Not yet created, will be added in the future.
...
+
+```rust
+// Close splashscreen
+if let Some(splashscreen) = window.get_window("splashscreen") {
+    splashscreen.close().unwrap();
+}
+// Show main window
+window.get_window("main").unwrap().show().unwrap();
+```
+There clearly is an attempt made to provide the user with the start of some error handling, something closer towhat they might have in production. However, if a user doesn't have a window labeled `splashscreen` the code won't crash and they'll have to figure out themselves what the issue is. For us to fix the code in the "proper" way we could for example add an else statement and panic in there, but is that high code quality enough? We could switch it to a `match` statement, and add our own custom error, or we could use `?` to raise the error, or we could...
```
dreyfus92, line 112:
```suggestion
In this code, there's an attempt to incorporate error handling, closer to what users might use in a production environment. However, if a user doesn't have a window labeled `splashscreen`, the code won't crash, and they'll have to figure out the issue themselves. To address this in a proper way, we could, for example, add an else statement and panic within it. But is that considered high-quality code? We could also switch to a `match` statement, add a custom error message, or use `?` to raise the error, and so on.
```

.github/CONTRIBUTING.md
```diff
@@ -96,6 +96,33 @@ Not yet created, will be added in the future.
...
+
+There's simply no end to how many improvements can be made to the code provided. By adopting our approach of making simpler more panic prone code we can get the best of both worlds, easy to follow code that provides users with simple to understand error messages regarding what went wrong.
```
dreyfus92, line 114:
```suggestion
The possibilities for code improvement are endless. By adopting our approach of simplifying the code and making it more prone to panics, we can achieve the best of both worlds—code that's easy to follow and provides users with straightforward error messages explaining what went wrong.
```

.github/CONTRIBUTING.md
```diff
@@ -96,6 +96,33 @@ Not yet created, will be added in the future.
...
+
+```rust
+// Close splashscreen
+window.get_window("splashscreen").expect("no window labeled 'splashscreen' found").close().unwrap();
+// Show main window
+window.get_window("main").expect("no window labeled 'main' found").show().unwrap();
+```
+
+This simplified version is no beauty and we hope nobody actually uses it in their production environment, but it gets the point across and helps users debug issues by themselves. Error handling is something we will want to write longer blog posts regarding as it's a rather big topic to cover, but showing off an API however is best done in as few lines as possible with as expressive panics as possible.
```
dreyfus92, line 123:
```suggestion
This simplified version may not be the epitome of elegance, and we hope users won't actually implement it in their production environments. However, it effectively conveys the point and empowers users to diagnose and debug issues on their own. While error handling is a substantial topic that deserves more in-depth coverage, showcasing an API is best achieved with as few lines as possible, coupled with expressive panic messages.
```

dreyfus92 (changes requested):
Awesome job @simonhyll, I've left you a couple of suggestions, let me know what you think 😁

PR #1579, opened 2023-09-29
feat: add additional info about signing builds and C libraries

mihirsamdarshi:

- New or updated content

#### Description

I added additional documentation about cross-platform that I didn't see. I thought this information would at least be useful to people who attempt to go down this rabbit hole.

About the signing addition, it wasn't working, so I did some code-spelunking [eventually found this](https://github.com/tauri-apps/tauri/blob/1bce7397a496ca0cecf67fb5e3ec69d90fc871b0/tooling/bundler/src/bundle/windows/nsis.rs#L175). I thought that it should be documented.

About the C library compilation addition, there were a number of environment variables I had to set in my Github runner in order to get this working. I had to cross-compile `libsqlite3-sys` and `ring` and found that the following worked for me, but wasn't sure if I should document it, so I added what I did.

```shell
xwin --accept-license --cache-dir ~/.xwin-cache/ splat --output ~/.xwin/
sudo ln -s $(which clang) /usr/bin/clang-cl
# this file contains all the environment variables necessary for cross-compilation of Tauri
{
    echo CC=/usr/bin/clang-cl
    echo CXX=/usr/bin/clang-cl
    echo TARGET_CC=/usr/bin/clang-cl
    echo TARGET_CXX=/usr/bin/clang-cl
    echo TARGET_AR=/usr/bin/llvm-link
    echo CC_x86_64_pc_windows_msvc=/usr/bin/clang-cl
    echo AR_x86_64_pc_windows_msvc=/usr/bin/llvm-lib
    echo CXX_x86_64_pc_windows_msvc=/usr/bin/clang-cl
    echo CFLAGS_x86_64_pc_windows_msvc="--target=x86_64-pc-windows-msvc -Wno-unused-command-line-argument -fuse-ld=lld-link /imsvc$HOME/.xwin/crt/include /imsvc$HOME/.xwin/sdk/include/ucrt /imsvc$HOME/.xwin/sdk/include/um /imsvc$HOME/.xwin/sdk/include/shared"
    echo CXXFLAGS_x86_64_pc_windows_msvc="--target=x86_64-pc-windows-msvc -Wno-unused-command-line-argument -fuse-ld=lld-link /imsvc$HOME/.xwin/crt/include /imsvc$HOME/.xwin/sdk/include/ucrt /imsvc$HOME/.xwin/sdk/include/um /imsvc$HOME/.xwin/sdk/include/shared"
    echo BINDGEN_EXTRA_CLANG_ARGS_x86_64_pc_windows_msvc="-I$HOME/.xwin/crt/include -I$HOME/.xwin/sdk/include/ucrt -I$HOME/.xwin/sdk/include/um -I$HOME/.xwin/sdk/include/shared"
    echo RCFLAGS="-I$HOME/.xwin/crt/include -I$HOME/.xwin/sdk/include/ucrt -I$HOME/.xwin/sdk/include/um -I$HOME/.xwin/sdk/include/shared"
    echo CARGO_TARGET_X86_64_PC_WINDOWS_MSVC_LINKER=/usr/bin/lld-link
} >> $GITHUB_ENV

cat <<EOF >> $HOME/.cargo/config.toml                 
[target.x86_64-pc-windows-msvc]
linker = "lld"
rustflags = [
  "-C",
  "linker-flavor=lld-link",
  "-Lnative=$HOME/.xwin/crt/lib/x86_64",
  "-Lnative=$HOME/.xwin/sdk/lib/um/x86_64",
  "-Lnative=$HOME/.xwin/sdk/lib/ucrt/x86_64"
]
EOF
```

### DIFF
--- a/docs/guides/building/cross-platform.md
+++ b/docs/guides/building/cross-platform.md
@@ -153,7 +153,7 @@ Make sure to check the [documentation for GitHub Actions][github actions] to und
 
 ## Experimental: Build Windows apps on Linux and macOS
 
-Tauri v1.3 added a new Windows installer type based on the [NSIS] installer framework. In contrast to WiX, NSIS itself can also work on Linux and macOS which makes it possible to build many Tauri apps on non-Windows hosts. Note that this is currently considered highly experimental and may not work on every system or for every project. Therefore it should only be used as a last resort if local VMs or CI solutions like GitHub Actions don't work for you.
+Tauri v1.3 added a new Windows installer type based on the [NSIS] installer framework. In contrast to WiX, NSIS itself can also work on Linux and macOS which makes it possible to build many Tauri apps on non-Windows hosts. Note that this is currently considered highly experimental and may not work on every system or for every project. Therefore it should only be used as a last resort if local VMs or CI solutions like GitHub Actions don't work for you. **Note that, at this time, signing cross-platform builds is currently unsupported.**
 
 Since Tauri officially only supports the MSVC Windows target, the setup is a bit more involved.
 
@@ -250,6 +250,15 @@ rustflags = [
 Keep in mind that this file is specific to your machine so we don't recommend checking it into git if your project is public or will be shared with anyone.
 
 #### Building the App
+
+:::note
+
+If your application has dependencies to C libraries such as `ring` or `libsqlite3-sys`, building your application cross-platform gets a little trickier, 
+as you will also need to set up the appropriate environment variables and packages to cross-compile those C libraries on your system. These may include 
+setting the `CC`, `CXX`, `AR` and other environment variables, although this depends heavily on the set up of your build environment. Please refer to 
+the documentation of the libraries you are using for more information about any additional configuration for cross-compilation.
+
+:::
 
 Now it should be as simple as adding the target to the `tauri build` command:
 


lorenzolewis:
@tauri-apps/wg-tauri Could we please get a technical review for this one?

FabianLars:
That's probably me then.

OP's remarks are/should be correct (back then i only tested quite simple tauri apps when i wrote the guide etc), but i think just telling the reader to set these env vars without showing what values they should have doesn't help anyone. So i guess more testing is required to see what's required and when, or if it's always save to just do it all.

Also, and that's the reason i ghosted this PR, maybe we should look at cargo-xwin again which handles this oob. I had problems with it in the past so i went with the barebones solution (especially since i didn't really have what's discussed here in mind) but i've seen others semi-successfully using it in their tauri app.

lorenzolewis:
The changes in the PR itself don't really specify setting specific env vars, more as a heads-up that you _may_ need to under specific conditions. So I think in that case this is fine, right? I have a few comments on the content itself but I'll put that in an actual review.

docs/guides/building/cross-platform.md
```diff
@@ -153,7 +153,7 @@ Make sure to check the [documentation for GitHub Actions][github actions] to und
 
 ## Experimental: Build Windows apps on Linux and macOS
 
-Tauri v1.3 added a new Windows installer type based on the [NSIS] installer framework. In contrast to WiX, NSIS itself can also work on Linux and macOS which makes it possible to build many Tauri apps on non-Windows hosts. Note that this is currently considered highly experimental and may not work on every system or for every project. Therefore it should only be used as a last resort if local VMs or CI solutions like GitHub Actions don't work for you.
+Tauri v1.3 added a new Windows installer type based on the [NSIS] installer framework. In contrast to WiX, NSIS itself can also work on Linux and macOS which makes it possible to build many Tauri apps on non-Windows hosts. Note that this is currently considered highly experimental and may not work on every system or for every project. Therefore it should only be used as a last resort if local VMs or CI solutions like GitHub Actions don't work for you. **Note that, at this time, signing cross-platform builds is currently unsupported**
```
lorenzolewis:
```suggestion
Tauri v1.3 added a new Windows installer type based on the [NSIS] installer framework. In contrast to WiX, NSIS itself can also work on Linux and macOS which makes it possible to build many Tauri apps on non-Windows hosts. Note that this is currently considered highly experimental and may not work on every system or for every project. Therefore it should only be used as a last resort if local VMs or CI solutions like GitHub Actions don't work for you. **Note that, at this time, signing cross-platform builds is currently unsupported.**
```

docs/guides/building/cross-platform.md
```diff
@@ -251,6 +251,10 @@ Keep in mind that this file is specific to your machine so we don't recommend ch
 
 #### Building the App
 
+If your application has dependencies has bindings to C libraries such as `ring` or `libsqlite3-sys`, building your application cross-platform gets a little trickier, as you will also need to set up the 
+appropriate environment variables and packages on your system. These may include setting the `CC`, `CXX`, `AR` and other environment variables, although this depend heavily on the set up of your build 
+environment
+
```
lorenzolewis:
```suggestion
:::note

If your application has dependencies to C libraries such as `ring` or `libsqlite3-sys`, building your application cross-platform gets a little trickier, as you will also need to set up the appropriate environment variables and packages on your system. These may include setting the `CC`, `CXX`, `AR` and other environment variables, although this depend heavily on the set up of your build environment.

:::
```

docs/guides/building/cross-platform.md
```diff
@@ -251,6 +251,10 @@ Keep in mind that this file is specific to your machine so we don't recommend ch
 
 #### Building the App
 
+If your application has dependencies has bindings to C libraries such as `ring` or `libsqlite3-sys`, building your application cross-platform gets a little trickier, as you will also need to set up the
```
lorenzolewis: If we don't have a specific thing we're telling the reader to do (such as set this env var value to "xyz") then I like to provide a link to a reference the reader could find that information at (or at least a place to learn more information). Maybe we could simply say "please consult the documentation for those libraries to learn more about any additional configuration required"

lorenzolewis (changes requested):
Just a few adjustments on formatting and maybe an extra piece of context we could add in.

FabianLars:
> The changes in the PR itself don't really specify setting specific env vars, more as a heads-up that you _may_ need to under specific conditions. So I think in that case this is fine, right? I have a few comments on the content itself but I'll put that in an actual review.

True, but in non cross platform contexts you don't have to set any of them normally so I doubt many users will know what to do tbh. But in your review comment you already addressed that ig by asking for a link or something.

mihirsamdarshi:
> Also, and that's the reason i ghosted this PR, maybe we should look at cargo-xwin again which handles this oob. I had problems with it in the past so i went with the barebones solution (especially since i didn't really have what's discussed here in mind) but i've seen others semi-successfully using it in their tauri app.

Actually, `cargo-xwin` was actually extremely helpful to me in debugging what environment variables needed to be set. I would build my application with `tauri build`, and this would fail, but the compile would succeed (or fail later) when I would just do `cargo xwin build --manifest-path src-tauri/Cargo.toml`

docs/guides/building/cross-platform.md
mihirsamdarshi: So, I wanted to give links to specific GitHub issues that I found while searching for it. The problem is that this kind of thing is truly highly variable on the crates that a user is compiling, and there isn't a good single reference for cross-compiling on Rust.

I added your suggestion regarding the last comment

lorenzolewis (approved):
Looks good! Thanks again for putting this in, @mihirsamdarshi !

PR #1591, opened 2023-10-04
Fix wrong variable name and adjust path

Phippe:

#### Description

When I followed this guide the last few times, I needed to change the variable name from `$env:WINDOWS_PFX_PASSWORD` to just `$WINDOWS_PFX_PASSWORD`, as defined in the step before. It's been a while, but I assume this is still the case. Additionally, I removed the `Certs` directory from the file path, since in step A2, we've already changed the current directory to `Documents/Certs`.


FabianLars (approved):
THank you!

Issue #1475, opened 2023-09-03
Create Clipboard guide


lorenzolewis:

lorenzolewis:
This is partially resolved with #1604 , but we still need to get some Rust examples on the page to completely finish this one up

lorenzolewis:
Closed by #1622

PR #1604, opened 2023-10-09
Create clipboard guide #1475

vasfvitor:
- New or updated content
#### Description
- Address #1475 but not fully 

The content is based on the [README.md](https://github.com/tauri-apps/plugins-workspace/tree/v2/plugins/clipboard-manager) from the plugin's repo but I'm not sure about the use of `assert`, shouldn't be `console.assert`?

Still have to write the Rust part.

I'm looking at writing the other plugins guides, is it enough to adapt the content from each plugin's README.md*? I could just copy over and adapt. But my intention is to first try to use locally when possible.

\* Most stubs give v2 plugin README as a reference to get started, and at least [windows customization](https://beta.tauri.app/features/window-customization/) points to [v1 docs](https://tauri.app/v1/guides/features/window-customization/).

Since I'm learning Tauri as I write these docs, I don't want to claim this current issue, for me there's no problem if someone else does.

### DIFF
--- a/src/content/docs/features/clipboard.mdx
+++ b/src/content/docs/features/clipboard.mdx
@@ -5,10 +5,98 @@ description: Read and write to the system clipboard.
 
 import Stub from '@components/Stub.astro';
 import PluginLinks from '@components/PluginLinks.astro';
+import { Tabs, TabItem } from '@astrojs/starlight/components';
+import CommandTabs from '@components/CommandTabs.astro';
 
 <PluginLinks plugin="clipboard-manager" />
 
-<Stub>
-	Based on
-	https://github.com/tauri-apps/plugins-workspace/tree/v2/plugins/clipboard-manager
-</Stub>
+Read and write to the system clipboard using the clipboard plugin.
+
+## Setup
+
+Install the clipboard plugin to get started.
+
+<Tabs>
+    <TabItem label="Automatic">
+
+    1. Use your project's package manager to add the dependency:
+
+    <CommandTabs npm="npm run tauri plugin add clipboard-manager"
+    yarn="yarn run tauri plugin add clipboard-manager"
+    pnpm="pnpm tauri plugin add clipboard-manager"
+    cargo="cargo tauri plugin add clipboard-manager" />
+
+    2. Modify `lib.rs` to initialize the plugin:
+
+    {/* TODO: Revise once https://github.com/tauri-apps/tauri/issues/7696 is in */}
+
+    ```rust
+    #[cfg_attr(mobile, tauri::mobile_entry_point)]
+    pub fn run() {
+        tauri::Builder::default()
+            // Initialize the plugin
+            .plugin(tauri_plugin_clipboard_manager::init())
+            .run(tauri::generate_context!())
+            .expect("error while running tauri application");
+    }
+    ```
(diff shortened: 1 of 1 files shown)


vasfvitor:
The overall structure is based on the Notification plugin docs, and in there it has:

{/* TODO: Demo component */}

and 

{/* TODO: Link to which language to use, frontend vs. backend guide when it's made */}

That I copied over this current guide, but I don't know what they refer to. Is the aim that each plugin have a guide and a demo component for every (not every) JS and Rust frameworks? Or just vanilla?

src/content/docs/features/clipboard.mdx
```diff
@@ -5,10 +5,98 @@ description: Read and write to the system clipboard.
...
+<TabItem label="JavaScript">
+
+```js
+import { writeText, readText } from '@tauri-apps/plugin-clipboard-manager';
+
+// Write text to the clipboard
+await writeText('Tauri is awesome!');
+
+// Read text from the clipboard
+assert(await readText(), 'Tauri is awesome!');
```
lorenzolewis:
```suggestion
const content = await readText()
console.log(content)
// Prints "Tauri is awesome!" to the console
```

I wonder if something like this could be more familiar to JS devs instead of assert?

src/content/docs/features/clipboard.mdx
```diff
@@ -5,10 +5,98 @@ description: Read and write to the system clipboard.
...
+    <CommandTabs
+        npm="npm install @tauri-apps/plugin-clipboard-manager"
+        yarn="yarn add @tauri-apps/plugin-clipboard-manager"
+        pnpm="pnpm add @tauri-apps/plugin-clipboard-manager"
+    />
+
+    </TabItem>
+
+</Tabs>
+
```
lorenzolewis, line 74: Is the allowlist config that's covered in https://tauri.app/v1/api/js/clipboard/ still required? If not then we could also add that to our migration guide in https://beta.tauri.app/guides/upgrade-migrate/from-tauri-1/#tauri-configuration to say it's been removed as well.

lorenzolewis (changes requested):
Awesome contribution @vasfvitor, thanks for putting this in! Just a few pieces of feedback and I think this one looks good to get merged in 💪

src/content/docs/features/clipboard.mdx
vasfvitor: The allowlist isn't required anymore. It's already covered in the docs v2 you refered.
(image from the migration link)
image

lorenzolewis: Good catch, I hadn't used it yet so wasn't 100% sure, thank you!

lorenzolewis (approved):

lorenzolewis:
Thanks @vasfvitor ! To answer your original questions:

> I'm looking at writing the other plugins guides, is it enough to adapt the content from each plugin's README.md*? I could just copy over and adapt. But my intention is to first try to use locally when possible.

I think this is a really good start, and then note any gaps (like the Rust part you did here) so we can keep track of them. That does cause some duplication between the READMEs and what we have in the docs, but I don't think that's an issue as we start to treat the docs as the main source for references and documentation.

PR #1622, opened 2023-10-16
feat: clipboard rust guide #1475

vasfvitor:
- New or updated content

#### Description

- Related to #1475
- There's some wording and code choices regarding Rust syntax. To facilitate I wrote the 'Read' in one way and 'Write' part in another way. To clarify, examples:
1. 

With `...` to indicate that it's possible to generate multiple handles:
```rust
    .invoke_handler(tauri::generate_handler![write])
```

or without:
```rust
    .invoke_handler(tauri::generate_handler![write])
```
By the way, I don't know if there any difference between calling `invoke_handle `multiple times vs only one.

2.
Use `match`
```rust
  let content = app.clipboard().read();
    match content {
        Ok(content) => println!("Clipboard:{:?}", content),
        Err(error) => println!("{:?}", error),
    }
```

or `unwrap()`
```rust
    let content = app.clipboard().read();
    println!("{:?}", content.unwrap());
}
```

Then another situation: The JavaScript part seems too crude compared to Rust. We could Improve by adding the whole function in JS. I had to add function signature in Rust because of the handle part.

Another issue is regarding how to call those functions from Rust frontend. Afaik there's Yew, Leptos and Sycamore to Rust frontend and each one would have a different way to write. The vanilla option is html/css/js. 
Because of this I'm thinking in a different approach in the last step: "3. Call it from your frontend -> Tabs (Rust | JavaScript)" with links for the frameworks specific guides on how to call function from Tauri/Rust backend.

I was reading some past pr discussions but if there's anything relevant to read I appreciate. Thanks

Note: I'm still learning Rust.

### DIFF
--- a/src/content/docs/features/clipboard.mdx
+++ b/src/content/docs/features/clipboard.mdx
@@ -96,9 +96,87 @@ console.log(content);
 </TabItem>
 <TabItem label="Rust">
 
-{/* TODO: */}
+{/* WRITE TO CLIPBOARD RUST */}
 
-<Stub />
+#### Write text to the clipboard:
+
+1. Set up the function in `main.rs` (Tauri backend)
+
+```rust
+use tauri_plugin_clipboard_manager::ClipboardExt;
+
+#[tauri::command]
+async fn write(content: String, app: tauri::AppHandle) {
+    // Copy content to the clipboard
+    let clipboard_content = tauri_plugin_clipboard_manager::ClipKind::PlainText {
+        label: Some("Label".to_string()),
+        text: content,
+    };
+    let result = app.clipboard().write(clipboard_content);
+    match result {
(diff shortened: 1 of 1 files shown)


lorenzolewis:
For the Rust part, I don't think that `invoke_handler` and bridging the function over to the JS side is needed. The user would instead just use the corresponding JS function available. Let's assume that they just want to read/write to/from the clipboard in Rust code and then they can do what they'd like from there (like we do on the JS tab).

I think that with that change in mind this could be condensed down to just one Rust code block very similar to the JS side. The goal is to really just show the basics of the function to get the user up and running with as minimal but functional code as possible.

vasfvitor:
It now matches the JS guide, anything else?
image

image

lorenzolewis (approved):
Looks good! Thank you very much for this!

PR #1623, opened 2023-10-16
feat: create JS and Rust dialog guide 

vasfvitor:

- New or updated content

#### Description

- ~~Partially~~ Addresses #1477 
-~~I'm still unsure what we should include or exclude from the guide. I mean, there are more options to each function that I didn't specify but it's written in the JS api docs.~~
- Fixed typo on notification guide.
~~- This only includes JavaScript~~

### DIFF
--- a/src/content/docs/features/dialog.mdx
+++ b/src/content/docs/features/dialog.mdx
@@ -208,17 +208,19 @@ console.log(file);
 
 ### Save to File Dialog
 
+Open a file/directory save dialog.
+
 <Tabs>
 <TabItem label="JavaScript">
 
 ```js
 import { save } from '@tauri-apps/plugin-dialog';
 
-// Prompt to save a 'Image' with extension .png or .jpeg
+// Prompt to save a 'My Filter Label' with extension .png or .jpeg
 const path = await save({
 	filters: [
 		{
-			name: 'Image',
+			name: 'My Filter Label',
 			extensions: ['png', 'jpeg'],
 		},
 	],
--- a/src/content/docs/features/dialog.mdx
+++ b/src/content/docs/features/dialog.mdx
@@ -86,6 +86,8 @@ in Rust:
 ***
 ## JavaScript
 
+See all [Dialog Options](/2/reference/js/dialog/) at the JavaScript API reference.
+
 {/* ASK */}
 ### Create Yes/No Dialog
 
@@ -126,7 +128,7 @@ console.log(confirmation);
 
 ### Create Message Dialog
 
-Shows a message dialog with an `Ok` button.
+Shows a message dialog with an `Ok` button. Keep in mind that if the user closes the dialog it will return `false`.
 
 ```js
 import { message } from '@tauri-apps/plugin-dialog';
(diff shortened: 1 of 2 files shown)


lorenzolewis:
Thanks for this! Overall really like the simplicity of this. For the headings could we use them to describe _what_ the function is doing more clearly? For example: instead of "Ask" as a heading, we could instead have "Create Yes/No Dialog". This way the reader is able to quickly look at the TOC and jump to the desired result they'd like.

vasfvitor:
Ok. Done

src/content/docs/features/dialog.mdx
```diff
@@ -5,10 +5,315 @@ description: Native system dialogs for opening and saving files along with messa
...
+Shows a question dialog with `Yes` and `No` buttons
+
+<Tabs>
+<TabItem label="JavaScript">
+
+```js
+import { ask } from '@tauri-apps/plugin-dialog';
+
+// Create a Yes/No dialog
+const yes = await ask('This action cannot be reverted. Are you sure?', {
```
lorenzolewis: I would rename this to something like `answer` since it could be either true or false and then update what's logged below.

src/content/docs/features/dialog.mdx
```diff
@@ -5,10 +5,315 @@ description: Native system dialogs for opening and saving files along with messa
...
+<Tabs>
+
+Select single file:
+
+<TabItem label="JavaScript">
+
+```js
+// Open a dialog
+const file = await open({
+	multiple: false,
```
lorenzolewis, line 156: Instead of listing out each permutation of `multiple` and `directory` I think just adding a paragraph before the code block explaining what `multiple` and `directory` mean would be just as effective and reduce a lot of duplication.

lorenzolewis (changes requested):
Awesome! Just a few adjustments. Would you like to take a stab at including Rust in this as well or would you like to leave that as a follow-up PR for you or someone else?

src/content/docs/features/dialog.mdx
vasfvitor: Ok, I reduced the duplication. Now, since there are multiple options, should those be explained as well? Not only the Open dialog, but Save, Message, and Confirm function, each have multiple options. 

Current Open dialog (The link to js api is inside the JavaScript tab)


To compare with Save dialog

vasfvitor:
I'll write the Rust guide

src/content/docs/features/dialog.mdx
lorenzolewis: I think how you've done in the "Open" one is perfect by linking to the API references. This lets the reader understand the very basic functionality (which is what this page is for) and let's them jump off to customise and configure it from there (which is what the reference page is for).

vasfvitor: I separated Rust and JavaScript guides into 2 blocks and moved stuff around. I hope that is ok. My reasoning is that this plugin have different ways to be used that does not translate into 1:1 blocks.

Also moved the link of Rust docs and JavaScript docs to the start of the block. Since it should be basic functionality I guess, content-wise, ready.

lorenzolewis:
I'm good with splitting them out when it's not 1:1. I would just increase all of the headings under "Usage" by one so that the JS and Rust headings (and their specific children) sit nicely under "Usage". You could also add `i18nReady: true` to the frontmatter now that this is written to signal that it's available for translators.

vasfvitor:
Done and done. Way better with the headings this way

PR #1635, opened 2023-10-19
[v2] fix: shikiConfig

### LINKED ISSUE #1634: [v2] Code syntax highlight - Build log: The language "..." doesn"t exist, falling back to plaintext.
### 📋 Explain your issue

Before: 
https://deploy-preview-1620--tauri-docs-starlight.netlify.app/2/reference/js/core/
image

Current version (no highlight):
image

Deploy log at `astro build`

#1621  Oct 16 at 10:35 AM:
https://app.netlify.com/sites/tauri-docs-starlight/deploys/652d3c0fc8fa140008ab46b9

#1620 Oct 16 at 9:56 AM:

Affects: It seems that only .md pages. Not .mdx. So, all js api reference and some other .md pages that have code blocks. 

I think I already have the solution, will open a PR

vasfvitor:

- Closes #1634, tested locally with tauri-docs and with a starlight starter repo.

While this fixes the issue, I'm not sure this is the real solution. This seems related to the recent change from shiki to shikiji in Astro. From what I gather shikiji needs langs to be specified. On the other hand if `shikiConfig` is removed, then it seems to work too.
```
shikiConfig: {
			langs: ['powershell', 'ts', 'rust', 'bash', 'json', 'toml'],
		},
```
The added items is based on the warnings

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -194,7 +194,7 @@ export default defineConfig({
 	],
 	markdown: {
 		shikiConfig: {
-			langs: ['powershell'],
+			langs: ['powershell', 'ts', 'rust', 'bash', 'json', 'toml'],
 		},
 		rehypePlugins: [
 			rehypeHeadingIds,


FabianLars (approved):
thank you!

PR #1637, opened 2023-10-21
create-tauri-app: add --alpha flag when using cargo command

tkanakamalla:

- Minor fix in the documentation

#### Description

While creating a project using `create-tauri-app`, the commands need the `--alpha` as parameter to get V2 features

### DIFF
--- a/src/content/docs/guides/create/index.mdx
+++ b/src/content/docs/guides/create/index.mdx
@@ -27,33 +27,33 @@ To get started using `create-tauri-app` run one of the below commands in the fol
 <Tabs>
 <TabItem label="Bash">
 ```sh
-sh <(curl https://create.tauri.app/sh)
+sh <(curl https://create.tauri.app/sh) tauri-app --alpha
 ```
 </TabItem>
 <TabItem label="PowerShell">
 ```ps
-irm https://create.tauri.app/ps | iex
+$env:CTA_ARGS="--alpha";irm https://create.tauri.app/ps | iex
 ```
 </TabItem>
 <TabItem label="npm">
 ```sh
-npm create tauri-app@latest
+npm create tauri-app@latest -- --alpha
 ```
 </TabItem>
 <TabItem label="Yarn">
 ```sh
-yarn create tauri-app
+yarn create tauri-app --alpha
 ```
 </TabItem>
 <TabItem label="pnpm">
 ```sh
-pnpm create tauri-app
+pnpm create tauri-app --alpha
 ```
 </TabItem>
 <TabItem label="Cargo">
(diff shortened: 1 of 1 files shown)


FabianLars:
Hey, thanks for opening this PR! The --alpha flag is also required on the other methods (npm, yarn, etc) so i'd appreciate it if you can add them too :)

tkanakamalla:
Hi @FabianLars thanks for the recommendation. I have made the corresponding change for the yarn command.
I have tried below for npm 
`npm create tauri-app@latest --alpha` 
But this does not seem to be the right way so I did not work. I am not an expert with npm or pnpm so could you please suggest what would be the right way to add `alpha` flag to the npm, pnpm and the rest. I can then push another update with that

vasfvitor:
hey, with npm you need to add an extra `--`
`npm create tauri-app@latest -- --alpha`

and pnpm is just like yarn:
`pnpm create tauri-app --alpha`

but bash and powershell I don't know

tkanakamalla:
Thanks @vasfvitor  . I figured for bash and powershell as well. Updated the commit now for all commands.

src/content/docs/guides/create/index.mdx
```diff
@@ -27,33 +27,33 @@ To get started using `create-tauri-app` run one of the below commands in the fol
 <Tabs>
 <TabItem label="Bash">
 ```sh
-sh <(curl https://create.tauri.app/sh)
+sh <(curl https://create.tauri.app/sh) tauri-app --alpha
```
FabianLars:
```suggestion
sh <(curl https://create.tauri.app/sh) --alpha
```

Just this works for me too, can you confirm that it does for you as well?

src/content/docs/guides/create/index.mdx
```diff
@@ -27,33 +27,33 @@ To get started using `create-tauri-app` run one of the below commands in the fol
...
 ```
 </TabItem>
 <TabItem label="PowerShell">
 ```ps
-irm https://create.tauri.app/ps | iex
+$env:CTA_ARGS="--alpha";irm https://create.tauri.app/ps | iex
```
FabianLars:
```suggestion
$env:CTA_ARGS="--alpha"; irm https://create.tauri.app/ps | iex
```

FabianLars (approved):

src/content/docs/guides/create/index.mdx
tkanakamalla: Yeah right, it works without adding the project name in the args. I don't know what I have missed previously. Thanks for pushing the patch for this

Issue #1484, opened 2023-09-03
Create OS Information guide


lorenzolewis:

lorenzolewis:
Closed by #1642

PR #1642, opened 2023-10-23
feat: create OS Information guide #1484

vasfvitor:
- New content

#### Description
I'm putting this out but I'm not sure what the Rust code, or even the JS should included, I mean, how much detail or step. Right now the Rust code is minimal, so is the JS, it's almost like the js-api reference. For comparison in the clipboard guide #1622, I've included three steps to use the Rust code, with the end goal of invoking in frontend.

edit: ok, I'm readin https://diataxis.fr/

### DIFF
--- a/src/content/docs/features/os-info.mdx
+++ b/src/content/docs/features/os-info.mdx
@@ -103,5 +103,3 @@ println!("Platform: {}", platform);
 
 </TabItem>
 </Tabs>
-
-See all available functions in the [JavaScript API](/2/reference/js/os/) or [Rust API](https://docs.rs/tauri-plugin-os/2.0.0-alpha.4/tauri_plugin_os/) references.
--- a/src/content/docs/features/os-info.mdx
+++ b/src/content/docs/features/os-info.mdx
@@ -73,7 +73,7 @@ Install the OS Information plugin to get started.
 
 ## Usage
 
-With this plugin you can query multiple information from current operational system. See all available functions in the [JavaScript API](/2/reference/js/os/) or [Rust API](https://docs.rs/tauri-plugin-os/2.0.0-alpha.4/tauri_plugin_os/) references.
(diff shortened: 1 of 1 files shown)


lorenzolewis:
To your point on this being a JS/Rust API reference I agree. I think that instead of listing out each different function, we could instead just have on function (such as `platform`) and then add a sentence saying additional functions as available with links to the API docs for each one.

vasfvitor:
Is that how you imagined?

src/content/docs/features/os-info.mdx
```diff
@@ -3,11 +3,105 @@ title: OS Information
...
+        pnpm="pnpm add @tauri-apps/plugin-os"
+    />
+
+    </TabItem>
+
+</Tabs>
+
+## Usage
+
+With this plugin you can query multiple information from current operational system, here is how you can utilize the OS Information plugin:
```
lorenzolewis:
```suggestion
With this plugin you can query multiple information from current operational system. See all available functions in the [JavaScript API](/2/reference/js/os/) or [Rust API](https://docs.rs/tauri-plugin-os/2.0.0-alpha.4/tauri_plugin_os/) references.
```

src/content/docs/features/os-info.mdx
```diff
@@ -3,11 +3,105 @@ title: OS Information
...
+```rust
+let platform = tauri_plugin_os::platform();
+println!("Platform: {}", platform);
+// Prints "windows" to the terminal
+```
+
+</TabItem>
+</Tabs>
+
+See all available functions in the [JavaScript API](/2/reference/js/os/) or [Rust API](https://docs.rs/tauri-plugin-os/2.0.0-alpha.4/tauri_plugin_os/) references.
```
lorenzolewis:
```suggestion
```

lorenzolewis (changes requested):
Would just move this link to be further up at the top so the reader can more easily see it.

lorenzolewis (approved):
Awesome! And good catch with the docs.rs link

PR #1648, opened 2023-10-25
Benchmark charts: use straight strokes for readability

olivierlemasle:

#### Description
For its [benchmarks page](https://tauri.app/v1/references/benchmarks/), Tauri uses ApexCharts with [`curve` strokes](https://apexcharts.com/docs/options/stroke/). They are nice, BUT when when two points are very close in time, it sometimes creates an additional stroke, which makes the chart unreadable. This PR changes to `straight` strokes, which removes the issue.

Before:

image

After:

image

The whole chart, with **straight** strokes:

image


FabianLars (approved):
thank you :)

PR #1649, opened 2023-10-26
GitHub Sponsors and Open Collective contribution graph

lorenzolewis:

#### Description

This is a draft POC to demo out pulling in GitHub sponsors and Open Collective contributions to highlight the awesome people who sponsor Tauri!

Expecting the build to fail due to env vars not being set up.


vasfvitor:
@lorenzolewis I want to continue this. Could you brief me about what you were envisioning?

lorenzolewis:
Sure thing @vasfvitor ! I'm not sure if there have been any discussion around sponsorship in general in the last few WG calls as I haven't been in them, so take all of this below with a grain of salt:

---

The idea was to surface all people who sponsor (via Open Collective or GitHub Sponsors) and not just the ones who reach out to put their logo on the site. This way we can be more inclusive, fair, and also automate this more if we ping the respective APIs.

# Open Collective

For Open Collective you're able to see the total over time an entity (person or company) has contributed. The results would be sorted based on that. There are a few different ideas of how to organise this:

1. Based on different "tiers" (gold, platinum, silver, etc.) with pre-defined amounts for each
2. Sorted based on total historical contributions

I was mocking these out in this PR to show WG and get an idea of what is preferred, so that would need to be discussed I imagine. 

# GitHub Sponsors

For GitHub Sponsors we aren't able to see how much a certain person donates, so we would show them all (yes, it is possible to do that with an access token with the correct permissions, but this should only be using publicly available data to avoid exposing sensitive information). 

Note that you will need an access key to query the GitHub GraphQL API (even for public data), so just make sure that is set up appropriately.

I would place all GitHub Sponsors below the Open Collective sponsors since we're not able to determine the amount donated.

# Non-Monetary Donations

Some companies donate to us via a service (service provider like Netlify). I would put those into their own separate section/heading outside of the purely financial donations. 

Astro does so in this way on their index page:

Screenshot 2024-03-03 at 12 43 54

I feel trying to sift them into different "tiers" (if that approach is what is wanted to be followed) just puts us in a weird situation where we have to put a monetary value on a service and would be nicer to treat them separately since it is a different kind of contribution.

# Code Contributions

Lastly it would be really nice to show the people who contribute via code (see Astro again at https://docs.astro.build/en/getting-started/ with our very own @dreyfus92 there!):

This would probably make since as a separate PR but wanted to capture the thought here :)

Screenshot 2024-03-03 at 12 46 05


---

Hopefully that helps a bit. It's not a purely technical implementation that just needs to be coded, there's a bit of trial and error and then asking for feedback from WG (which is why it got stalled on my part).

Of course, if you have any questions along the way then feel free to reach out and ask! 💟

vasfvitor:
@lorenzolewis thank you very much for writing this up

> Hopefully that helps a bit

This helps a LOT!

> It's not a purely technical implementation that just needs to be coded, there's a bit of trial and error and then asking for feedback from WG (which is why it got stalled on my part).

I see

vasfvitor:
- [x] add title back and hide only on home
- [x] make logos responsive

.github/workflows/fetchSponsorsData.yml
```diff
@@ -0,0 +1,50 @@
+name: 'Sync Sponsors and Contributors Data'
```
vasfvitor, line 1: the data is commited to the repo and updated weekly, maybe bi-weekly would be enough. Only did this to avoid having to add a gh_token elsewhere

src/styles/custom.scss
```diff
@@ -33,158 +35,10 @@
   color: var(--sl-color-white) !important;
 }
```
vasfvitor, line 37: moved to another file, welp... it should be the red diff on custom.scss

vasfvitor (commented):
components only read the raw data, all fetching happens on fetch-sponsors. For now contributors is fetched but not used

.github/workflows/fetchSponsorsData.yml
vasfvitor: I haven't tested the action and not sure the tauri-bot setup is correct

FabianLars (approved):

PR #1688, opened 2023-11-18
feat: add `/releases` page

amrbashir:

- New or updated content

#### Description

Add a release notes page.


lorenzolewis:
I'm really curious if we need this when the same information is on GitHub. What's the objective you're wanting from this? Just a more focused and filtered list of releases to highlight the more important/popular packages?

amrbashir:
That and Github releases page UX is awful and doesn't support quick navigation to a certain release (unless you know exactly what release you want)

lorenzolewis:
Hmm, that's a good point. I also wonder if change logs for the plugins would be nice to have? That could be included on the respective plugin pages.  At first I was a bit torn at having the change details inline and running into formatting issues, but it's all markdown so that shouldn't be an issue. 

On the implementation, I'd prefer polling the GitHub API during build and statically have the changelog on pages instead of dynamically on each page load. That'd reduce overall page load time for the viewer and make it feel snappier (plus makes it so we have less client-side JS to ship).

I'd like to play around with the UI a little bit, but what are your thoughts on the above before that?

amrbashir:
> Hmm, that's a good point. I also wonder if change logs for the plugins would be nice to have? That could be included on the respective plugin pages. At first I was a bit torn at having the change details inline and running into formatting issues, but it's all markdown so that shouldn't be an issue.

Nice one, but inlining is probably not the way to go, just to not pollute documentation, I think all releases just be under `/releases`, because ideally we want to support `tauri.app/releases/tauri/1.0.2` (which I just thought of now, so I will convert this PR to a draft until I get some time to get back to ti)

> On the implementation, I'd prefer polling the GitHub API during build and statically have the changelog on pages instead of dynamically on each page load. That'd reduce overall page load time for the viewer and make it feel snappier (plus makes it so we have less client-side JS to ship)

That particular page doesn't need to have the best performance but yeah that would probably be better as it will make implementing `tauri.app/releases/tauri/1.0.2` easier.
 
> I'd like to play around with the UI a little bit, but what are your thoughts on the above before that?

Definitely feel free to do so, I don't very much like the styling I did

lorenzolewis:
Cool. If we do that then I can also just update some of the plugin meta links on the pages to point to a changelog page. Give me a shout if you want me to take a look at anything in the meantime.

amrbashir:
Shifted this PR to generate the releases pages on build, should be ready for review.

amrbashir:
I don't know why build fails with links validation tbh, @lorenzolewis do you have any idea why this happens?

lorenzolewis:
I believe because the links are formatted like `/releases/@tauri-apps/cli/v2_0_0-alpha_18` but astro with change those to `/releases/tauri-apps/cli/v2_0_0-alpha_18` to get rid of any special characters (note the `@` is removed). They use https://www.npmjs.com/package/github-slugger under the hood (or a custom impl that gets the same output) so I'd use that to generate any links (and possibly any filenames as well).

src/content/docs/releases/index.mdx
```diff
@@ -0,0 +1,12 @@
...
+template: splash
+---
+
+- [`tauri`](/releases/tauri)
+- [`@tauri-apps/api`](/releases/tauri)
+- [`tauri-cli`](/releases/tauri)
+- [`@tauri-apps/cli`](/releases/tauri)
+- [`tauri-bundler`](/releases/tauri)
+- [`wry`](/releases/tauri)
+- [`tao`](/releases/tauri)
```
lorenzolewis: I think these links need to be updated, right?

amrbashir: nice catch

amrbashir:
> I believe because the links are formatted like `/releases/@tauri-apps/cli/v2_0_0-alpha_18` but astro with change those to `/releases/tauri-apps/cli/v2_0_0-alpha_18` to get rid of any special characters (note the `@` is removed). They use [npmjs.com/package/github-slugger](https://www.npmjs.com/package/github-slugger) under the hood (or a custom impl that gets the same output) so I'd use that to generate any links (and possibly any filenames as well).

Can we override this behavior somehow? ideally the link should contain the `@` just fine, similar to https://paka.dev/npm/@tauri-apps/api

lorenzolewis:
Hmm, tbh I don't see anything in Starlight that would affect this, and generating a normal Astro page seems to retain special characters. We're using a slightly old version due to a few dependency incompatibilities so let me check in the morning if I can figure out whats happening.

lorenzolewis:
Actually, it may have to do with Content Collections because they do a tad bit of magic sauce on top of the default route handling. We can try this and see if it works (away from computer right now): https://docs.astro.build/en/guides/content-collections/#defining-custom-slugs

amrbashir:
> Actually, it may have to do with Content Collections because they do a tad bit of magic sauce on top of the default route handling. We can try this and see if it works (away from computer right now): [docs.astro.build/en/guides/content-collections/#defining-custom-slugs](https://docs.astro.build/en/guides/content-collections/#defining-custom-slugs)

It didn't work unfortunately, the error is coming from this plugin though https://github.com/HiDeoo/starlight-links-validator

lorenzolewis:
I'm thinking this is caused by this regex upstream: https://github.com/HiDeoo/starlight-links-validator/blob/e4d378d7484995b50a9a6f5f22a098bdd597cf70/packages/starlight-links-validator/libs/remark.ts#L146

We're not on the latest version due to version incompatibilities with the blog plugin. There are a few changes in this area of the link validator in later versions. Let me message HiDeoo and see what he recommends since he maintains both plugins. I'm assuming the blog plugin isn't going to be updated immediately so I may just quickly throw something together to fix that and clear that blocker. 

Will reply back here with the feedback.

lorenzolewis:
Most likely will be after the holiday season as the author is indeed wanting to push through https://github.com/withastro/starlight/pull/1175 in order to unlock the other issues (and thus letting us upgrade all of our deps). I'm thinking to work through a few other things on my list and then just rolling a custom blog impl since Astro's Content Collections make it fairly trivial tbh (and can let us add RSS as well).

vasfvitor:
seeing if/how we can unblock this from upstream

amrbashir:
thanks @vasfvitor, glad to see you have resolved the issues, I hadn't have enough time work on this lately but let me know if I can be of any help.

vasfvitor:
@amrbashir I was just writing this up😄. What do you need to be done to finish this one? I can get it going, just point me generally

image

amrbashir:
Haha, I guess "great minds think alike"

I don't think I have more to add to this PR atm so feel free to merge or add anything if you feel there is something missing.

vasfvitor:
Added navigation: next-prev buttons, and "return to: index"




"All releases"



Card links @simonhyll idea


still some stuff todo

simonhyll:
If we can get the sidebar to show on all pages I'd say we can merge this and think of improvements later.

Shows here: https://deploy-preview-1688--tauri-docs-starlight.netlify.app/releases/
Doesn't show here: https://deploy-preview-1688--tauri-docs-starlight.netlify.app/releases/tauri/v1.2.2/

vasfvitor:
@simonhyll to account for this then I changed the design so the list of releases is enclosed in a dropdown like menu, as if it's mobile. Not quite good looking imo, but gtg I'd say

simonhyll (approved):
While I would want the right sidebar to show up in a manner more similar to the existing Starlight right sidebar, I think what we have currently is good enough to merge. We can toy around with the exact appearance down the line.

lucasfernog (approved):
I agree with Simon but it's tricky to generate a custom table of contents view, so let's go with this for now.

lucasfernog (approved):
I agree with Simon but it's tricky to generate a custom table of contents view, so let's go with this for now.

PR #1703, opened 2023-11-26
docs: add pkg-config env variable example

wralith:
- Add example command to export environment variable required for "Build Conflict with Homebrew on Linux" section

#### Description
Docs states set the env variable, but doesn't explains what to do very well.
https://github.com/tauri-apps/tauri/issues/2798#issuecomment-1089822619 I thought adding this to the docs will be nice for new users who uses brew on linux.


FabianLars (approved):
thanks!

PR #1713, opened 2023-11-27
Feat: Rebuild Configuration Generator

### LINKED ISSUE #1472: Fix Tauri Config Generator
The generator that builds https://beta.tauri.app/2/reference/config/ is currently broken. Would like to switch over to something like https://github.com/adobe/jsonschema2md or similar so that we don't need to maintain the current generator at https://github.com/tauri-apps/tauri-docs/blob/next/src/plugins/configGenerator.ts

lorenzolewis:

- Something else!

#### Description

- Closes #1472
- Move Tauri config builder to a separate package

#### Feedback Items

- Would we prefer to render each of the items under "Definitions" on their own page to reduce overall length?
- What's the proper licensing, headers, etc. I should be putting on files? (I think I need to update this across a few things in this repo)

---

Once this is in then I'll do a separate PR to shift over the v1 generator to this.


packages/config-generator/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "version": "1.0.0",
+  "description": "",
+  "main": "index.js",
+  "type": "module",
+  "scripts": {
+    "build": "tsm ./build.ts"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
amrbashir: We use MIT or Apache-2.0

packages/config-generator/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "config-generator",
+  "version": "1.0.0",
```
amrbashir, line 3: since this is private package and it not meant to be published, it is better to remove unnecessary fields and set `"private": true`

amrbashir (commented):
I don't think it is beneficial to review the code now so instead, I will focus on the rendered structure.

I think we can skip the type in the root object because this doesn't feel nice 
image

Some code blocks are not rendered correctly 
image

Headers need a bit more size variance, using h2 followed by h3, doesn't  give enough hierarchy IMO
image

The array types are rendered incorrectly 
image
it should be rendered:
```md
`string[] | null`
```

The Object structure is a little weird to me:
   - I don't see a value in listing the properties without any info and then listing them again individually, instead we could remove the `Properties` heading and instead append the type before the property
   ```md
   ## DebianConfig
   ... type
   ... description
   #### DebianConfig.depends
   ... type
   ... description
   #### DebianConfig.desktopTemplate
   ... type
   ... description
   and so on
   ```
   
   - The type could be rendered like this to give it more visibility:
   ```md
   ## DebianConfig
   - type: `object`
   ... description here
   ... then list properties
   ### depends
   - type: `string[] | null`
   ... property description here
   ```
   - I don't like these sub definitions, it also renders invalid headers like `########` which exceeds the limit, it also has an index in its name, so instead we should just try to inline the definition

amrbashir:
If you don't understand any of my ramblings we can pair together sometime to go over these changes if you want.

lorenzolewis:
@amrbashir Thanks, makes sense and I can use that to revise a bit. I tried several different iterations and designs on the sub-definitions and this was the best formatting I could come up with (without going into custom CSS and HTML territory). Do you have any thoughts off the top of your head for a revision?

amrbashir:
I don't have any revisions atm, but I don't think we should be limited with what markdown supports, we should be allowed to use HTML/CSS too

lorenzolewis:
Put in a few of the changes in place. I'll catch up with you async to discuss a few other things in detail.

vasfvitor:
Recently I noticed at least 2 PRs targeting auto generated files, so I remembered that over Astro Docs they add a "DO NOT MAKE EDITS" warning to auto-generate pages.

To illustrate:
This page https://www.github.com/withastro/docs/blob/main/src/content/docs/en/guides/integrations-guide/alpinejs.mdx
is generated from https://www.github.com/withastro/docs/blob/main/scripts/generate-integration-pages.ts

`<DontEditWarning/>` component: https://www.github.com/withastro/docs/blob/main/src/components/DontEditWarning.astro renders this in dev mode

My initial suggestion is to just place a text warning on top of page

lorenzolewis:
Good call @vasfvitor , I just put a quick note in the generator for it. 

@amrbashir if we'd like to go beyond the power of just markdown for this then I'd love to double-down on that and make some sort of tree-view UI where you can expand/collapse the nodes and that will solve some of the awkward hierarchy issues you highlighted.

If we do that then I'd still like to at least get a fix in place for the current impl since it's a bit broken: https://beta.tauri.app/2/reference/config/

amrbashir:
> make some sort of tree-view UI where you can expand/collapse the nodes 

This sounds to me like the old tauri config page (prior stable) which were not very practical in terms of linking to a specific option.

> If we do that then I'd still like to at least get a fix in place for the current impl since it's a bit broken: https://beta.tauri.app/2/reference/config/

I don't mind pushing this PR to fix this page and do another iteration on it in another PR.

lorenzolewis:
Cool. I updated a bit to address some of your feedback but left a few things alone (such as the sub-definition feedback) since I'd prefer to tackle those in the future PR. Does this look good to merge for the time being @amrbashir ?

amrbashir (approved):
yeah let's merge this and improve on it later

PR #1716, opened 2023-11-28
Update index.mdx

bhark:

#### Description

Updated the default Linux installation path, and separated it from macOS, as they are not the same. 
The SDK folder is, by default with a capital S ("Sdk"), and the Android folder is by default directly in $HOME.


FabianLars (approved):

FabianLars:
thank you :)

PR #1670, opened 2023-11-08
v2 migration: `guides/development` to next

vasfvitor:
#### What kind of changes does this PR include?

- New or updated content
- Changes to the docs site code

#### Description

Pages: 
https://tauri.app/v1/guides/development/development-cycle
https://tauri.app/v1/guides/development/updating-dependencies

- Added redirects
- Diff at https://github.com/tauri-apps/tauri-docs/commit/3b7adc3ba9da1fc80e6e3f98575bb2958692b9b0
- Sidebar updated according to https://starlight.astro.build/guides/sidebar/#autogenerated-groups
- Modified the card below from /development-cycle: bold text, caution box and code filenames/path:

v2 (next):

Original from v1:

sidebar preview:

#### Personal Notes (can be ignored)

`import Tabs from '@theme/Tabs';`
`import TabItem from '@theme/TabItem';`

can be replaced by `import { Tabs, TabItem } from '@astrojs/starlight/components';` to place code, if needed
or  `import CommandTabs from '@components/CommandTabs.astro';` to place commands

This `import Command from '@theme/Command'`
to `import CommandTabs from '@components/CommandTabs.astro';`

The Command
`<Command name="dev" />
`
becomes
```
<CommandTabs
	npm="npm run tauri dev"
	yarn="yarn tauri dev"
	pnpm="pnpm tauri dev"
	cargo="cargo tauri dev"
/>
```

https://starlight.astro.build/guides/authoring-content/#note-aside
`
:::info title`
into 
`:::caution[title]`

`:::info` can be: `note`, `tip`, `caution` or `danger` depending on context.

PR #1717, opened 2023-11-29
feat: Create `Directory.astro` component

vasfvitor:

#### Description
This is updated so It can be both be used in the current situation and in the migration list or elsewhere.
For reference: https://github.com/tauri-apps/tauri-docs/pull/1670#issuecomment-1831786609  

Usage:
```html
<Features path="features" locale="en" />
```

I chose for now to make `locale` mandatory so it's clear that it is needed on translated pages. With a bit of more work it could be removed by matching `path` to `locales.json` but first I want to put this out here to gather feedback on which direction should this go.

Currently, on i18n pages, first is rendered translated cards and then fallback pages, excluding translated. Another possibility is to leave them mixed, in which _Notifications_ card would be down below.

"(en)" is for clarity, but I do think something to indicate it's fallback is nice to have.

Preview:
```html
<Features path="fr/guides/develop/" locale="fr" />
```

Another use case:
```html
<Features path="/guides/develop/" locale="en" />
```

I think before merging, it should be decided this will be the same component in the aforementioned issue comment or change it's name.

Thank you.


lorenzolewis:
Have a few thoughts on this but going to put them in Discord for us to chat through when you're online: https://discord.com/channels/616186924390023171/662624589037436928/1179737597925392434

lorenzolewis:
I wonder if this should also respect `order` in https://starlight.astro.build/reference/frontmatter/#sidebarconfig?

vasfvitor:
> I wonder if this should also respect `order` in https://starlight.astro.build/reference/frontmatter/#sidebarconfig?

good point, yes

vasfvitor:
DO NOT MERGE

Updates: 
- Removed `locales` props, whether `path` it's localized or not, the component forces current page locale.
- Sort following sidebar order, with option to override it

Anything else?

src/components/list/Directory.astro
```diff
@@ -0,0 +1,145 @@
...
+	 */
+	path: string;
+	/**
+	 *  Filter out pages by keyword - case insensitive
+	 */
+	filterOut?: string[];
+	/**
+	 * Force alphabetical order
+	 */
+	forceAbc?: boolean;
```
lorenzolewis: Could we rename this to `sortAlphabetically` or similar to be a bit more self-descriptive?

src/components/list/Directory.astro
```diff
@@ -0,0 +1,145 @@
...
+const rootUrl = locale !== defaultLocale ? path.slice(locale.length + 1) : path;
+
+/**
+ * Collections structure:
+ * @const rootList contains english pages
+ * @const i18nList contains chosen locale pages
+ * @const fallbackList is subtraction of both
+ * Final composition is @var list then and @const fallbackList
+ */
+const rootList = await getCollection('docs', ({ slug }) =>
```
lorenzolewis: Are there any helper functions in Starlight that do this same thing we could rely on? I'm not too familiar with this area of upstream 🤔

lorenzolewis (commented):
Just put a few general feedback comments. Overall pretty happy with this. There's a few moving parts upstream with the Badge in a LinkCard, but happy to merge this when you're ready and just refactor when those land.

src/components/list/Directory.astro
vasfvitor: Well, I looked into it and found something that I wanted to try and this led me to refactor haha, now it utilizes `const routes` exported from Astro itself, not sure it's better,. But give me your feedback. If it's ok then it's ready to merge. I removed all the demos and unrelated changes.

lorenzolewis: Looks good to me, will put a few comments separately.

lorenzolewis:
Happy with this overall. I think the final piece is putting it in place on the respective pages in this PR so we can double-check that we're happy with the results.

vasfvitor:
had to add a filter by filename, how do you feel of having both options(`filterOutByFileName` and `filterOutByTitle`)? Was thinking of making a more complex filter but there's no need yet, I'd say.

lorenzolewis:
I would do `slug` to roughly match what Content Collections can do: https://docs.astro.build/en/guides/content-collections/#filtering-collection-queries

vasfvitor:
done, anything else?

lorenzolewis:
Looks good. I think the `references` usages need to be updated with the latest in `next` as those URLs were changed with #1762

lorenzolewis:
Awesome! Let's get this merged in 💪

PR #1718, opened 2023-11-30
Add blogs back

lorenzolewis:
This adds back in the blog posts to the render pipeline. This is so we can have accurate metrics for comparing performance. This does not affect redirects.

#### Description

- Closes # 


FabianLars:
does this mean we have to add new blog posts to both branches?

lorenzolewis:
@FabianLars nope, just to `next`. I just wanted to throw these back in to get a bit closer to actual metrics when I measure build times in the future, but these effectively won't be used.

FabianLars (approved):

PR #1725, opened 2023-12-03
feat: Create plugin http guide 

### LINKED ISSUE #1480: Create HTTP Client guide

vasfvitor:
- New or updated content

#### Description
- Closes #1480
- Bit of context about Rust part https://discord.com/channels/616186924390023171/662624589037436928/1177657907131727952

What about the line highlight? For this and other guides
Current:
`/\\.plugin.*$/`

Another ways:
`{6}`

`ins=/\\.plugin.*$/`

`ins={6}`

there may be more ways, not sure, reference: https://starlight.astro.build/guides/authoring-content/#expressive-code-features


lorenzolewis:
I prefer using line numbers to highlight I think. That's what's most common and I've seen in other places so I think it will be easier for contributors to understand (especially since what's being highlighted is a full line and not just a word or two).

src/content/docs/features/http-client.mdx
```diff
@@ -3,11 +3,125 @@ title: HTTP Client
...
+</Tabs>
+
+## Usage
+
+The http plugin is available in both as an JavaScript API and in Rust as a [reqwest](https://docs.rs/reqwest/) re-export.
+
+
+### JavaScript
+
+#### 1. Configure allowed scope
```
lorenzolewis: Instead of these being headings, let's keep them as normal text (similar to how we format the steps under the Setup heading).

src/content/docs/features/http-client.mdx
```diff
@@ -3,11 +3,125 @@ title: HTTP Client
...
+
+```json
```
lorenzolewis, line 84: Could we add a link to the respective reference docs for the config?

src/content/docs/features/http-client.mdx
```diff
@@ -3,11 +3,125 @@ title: HTTP Client
...
+  method: "GET",
+});
+console.log(response.status); // e.g. 200
+console.log(response.statusText); // e.g. "OK"
+
+```
+
+:::note
+
+ The current `fetch` method is an API to Rust backend. It tries to be as close and compliant to the `fetch` Web API as possible.
```
lorenzolewis:
```suggestion
The current `fetch` method is an API to Rust backend. It tries to be as close and compliant to the `fetch` Web API as possible.
```

lorenzolewis (changes requested):
Just a few changes here then it's good to go!

src/content/docs/features/http-client.mdx
vasfvitor: There's https://beta.tauri.app/2/reference/config/#pluginconfig that points to v1 but afaik this config is v2 specific.

I only know that the scope config is needed because it's commented on the source code, then I tested it. And if I tested correctly (not sure this is obvious) but the scope only applies to js code.

Another possible link is https://beta.tauri.app/2/reference/config/#fsscope but I haven't tested "allow" and "deny". This week I'll test with them to make sure it works and use this link if that's matches what you were referring to

vasfvitor: so I've tested and looked on the source code and "scope" only allows an array. So I've added a link to JS api reference, is it ok? https://github.com/tauri-apps/tauri-docs/pull/1725/commits/9ea203b7afd1638562ee5030436c2123a40d855c

lorenzolewis: Yes, that looks good to me. This area I think is still being actively worked on so we'll keep an eye on it for any future changes as well.

src/content/docs/features/http-client.mdx
```diff
@@ -3,11 +3,125 @@ title: HTTP Client
...
+  method: "GET",
+});
+console.log(response.status); // e.g. 200
+console.log(response.statusText); // e.g. "OK"
+
+```
+
+:::note
+
+The current `fetch` method is an API to Rust backend. It tries to be as close and compliant to the `fetch` Web API as possible.
```
lorenzolewis:
```suggestion
The current `fetch` method is an API to Rust backend. It tries to be as close and compliant to the [`fetch` Web API](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API) as possible.
```

lorenzolewis:
Thanks @vasfvitor ! Just added a link to the Fetch API and the rest is good to go!

PR #1730, opened 2023-12-05
Add info for the `additionalBrowserArgs` property

Alaricus:
- New or updated content

#### Description
- This provides additional information for the `additionalBrowserArgs` property. Specifically, it mentions the fact that in case of applications with multiple windows, this property needs to be added to all windows. Otherwise, the built app does not execute.

### DIFF
--- a/docs/api/config.md
+++ b/docs/api/config.md
@@ -29,7 +29,7 @@ The configuration is composed of the following objects:
 - [`build`](#buildconfig): The build configuration
 - [`plugins`](#pluginconfig): The plugins config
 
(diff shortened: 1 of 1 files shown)


FabianLars:
Hey 👋 Thanks for contributing!

The config page is autogenerated and changes must be added in the upstream tauri repo here https://github.com/tauri-apps/tauri/blob/dev/core/tauri-utils/src/config.rs#L1124-L1125

To be completely honest though, i don't think this change adds _that_ much value because the new note is true for every single window config property, not just additionalBrowserArgs. Thank you nonetheless!!

PR #1731, opened 2023-12-05
v2 migration: `guides/building` to next

vasfvitor:
~There's no need to review now, just need a feedback on the last item on TODO list (Revise Copy)~

- New or updated content

#### Description
Diff at: https://github.com/tauri-apps/tauri-docs/commit/f101f882b0e9dc0b0f9a785ef48c779d626980e5
The Checklist' TOC changes on `app-size.md` was made automagically so I decided to keep them if that's ok

Pages: 
https://tauri.app/v1/guides/building/
https://tauri.app/v1/guides/building/windows
https://tauri.app/v1/guides/building/macos
https://tauri.app/v1/guides/building/linux
https://tauri.app/v1/guides/building/cross-platform
https://tauri.app/v1/guides/building/app-size
https://tauri.app/v1/guides/building/sidecar
https://tauri.app/v1/guides/building/resources

TODO:
- [x] Redirects
- [ ] Update/Improve index page (card positions)
- [x] Fix links and navigation
- [x] i18n pages, if any
- [ ] Revise changes
- [ ] Revise Copy? - mostly because it mentions Allowlist and this has changed, plus, there may be other changes.

Now I'm wondering, if we change those, how will it be for folks utilizing v1?

~~I was thinking of transforming `_tauri-build.mdx` into a fragment as `cta.mdx` but leave only tabs, then text contents will be placed on each two page with <TauriBuild/>~~ I guess it's better to keep it as is just to avoid text duplication and possible different translations


src/content/docs/guides/build/cross-platform.mdx
```diff
@@ -0,0 +1,285 @@
...
+	yarn="yarn build --target x86_64-pc-windows-msvc"
+	pnpm="pnpm build --target x86_64-pc-windows-msvc"
+	cargo="cargo build --target x86_64-pc-windows-msvc"
+/>
+
+The build output will then be in `target/x86_64-pc-windows-msvc/release/bundle/nsis/`.
+
+{/* TODO: Check distribute links once those are in */}
+[windows code signing with github actions]: /guides/distribute/sign-windows#bonus-sign-your-application-with-github-actions
+[macos code signing with github actions]: /guides/distribute/sign-macos#example
```
vasfvitor: links breaking the build, from `/guides/distrubute` to be migrated

lorenzolewis: Let's point to prod for now (https://tauri.app/foo...) to get this through and then we can update later.

vasfvitor: so, I was looking at the pages that was changed over and turns out if I understand correctly from the Discord thread, it's the Code Signing pages. Now we can link them here in /build

lorenzolewis:
@vasfvitor is this one ready for a full review now?

vasfvitor:
@lorenzolewis yeah it is

vasfvitor:
consider #1961 changes

tweidinger:
Imho I think we only need to port app size from v1 the rest seems to be already live in the current v2 branch. Can you double check with the current v2 branch @vasfvitor what could/would be transferred from this PR?

vasfvitor:
> Can you double check with the current v2 branch @vasfvitor what could/would be transferred from this PR?

sure, I'll do over the weekend

PR #1738, opened 2023-12-07
chore(deps): Update astro to v4

FabianLars:
Just wanna check if it fails to build here too as it does locally for me. Couldn't figure out the issue yet :/

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,15 +18,15 @@
     "preview": "astro preview"
   },
   "dependencies": {
-    "@astrojs/markdown-remark": "^3.5.0",
-    "@astrojs/starlight": "^0.13.0",
+    "@astrojs/markdown-remark": "^4.0.0",
+    "@astrojs/starlight": "^0.15.0",
     "@types/json-schema": "^7.0.15",
-    "astro": "^3.6.4",
+    "astro": "^4.0.3",
     "astro-feelback": "^0.3.4",
     "prettier": "^3.1.0",
     "prettier-plugin-astro": "^0.12.2",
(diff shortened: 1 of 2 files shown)


vasfvitor:
From what I gathered it's a compatibility issue between Starlight and starlight-blog plugin (I've tested and it is). It has to be first improved on Starlight then on plugin. The plugin's maintainer is already on it and opened a PR on Starlight

All details on Astro discord channel: https://discord.com/channels/830184174198718474/1070481941863878697/1181965518853394523
invite: https://astro.build/chat

lorenzolewis:
I'm thinking about re-creating that blog functionality after I get past a few pieces of the API generators. There's some UX improvements with the page layouts that we could get and it would also unlock this update. But we'll see who's faster: me or Starlight 😅

vasfvitor:
> I'm thinking about re-creating that blog functionality after I get past a few pieces of the API generators. There's some UX improvements with the page layouts that we could get and it would also unlock this update. But we'll see who's faster: me or Starlight 😅

If that's the case, wouldn't be better to just make a blog.tauri website? Speaking of which, is there a reason that the only website is the docs? Like, it could have a main page with blog, about and stuff, and then, docs in docs.tauri

lorenzolewis:
Potentially, but that also means you need to maintain the same things in those respective places. If we have it all under one repo then it makes it a lot easier to keep everything updated by only needing to monitor one place. I do like other places that have it separate but I feel for us it's a better to call to bundle docs, blog posts, and "about" pages under one roof to be as easy to maintain as possible.

simonhyll:
Resolved in #1858

PR #1740, opened 2023-12-08
feat: Features list search

vasfvitor:
- TODO done!

#### Description

https://github.com/tauri-apps/tauri-docs/blob/2ecf1bb8c22c1822356fa0dbd4422da47fabbc8d/src/content/docs/features/index.mdx#L14

This a POC that implement the search bar. It searches both title and description because it was easier to do this way.

I'll leave as draft to be discussed the shape of the component and such.

https://deploy-preview-1740--tauri-docs-starlight.netlify.app/features/
Quick preview: 

firefox_Lieysvxiux


lorenzolewis:
Awesome first draft of it. I wonder if it should also search the Community Resources? For styles, can we match what the search input has? It should roughly be the styles here (https://github.com/withastro/starlight/blob/5df5798405ed0a29f0a2b2ca9252cda6e28e72a3/packages/starlight/components/Search.astro#L139-L171) but it will need to be adapted for `input` instead of `button` I guess.

vasfvitor:
>Awesome first draft of it. I wonder if it should also search the Community Resources? 

I'd say they should be separated searches, if/when the Community Resources grows we could place another search there. I see they together if the cards had badges to make easier to identify which is which, for this we'll have to wait the badge PR on starlight.

>For styles, can we match what the search input has? 

Sure, lemme try

Update: I got the CSS to match search box of the modal open, so the icon is slightly different.

lorenzolewis:
A few UX/UI pieces of feedback:

- The placeholder font needs to be a few shades darker to match the search box UI. 

Screenshot 2023-12-20 at 10 59 22

- If you change the type to `type="search"` then you should automatically get a little `x` button to clear the search. 
- Pressing `enter` causes the page to reload (I think it counts as submitting a form). This will need to be ignored.
- When there are not results then the entry is just blank. Could we add some helper text to explain no results when that's the case?

vasfvitor:
I need to test this with the Directory Component https://github.com/tauri-apps/tauri-docs/pull/1717?notification_referrer_id=NT_kwDOA65hNbM4NTg5Njc2NzI5OjYxNzU5Nzk3#issuecomment-1864270630 so I think it's better to wait it. Actually I'll try to make it search within it's children. So we could just wrap anything with it.

You can take a look at the changes, I guess I addressed everything, have not tested on mobile tho. What about autocomplete?

vasfvitor:
while working on this I found out I was wrong on this below when I said it wasn't a good idea. Turn out after making changes to make possible to search within children components I found a way to keep the headings and search separated. Well haven't tested for performance with multiple (more than 2) children.

https://github.com/tauri-apps/tauri-docs/pull/1740#issuecomment-1854728901

Example with headings:
```tsx
<Search>
  ## Features
  <FeaturesList />
   ## Features 2
  <FeaturesList />
   ## Features 3 
  <FeaturesList />
   ## Features 4
  <FeaturesList />
  ## Community Resources
  <CommunityList />
</Search>
```


-

Example with some headings:

```tsx
<Search>
  ## Features
  <FeaturesList />
  <FeaturesList />
   ## Features 3 
  <FeaturesList />
   ## Features 4
  <FeaturesList />
  ## Community Resources
  <CommunityList />
</Search>
```


Now need to find a way to expose it to i18n

lorenzolewis:
You shouldn't need to include the close button or any special logic for the close button, the browser should handle that automatically I believe. Right now it's causing duplicate close buttons (Chromium)
Screenshot 2024-01-02 at 11 17 13

vasfvitor:
ohhh my, that's what happens when you (me) only test on firefox. Firefox (Gecko) layout for `type="search"` is a [experimental feature](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Experimental_features#layout_for_input_typesearch). 

It's possible to [implement it with vanilla css](https://codepen.io/shidhincr/pen/kaPPXN) but when I briefly tried, the js reduction was at best 2 lines.

For now `type="search"` is removed so the duplication issue is gone.

Plus I noticed that the placeholder was too dark on Chrome, that's fixed now.

Should I pursue this with vanilla CSS or is that ok?

lorenzolewis:
In general I try to use a non-JS method to do something if it exists. So to clear a form I think that would be `<input type="reset">` like you've already found in https://codepen.io/shidhincr/pen/kaPPXN. But tbh this seems like one of the areas where there isn't a super straightforward answer (and browsers aren't quite settled on the same way of handling the different options either), so I'm okay with this approach if you are.

I'm going to go through and give a bit of feedback on CSS styles just to try to match the theme as best as possible. Will put those inline in a review.

src/components/CardGridSearch.astro
```diff
@@ -0,0 +1,255 @@
...
+						<style>
+							.warning {
+								color: var(--sl-color-orange-high);
+								font-size: larger;
+							}
+							.hidden {
+								display: none;
+							}
+						</style>
+						<span>Nothing found in ${heading.textContent}</span>
```
lorenzolewis: I would style these the same as `<p>` tags, so having a color of `var(--sl-color-text)`. Could possibly make the text in italics just to set it apart a bit. 

Would also change the text to "No results found in [...]"

src/components/CardGridSearch.astro
```diff
@@ -0,0 +1,255 @@
...
+ * https://github.com/CloudCannon/pagefind/blob/main/pagefind_ui/default/svelte/ui.svelte
+ * And CSS variables from Starlight from:
+ * https://github.com/withastro/starlight/blob/a0f6fa80d9be1fbe85530323a0b2523acb83a15a/packages/starlight/components/Search.astro
+ */
+
+interface Props {
+	placeholder?: string;
+	title?: string;
+}
+const { placeholder = 'Quick search a feature...', title = 'Search a feature' } = Astro.props;
```
lorenzolewis:
```suggestion
const { placeholder = 'Search features and community resources...', title = 'Search a feature' } = Astro.props;
```

I imagine the placeholder and title could be the same value 🤔

src/components/CardGridSearch.astro
```diff
@@ -0,0 +1,255 @@
...
+			'San Francisco', 'Roboto', 'Segoe UI', 'Helvetica Neue', 'Lucida Grande', 'Ubuntu', 'arial',
+			sans-serif;
+	}
+	@media (min-width: 50rem) {
+		:root {
+			--sl-search-cancel-space: 0px;
+		}
+	}
+	.pagefind-ui {
+		width: 100%;
```
lorenzolewis, line 158: Could we make this span all the way across to match the width of any `LinkCards` below it on smaller viewports? See this screenshot for an example:


image

src/components/CardGridSearch.astro
```diff
@@ -0,0 +1,255 @@
...
+	placeholder?: string;
+	title?: string;
+}
+const { placeholder = 'Quick search a feature...', title = 'Search a feature' } = Astro.props;
+
+// TODO: Expose placeholder and title to i18n
+---
+
+<card-search>
+	<div id="search-input" class="pagefind-ui pagefind-ui--reset pagefind-ui--reset">
```
lorenzolewis: Duplicate css of `pagefind-ui--reset`, but this also doesn't seem to be defined below. You may be looking for something like [`.not-content`](https://starlight.astro.build/guides/components/#compatibility-with-starlights-styles)?
```suggestion
	<div id="search-input" class="pagefind-ui pagefind-ui--reset">
```

src/components/CardGridSearch.astro
```diff
@@ -0,0 +1,255 @@
...
+</script>
+
+<style>
+	:root {
+		/* Starlight */
+		--pagefind-ui-primary: var(--sl-color-accent-light);
+		--pagefind-ui-text: var(--sl-color-gray-2);
+		--pagefind-ui-font: var(--__sl-font);
+		--pagefind-ui-background: var(--sl-color-black);
+		--pagefind-ui-border: var(--sl-color-gray-5);
```
lorenzolewis: Since the logic for this component is 100% created by us then I think it makes sense to instead use the respective `--sl-*` vars directly instead of re-defining them with `--pagefind-ui-*` and then using that below again.

lorenzolewis (changes requested):
Mainly some nit-picky details on the CSS

src/components/CardGridSearch.astro
vasfvitor: done
image

src/components/CardGridSearch.astro
vasfvitor: `.pagefind-ui--reset` comes from https://github.com/CloudCannon/pagefind/blob/main/pagefind_ui/default/svelte/reset.svelte I can probably copy over some styles so we don't have this hidden dependency. What do you think? That's why I left pagefind-ui prefix in all classes

src/components/CardGridSearch.astro
vasfvitor: I think that those variables gives semantic meaning and allows easier customization, if needed. Anyways, I already made the change in https://github.com/tauri-apps/tauri-docs/pull/1740/commits/67d5bdca3b8efe355ad2053a76b63f271a8f0e0d

simonhyll (approved):

PR #1762, opened 2023-12-20
Add Tauri V1 References

lorenzolewis:

#### Description

- Change reference slugs from `/2/reference/...` to `/references/2/...` and update respective internal links
- Update submodules from `packages/tauri` to `packages/tauri-v2`
- Add `packages/tauri-v1` submodule used to generate v1 docs
- Update package scripts to now setup and generate JS and configs for Tauri v1 on build
- Refactor config generator to support generating multiple configurations with different entrypoints
- Cleanup JS API generator and add Tauri v1 to generation. Note that this is planned on being modified again in the future to support cleaner dependency upgrades.
- Add `references/1/cli` and `references/1/config`
- Update reference index page to now list V1 references


dreyfus92 (commented):
LGTM 🫡 just wondering, `/2/reference/...` the `2` stands for `v2`? if that's the case wouldn't be better to change it to `v2`?

lorenzolewis:
> LGTM 🫡 just wondering, `/2/reference/...` the `2` stands for `v2`? if that's the case wouldn't be better to change it to `v2`?

That's a good call. @FabianLars I think you'd mentioned having the "v" prefix before, right? I guess this PR would be good a time as any to take care of that. I can adjust it in the morning.

FabianLars:
Yes I did :) Looks a bit better/clearer to me but i didn't have any hard feelings about it

lorenzolewis:
Updated all the links and should be good to go! Thanks for the reviews!

PR #1769, opened 2023-12-24
chore: fix `zh-CN` language label

kaikaibenkai:

- Translated content

#### Description

China includes China Mainland and Special Administrative Regions (Hong Kong, Macao, Taiwan). China Mainland uses Simplified Chinese while the Special Administrative Regions use Traditional Chinese.

This means, Chinese (China) is divided into Simplified Chinese and Traditional Chinese.

`zh-CN` means Simplified Chinese here, as evident in translated text.

> [!Tip]
> Chinese (China) = 中文（中国）
> Simplified Chinese = 简体中文
> Traditional Chinese = 繁体中文


lorenzolewis:
Thanks for this! @dklassic could I get a thumbs-up on this one when you have a moment please?

dklassic:
> Thanks for this! @dklassic could I get a thumbs-up on this one when you have a moment please?

*Technically* language code `zh-cn` itself does literally mean "Chinese (China)", but it does makes sense that not all of China uses Simplified Chinese and many people outside of China also uses Simplified Chinese.

I wouldn't be against this change and it does make sense functionally.

lorenzolewis:
Also seems to be what Starlight is using as well; let's merge it!

Screenshot 2024-01-02 at 17 26 08

PR #1775, opened 2023-12-26
Update splashscreen.md to account for distDir recreation for Vite-based projects

bikathi:
I noticed that the documentation causes unwated behaviours with the splash screen especially for Vite-based projects(I use Tauri with Vue). Manually creating the `splashscreen.html` inside the `distDir` causes the file to be deleted each time you rebuild the project using `npm run tauri build`. This documentation PR should allow the reader to automate this manual creation by copying the file from the root into their `distDir`

### DIFF
--- a/docs/guides/features/splashscreen.md
+++ b/docs/guides/features/splashscreen.md
@@ -16,7 +16,7 @@ First, create a `splashscreen.html` in your `distDir` that contains the HTML cod
   },
 ```
 
-This will copy the `splashscreen.html` into the `distDir` folder whenever you run `npm run tauri build`. In previous documentation, this npm command usually deleted the `splashscreen.html` that you were manually creating in the `distDir`, as it recreates the `distDir` folder each time you run the command.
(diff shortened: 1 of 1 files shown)


FabianLars:
Hi, thanks for the PR! I think we can simplify this a bit by using vite's `public/` folder (next to package.json) which doesn't need any configuration - vite will copy those files into the distDir automatically. Pretty much all alternatives to vite have a similar mechanism.

Also the note about prior docs is something for the PR description, not the docs themselves so please remove that.

bikathi:
Hi @FabianLars. You're actually right. I have tested with you suggestion and it actually works fine without much effort. Should I update the PR to reflect your suggestion?

FabianLars:
yes please :)

bikathi:
Hi @FabianLars sorry for the late action. The PR is now updated.

FabianLars (approved):
Thank you :)

PR #1786, opened 2024-01-01
docs: Add an example for apple_signing_identity

FabianLars:
I forgot to open a PR before my vacation 👏


FabianLars:
i kinda hope we can reference the new Walter Tauri persona throughout the docs, idk why but i find it fun 🙃

PR #1794, opened 2024-01-08
docs: Disable outdated and unused benchmarks page

FabianLars:
*unused in the context of how it was meant to be used, for us a control mechanism.

Edit: iirc we won't add this to the new docs anyway, so may as well pull the plug already.


FabianLars:
It looks a bit funny with it being the only card without a learn more button but better than linking to a broken page i guess (it will only be used for a few more weeks/months anyway.

FabianLars:
Since this is easily revertable i'll quickly pull the trigger

PR #1803, opened 2024-01-12
Add warning to the relaunch function description

skyslide22:

- New or updated content

#### Description

Add a warning related to the relaunch function (not supported when app runs in dev mode)


FabianLars:
Hi, thanks for contributing! First of all, the build is failing because of the `<br>`, i think a `</br>` should work fine, not sure.
Secondly, this file is autogenerated on each release so you haev to make the changes here instead: https://github.com/tauri-apps/tauri/blob/1.x/tooling/api/src/process.ts#L37-L48
(i'd probably would remove the br tag altogether and use simple newlines)

Thanks again :)

PR #1812, opened 2024-01-18
fix dependency for linux on ubuntu22.04 and current taori version

Bennexy:
- New or updated content

#### Description
Fixed a wrong dependency. 
`cargo run --release --example helloworld` build failed because of an old dependency. 
Im running `v2.0.0-alpha.21`


FabianLars:
Hi, thanks for contributing! The _current_ version of tauri is still v1 so the docs are correct. We're working on docs for v2 (which is still in alpha) here https://beta.tauri.app/guides/prerequisites/ which already shows the new webkitgtk version.

Thanks again :)

PR #1844, opened 2024-02-08
Update sveltekit.mdx

NC-PB:
SvelteKit no longer re-exports vitePreprocess. You should import it directly from @sveltejs/vite-plugin-svelte.

#### Description

When copy the whole code block, the project won't build.
vitePreprocess is no longer exported from @sveltejs/kit/vite
That's why @sveltejs/vite-plugin-svelte should be used.


FabianLars (approved):
thank you :)

PR #1855, opened 2024-02-14
Update next-js.mdx 

przytrzask:

Minor change in a docs 

#### Description
Should be `pages` instead of `routes` I believe
CleanShot 2024-02-14 at 17 10 27@2x


FabianLars (approved):
thanks :)

PR #1874, opened 2024-02-24
chore(docs): update CTA --beta flag

samcday:
#### Description

Now that tauri v2 is in beta, invoking CTA with `--alpha` flag yields the following warning:

```
warning: The `--alpha` option is now an alias for `--beta` and may be removed in the future.
```


FabianLars (approved):
thank you :)

PR #1881, opened 2024-02-25
feat: Compatilibity component

vasfvitor:
# _OUTDATED_ read next comments
Opening this PR just so it can be seen live. I think this components could be like the "plugin links" that accepts the name of plugin as a prop and that's it. Example `<PluginLinks plugin="clipboard-manager" />`.

In this PR there are two versions of the compatibility component that implements Amr and Simon ideas, respectively:

1) Showing Linux distros inline:
Sample pages: `features/authenticator`, `features/autostart` and `features/barcode-scanner`
image

2) Showing Linux distros in a tooltip, if necessary (needs to make mobile accessible):
Sample pages: `features/cli` and `features/clipboard`
image
image

Aesthetically both can be improved, even more the tooltip by using JS (which I don't think should be done). Plus, neither version is supposed to be the definitive or the only possibles implementation of this feature

TODO: Take in account feedback sent by AlmostSenselessCoder in discord
https://discord.com/channels/616186924390023171/662624589037436928/1211237670110629978
- [ ] Screen readers (sr-only). Colours and hover alone creates an inaccessible or hard to access UI. It needs to expand on click too, not just on hover
- [ ] Click targets, see minimum recommended

TODO: Rename component
TODO: Consider icons


vasfvitor:
CanIuse sample json, if need something alike.
https://github.com/Fyrd/caniuse/blob/main/sample-data.json

For now I created a JSON with all the plugins and modified the component api to match the `PluginLinks.astro` api, so in the future we may consider combining them into a `PluginsHeader.astro` component

```tsx
<CompatList plugin="authenticator" />
```
So there's no need to make changer per page, one json to rule them all

vasfvitor:
Are we gonna use icons?

As of now I think it's ok to be reviewed. The JSON was made from the workspace/v2 table

About accessibility, is there any improvements that could be done?

src/components/CompatList.astro
```diff
@@ -0,0 +1,39 @@
...
+}
+// TODO: Find proper yellow color scheme with text, preferably with starlight variables
+
+let { plugin } = Astro.props;
+const platformList: PluginTable | undefined = compatTable.find((p) => p.name === plugin);
+const platforms = platformList ? platformList.platforms : [];
+---
+
+<div class="flex row not-content">
+	{platforms.map((platform) => <p>{platform}</p>)}
```
dreyfus92, line 20: I think that adding the icon next to the name/platform would be even better, what you guys think? @tauri-apps/wg-doc

PR #1890, opened 2024-02-27
feat: Plugin permissions component

### LINKED ISSUE #2006: (next) Some of the documentation for the fs plug-in is incorrect
`fs:allow-app-meta-recursive` has the same description as `fs:allow-app-meta`.

image

vasfvitor:
This is mostly to improve DX. Right now it just gets the generated md file and outputs it. Pending i18n functionality

> Preview:
> 
> ~~But right now it is like that below, until changes are implemented upstream, or ,alternatively, process the md file in the component. I'm already checking this out.~~
> 
> 
> ~For now this is pending tauri-plugin-workspace submodule update,~ ~but it's best to wait to update it until/if the generated markdown is changed in [Tauri core](https://github.com/tauri-apps/tauri/blob/e1d5b7906369a40df19e8ee86c56f90a27d6357c/core/tauri-utils/src/acl/build.rs#L234)~ https://github.com/tauri-apps/tauri/pull/9019

Usage:
```tsx
 <Permissions plugin="authenticator" />
```


tillmann-crabnebula:
Added the plugin identifier to the permission id in https://github.com/tauri-apps/tauri/pull/10113 so you should be able to use this somehow. Also open to improve the generated tables.

tillmann-crabnebula:
Hey @vasfvitor can you update the tauri and plugin workspace submodule (the plugins default permissions and the upstream PR are merged)?
Would like to see it in rendered preview :)

tillmann-crabnebula:
Also wondering how we want to translate the auto generated content and content based on the plugins permissions. Maybe we use some auto translation for these generated sections?

vasfvitor:
> can you update the tauri and plugin workspace submodule (the plugins default permissions and the upstream PR are merged)?

@tillmann-crabnebula it looks https://github.com/tauri-apps/plugins-workspace/blob/v2/plugins/autostart/permissions/autogenerated/reference.md broken? I'll update the submodules 

> Also wondering how we want to translate the auto generated content and content based on the plugins permissions. Maybe we use some auto translation for these generated sections?

this is something I've been wondering since. And somehow having it auto translated would be the best, because the strings are very repetitive and prone to change. And I think it should be translated because it's part of a 'normal' page where eveything else can be translated and not a full auto generated reference page

vasfvitor:
https://deploy-preview-1890--tauri-v2.netlify.app/plugin/global-shortcut/

vasfvitor:
@simonhyll for some reason it's failing on set_statuses, any ideas?

src/components/Permissions.astro
```diff
@@ -0,0 +1,45 @@
+---
+/**
+ * Usage:
+ * <Permissions plugin="authenticator" />
+ */
```
dreyfus92: I don't think that this is necessary cause the props are defined.

dreyfus92 (commented):
Really neat solution, I just left a small suggestion.

lucasfernog:
I didn't know about this one and opened #2522 just pick the one you prefer and close the other (or we could merge this one and apply my other changes idk)

lucasfernog:
I'm kinda leaning into yours now that I have to write some scope docs.. yours is more flexible :)

lucasfernog:
Something I wish we could have is to keep the headers in the "On this page" section :/ doesn't seem like it's possible since we're inlining HTML instead of markdown

lucasfernog:
fyi alternatively we could import the markdown files directly:
```
import { Content as PluginPermissions } from '../../../../packages/plugins-workspace/plugins/fs/permissions/autogenerated/reference.md';
```
but that still doesn't preserve the headers :(

vasfvitor:
> I didn't know about this one and opened https://github.com/tauri-apps/tauri-docs/pull/2522 just pick the one you prefer and close the other (or we could merge this one and apply my other changes idk)

np, I just looked at it and was thinking about using yours because you already provided more usage context. Notice that mine changes around 26 files and yours 17, so a few pages was changed without need, or are missing tables (I added to all pages with or without content)

vasfvitor:
and I always enjoy seeing how people solves the same problem that I did in a totally different way

lucasfernog:
i'll go ahead and merge this one, i wanna move a bit faster to write as much content as i can :D

PR #1873, opened 2024-02-23
feat: added Awesome Tauri to community resources

simonhyll:
#### What kind of changes does this PR include?

- Minor content fixes (broken links, typos, etc.)
- New or updated content
- Translated content
- Changes to the docs site code
- Something else!

#### Description

- Closes # 
- What does this PR change? Give us a brief description. 
- Did you change something visual? A before/after screenshot can be helpful.

PR #1900, opened 2024-02-29
Fix/search no results

simonhyll:
Based on my PR for Awesome Tauri.

https://github.com/tauri-apps/tauri-docs/pull/1873


vasfvitor:
When there's no results, the scrollbar disappears and cause the page to shift horizontally

simonhyll:
> When there's no results, the scrollbar disappears and cause the page to shift horizontally

Nothing can be done about that, the reason it disappears and the page shifts horizontally is because the page simply isn't long enough anymore to warrant one. You'd have to make more content for the page

vasfvitor:
For me having "have something to share" at the top make it looks odd. Also it is in different positions based on the result



---


I guess that ideally it should be fixed at the very bottom of the results.

Or maybe just remove this from the "Community Resource" section and put above the search bar on prime real state haha. Could be just a aside, just to give it a flair. Not sure, just thinking out written

PR #1911, opened 2024-03-04
fix: removed command tab frames

simonhyll:
Gets rid of frames around CommandTabs code blocks.

#### Description

- Closes # 


vasfvitor (approved):
If the intention is to remove sitewide, this will not suffice. Code blocks \``` with [specific langs](https://expressive-code.com/key-features/frames/#terminal-frames) will have frames by default. Not sure if there's a global option to disable it, but it's possible to disable by snippet

simonhyll:
The goal is just to go from this
image
To this
image
For now at least. It's an improvement, I'll merge it, we can think about other places to fix the frames in a followup PR.

PR #1924, opened 2024-03-05
docs(features): update `upload.mdx`

### LINKED ISSUE #1494: Create Upload guide

dreyfus92:

- New or updated content

#### Description

- Closes #1494 
Updates `/features/upload` page

### DIFF
--- a/src/content/docs/features/upload.mdx
+++ b/src/content/docs/features/upload.mdx
@@ -13,7 +13,7 @@ Upload files from disk to a remote server over HTTP. Download files from a remot
 
 ## Setup
 
-_This plugin requires a Rust version of at least **1.64**_
+_This plugin requires a Rust version of at least **1.75**_
 
 <Tabs>
 	<TabItem label="Automatic">
--- a/src/content/docs/features/upload.mdx
+++ b/src/content/docs/features/upload.mdx
@@ -18,44 +18,45 @@ _This plugin requires a Rust version of at least **1.64**_
 <Tabs>
 	<TabItem label="Automatic">
 
-	Use your project's package manager to add the dependency:
+    Use your project's package manager to add the dependency:
 
-	<CommandTabs
-	npm="npm run tauri add upload"
-	yarn="yarn run tauri add upload"
-	pnpm="pnpm tauri add upload"
-	cargo="cargo tauri add upload"
-	/>
+    <CommandTabs
+    npm="npm run tauri add upload"
+    yarn="yarn run tauri add upload"
+    pnpm="pnpm tauri add upload"
+    cargo="cargo tauri add upload"
+    />
 
-	</TabItem>
-	<TabItem label="Manual">
+    </TabItem>
+    <TabItem label="Manual">
 
-		<Steps>
+    	<Steps>
 
-		1. Run `cargo add tauri-plugin-upload` to add the plugin to the project's dependencies in `Cargo.toml`.
+    	1. Run `cargo add tauri-plugin-upload` to add the plugin to the project's dependencies in `Cargo.toml`.
 
-		2. Modify `main.rs` to initialize the plugin:
+    	2. Modify `main.rs` to initialize the plugin:
 
-			```rust title="main.rs" ins={4}
-			fn main() {
-				tauri::Builder::default()
-					// Initialize the plugin
-					.plugin(tauri_plugin_process::init())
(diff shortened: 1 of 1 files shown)


FabianLars:
ah damn, that's my bad. @amrbashir We still want to merge it into the http plugin right?

amrbashir:
I think it is fine to keep as is (or removed) since a combination of `http + fs` plugins can achieve the same thing?

I am a bit hesitated to merge into http plugin because we will need to handle scope for the filesystem as well and also upload plugin have a bit more dependencies that may not concern most users using http plugin.

FabianLars:
> a combination of http + fs plugins can achieve the same thing?

That'd be much slower because the data would have to go through the ipc. The v2 ipc may be faster, but not _that_ much faster :/

> also upload plugin have a bit more dependencies that may not concern most users using http plugin.

imo feature flags would be fine for that 🤷 

> because we will need to handle scope for the filesystem as well

Hmm, that's indeed annoying and a good reason...

amrbashir:
I guess better to be kept as a standalone plugin

FabianLars:
should we rename it to http-extra? 😂 (not actually meant as a joke, i don't like the upload name when it can download too)

amrbashir:
sure, whatever sounds right with you

dreyfus92:
@FabianLars, so should I close this?

FabianLars:
No, seems like we keep the plugin. If we indeed rename it then we can just change the docs 🤷

src/content/docs/features/upload.mdx
```diff
@@ -3,12 +3,87 @@ title: Upload
...
 
-<Stub>
-	Based on
-	https://github.com/tauri-apps/plugins-workspace/tree/v2/plugins/upload
-</Stub>
+Upload files from disk to a remote server over HTTP. Download files from a remote HTTP server to disk.
+
+## Setup
+
+_This plugin requires a Rust version of at least **1.64**_
```
vasfvitor: 1.75 (we need to find a better way to specify the plugin version)

src/content/docs/features/upload.mdx
```diff
@@ -3,12 +3,87 @@ title: Upload
...
+    pnpm="pnpm tauri add upload"
+    cargo="cargo tauri add upload"
+    />
+
+    </TabItem>
+    <TabItem label="Manual">
+
+    	<Steps>
+
+    	1. Run `cargo add tauri-plugin-upload` to add the plugin to the project's dependencies in `Cargo.toml`.
```
vasfvitor, line 35: I liked the steps, now we should update all the other guides to match this.

vasfvitor (commented):
I think it's ok, have you used the plugin yourself to make sure it works as intended and this code is reproducible?

src/content/docs/features/upload.mdx
dreyfus92: I was thinking about adding a note e.g.

:::note
This plugin requires a Rust version of at least **1.64**
:::

src/content/docs/features/upload.mdx
dreyfus92: yeah, I wanted to implement this one so everybody could see them in action. I liked the result tbh.

src/content/docs/features/upload.mdx
vasfvitor: think it's too much, I may have expressed myself badly, I was referring to the fact that we would have to manually check and update versions

dreyfus92: If we could unify the Rust version of all the plugins would be ideal tho.

vasfvitor: https://github.com/tauri-apps/plugins-workspace/blob/v2/README.md?plain=1#L34

So it seems they all share the same Rust version

vasfvitor:
```suggestion
_This plugin requires a Rust version of at least **1.75**_
```

vasfvitor (approved):

src/content/docs/features/upload.mdx
FabianLars: At least currently they do, yes. I just added it to all readmes so it's displayed correctly on crates.io/npm. If you think it's valuable we could keep them always matched (so raise msrv for all, if required by one), and/or matching it with tauri too.

PR #1936, opened 2024-03-09
fix: added `run` to npm scripts

ZibanPirate:

#### Description

The command snippets were missing the mandatory `run` command for `npm` to run `tauri` as a script.

before:
```sh
npm tauri [tauri-command]
```
after:
```sh
npm run tauri [tauri-command]
```

### DIFF
--- a/src/content/docs/references/v1/cli.mdx
+++ b/src/content/docs/references/v1/cli.mdx
@@ -8,7 +8,7 @@ import CommandTabs from '@components/CommandTabs.astro';
 ## `info`
 
 <CommandTabs
-	npm="npm tauri info"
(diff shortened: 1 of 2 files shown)


dklassic:
Thanks for noticing this issue! Though the page is auto-generated and we should be fixing this issue by patching up the generator instead or else it'll just be overwritten.

I've opened an issue #1985 for this problem for now.

FabianLars:
Re-opening because I was wrong, apologies.

dklassic (approved):
We still have move to auto generated solution in the future anyways to prevent missing commands but let's fix it right now with this I think. LGTM thanks for contributing!

PR #1957, opened 2024-03-13
docs(features): update store page

### LINKED ISSUE #1491: Create Store guide

pewsheen:

- New or updated content
  - Update `features/store` page
#### Description

- Closes #1491
  - Update `features/store` page


vasfvitor:
I feel that it could be explicit that `with_store` interoperates with the frontend. What do you think?

pewsheen:
> I feel that it could be explicit that `with_store` interoperates with the frontend. What do you think?

Actually I do not know the purpose of `with_store`. 🤔  I guess it is some kind of builder patent, but its usage has changed over the years.

I was testing the plugin on my Mac and iOS, the behavior seemed different and complicated on mobile.

**Mac** - The backend can interoperate with the frontend without using `with_store`. (maybe it's a coincidence?)

**iOS** - The `StoreBuilder` without assigning `mobile_plugin_builder` can't load or save the store to the mobile storage. And the way to make it work on mobile is just like how `with_store` does.

So I'm starting to wonder if the original purpose of the plugin is to let the backend and frontend share the store by default. And the reason we need to use `with_store` to interoperate with the frontend is caused by some other reason, like on the iOS?

FabianLars:
> Mac - The backend can interoperate with the frontend without using with_store. (maybe it's a coincidence?)

It can? How so? Or differently: Do you have a small code example?

> iOS - The StoreBuilder without assigning mobile_plugin_builder can't load or save the store to the mobile storage. And the way to make it work on mobile is just like how with_store does.

Yeah, mobile support is quite fresh and was contributed by a community member. The PR even made with_store private (probably by accident?) so the relation between mobile and with_store is still a bit unknown.

----

I'm also not a fan of with_store in general but back then it looked like it, or something like it would be required (i didn't add it myself). I'd like to rework large parts of the plugin, maybe get it closer to electron-store, maybe we can get rid of with_store then.

FabianLars:
okay, looking at with_store right now it indeed doesn't make that much sense to me anymore to let users use it. The only thing we'd have to think about if we want to remove it is the mobile plugin handle it sets for new stores. I'm not a fan of the explicit api for that, maybe a new storecollection::builder api makes sense idk. That's something to discuss in the plugins repo.

If someone agrees with me to remove with_store from the public interface (still makes sense internally ig) then i think we should get rid of it in this PR already.

pewsheen:
> okay, looking at with_store right now it indeed doesn't make that much sense to me anymore to let users use it. The only thing we'd have to think about if we want to remove it is the mobile plugin handle it sets for new stores. I'm not a fan of the explicit api for that, maybe a new storecollection::builder api makes sense idk. That's something to discuss in the plugins repo.
> 
> If someone agrees with me to remove with_store from the public interface (still makes sense internally ig) then i think we should get rid of it in this PR already.

I think the name `with_store` is confusing. It's like a "find or create store" function. I would prefer to extract the "find or create" logic and use it in the `StoreBuilder` and commands, and then we can refactor or remove the `with_store`.

FabianLars:
completely agree

vasfvitor (approved):

dreyfus92 (approved):
Thanks everyone, LGTM ✅

PR #1968, opened 2024-03-18
docs(features): update updater page

### LINKED ISSUE #1493: Create Updater guide

pewsheen:

- New or updated content
  - Update `features/updater` page

#### Description

- Close #1493 
  - Update `features/updater` page

### DIFF
--- a/src/content/docs/features/updater.mdx
+++ b/src/content/docs/features/updater.mdx
@@ -4,7 +4,7 @@ description: In-app updates for Tauri applications.
 ---
 
 import PluginLinks from '@components/PluginLinks.astro';
-import { Steps, Tabs, TabItem } from '@astrojs/starlight/components';
+import { Aside, Steps, Tabs, TabItem } from '@astrojs/starlight/components';
 import CommandTabs from '@components/CommandTabs.astro';
 
 <PluginLinks plugin="updater" />
@@ -17,6 +17,10 @@ The Updater plugin offers an update mechanism for Tauri applications distributed
 - App bundle (macOS)
 
 Once your Tauri project is ready, you can configure Tauri's updater to enable auto updating for your users.
+
+<Aside type="note" title="Choose only one of the Windows bundles">
+	Although we provide both `NSIS` and `MSI` for Windows, you should only choose one for package installation and update. If users have installed the package via `NSIS`, they can't update it using the `MSI`.
+</Aside>
 
 ## Supported Platforms
 
--- a/src/content/docs/features/updater.mdx
+++ b/src/content/docs/features/updater.mdx
@@ -129,7 +129,7 @@ Now you need to configure Tauri's updater. To do this, add this to your Tauri co
 }
 ```
 
-The required keys are `"endpoints"` and `"pubkey"` to enable the updater.
+The required keys are `"endpoints"` and `"pubkey"` to enable the updater. The `"endpoints"` field can also be set later in Rust.
 
 `"endpoints"`<a id="config-enpoints"/>:  must be an array of updater endpoint URLs as strings. TLS is enforced in production mode.<br/>
 Each updater URL can contain the following variables allowing you to determine [server-side](#static-json-file) if an update is available:
--- a/src/content/docs/features/updater.mdx
+++ b/src/content/docs/features/updater.mdx
@@ -277,7 +277,7 @@ The required keys are "url", "version" and "signature"; the others are optional.
 - `"version"` must be a valid semver, with or without a leading `v`, meaning that both `1.0.0` and `v1.0.0` are valid.
 - `"url"` must be a valid url to the update bundle.
 - `"signature"` must be the **content** of the generated `.sig` file. The signature may change each time you run `tauri build` so make sure to always update it.
-- `"notes"`: Here you can add notes about the update, like release notes. Tauri's default dialog will present this to the user when it asks if it's allowed to update.
+- `"notes"`: Here you can add notes about the update, like release notes.
 - `"pub_date"` must be formatted according to [RFC 3339](https://datatracker.ietf.org/doc/html/rfc3339#section-5.8) if present.
 
 ## Usage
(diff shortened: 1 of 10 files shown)


src/content/docs/features/updater.mdx
```diff
@@ -3,12 +3,362 @@ title: Updater
...
+import { Steps, Tabs, TabItem } from '@astrojs/starlight/components';
+import CommandTabs from '@components/CommandTabs.astro';
 
 <PluginLinks plugin="updater" />
 
-<Stub>
-	Based on
-	https://github.com/tauri-apps/plugins-workspace/tree/v2/plugins/updater
-</Stub>
+Updater plugin offers a updater for:
```
FabianLars:
```suggestion
The Updater plugin offers an update mechanism for Tauri applications distributed via:
```

src/content/docs/features/updater.mdx
```diff
@@ -3,12 +3,362 @@ title: Updater
...
+}
+```
+
+The required keys are `"version"`, `"platforms.[target].url"` and `"platforms.[target].signature"`; the others are optional.
+
+- `"version"` must be a valid semver, with or without a leading `v`, meaning that both `1.0.0` and `v1.0.0` are valid.
+- `"platforms"`: Each platform key is in the `OS-ARCH` format, where `OS` is one of `linux`, `darwin` or `windows`, and `ARCH` is one of `x86_64`, `aarch64`, `i686` or `armv7`.
+- `"url"` must be a valid url to the update bundle.
+- `"signature"` must be the **content** of the generated `.sig` file. The signature may change each time you run `tauri build` so make sure to always update it.
+- `"notes"`: Here you can add notes about the update, like release notes. Tauri's default dialog will present this to the user when it asks if it's allowed to update.
```
FabianLars:
```suggestion
- `"notes"`: Here you can add notes about the update, like release notes.
```
the default dialog was removed in v2

src/content/docs/features/updater.mdx
```diff
@@ -3,12 +3,362 @@ title: Updater
...
+	"notes": "These are some release notes"
+}
+```
+
+The required keys are "url", "version" and "signature"; the others are optional.
+
+- `"version"` must be a valid semver, with or without a leading `v`, meaning that both `1.0.0` and `v1.0.0` are valid.
+- `"url"` must be a valid url to the update bundle.
+- `"signature"` must be the **content** of the generated `.sig` file. The signature may change each time you run `tauri build` so make sure to always update it.
+- `"notes"`: Here you can add notes about the update, like release notes. Tauri's default dialog will present this to the user when it asks if it's allowed to update.
```
FabianLars:
```suggestion
- `"notes"`: Here you can add notes about the update, like release notes.
```
same

src/content/docs/features/updater.mdx
```diff
@@ -3,12 +3,362 @@ title: Updater
...
+			"windows": {
+				"installMode": "passive",
+				"installerArgs": []
+			}
+		}
+	}
+}
+```
+
+The required keys are `"endpoints"` and `"pubkey"` to enable the updater.
```
FabianLars:
```suggestion
The required keys are `"endpoints"` and `"pubkey"` to enable the updater. The `"endpoints"` field can also be set later in Rust.
```
idk how to phrase this but this is something that came up on discord a few days ago

src/content/docs/features/updater.mdx
```diff
@@ -3,12 +3,362 @@ title: Updater
...
+
+`"endpoints"`<a id="config-enpoints"/>:  must be an array of updater endpoint URLs as strings. TLS is enforced in production mode.<br/>
```
FabianLars, line 138:
```suggestion
`"endpoints"`<a id="config-enpoints"/>:  must be an array of updater endpoint URLs as strings. TLS is enforced in production mode.<br/>
All endpoints should generally return the same content, Tauri will fall back to the next endpoint URL if it couldn't connect to the first one or if it returned a non-2XX status code.<br/>
```
The logic in the plugin looks a bit off on first glance but this is how it worked in v1 and afaik how it's still supposed to work. Also adding this because of discord questions :)

FabianLars (changes requested):
One thing i don't know where to add is a note about that you have to choose between msi and nsis for the updater to work correctly. maybe also showing how?

(maybe also adding a note somewhere that you can update from msi to nsis if nsis is and stays in perMachine mode idk)

src/content/docs/features/updater.mdx
pewsheen: Nice catch. I forgot to remove it

pewsheen:
> One thing i don't know where to add is a note about that you have to choose between msi and nsis for the updater to work correctly. maybe also showing how?
> 
> (maybe also adding a note somewhere that you can update from msi to nsis if nsis is and stays in perMachine mode idk)

Do you have any reference or discussion or example (maybe?) about this? 

The server side needs to return only one of the bundler (.nsis or .msi)?

FabianLars:
> Do you have any reference or discussion or example (maybe?) about this?

Hmm, not really no.

Basically the problem here is that if a user installed the app via the nsis exe then the updater artifact must also be the nsis exe (==the zip with it inside). Same for the .msi

The updater however only differentiates between the OS and arch, but not the package type, and there's currently no easy way to know how an app was installed either so devs can't really polyfill this themselves.

pewsheen:
> > Do you have any reference or discussion or example (maybe?) about this?
> 
> Hmm, not really no.
> 
> Basically the problem here is that if a user installed the app via the nsis exe then the updater artifact must also be the nsis exe (==the zip with it inside). Same for the .msi
> 
> The updater however only differentiates between the OS and arch, but not the package type, and there's currently no easy way to know how an app was installed either so devs can't really polyfill this themselves.

Thanks for the explanation! I think I'll add this as a note to the `Server Support` section, so the static JSON and dynamic response should be aware of this.

FabianLars (approved):

PR #1974, opened 2024-03-20
docs(features): update Commands page

pewsheen:

- New or updated content
  - Update `/features/commands/`

#### Description

  - Update `/features/commands/`

It's basically copied from 1.0 guide, let's see what we can improve it

### DIFF
--- a/src/content/docs/features/commands.mdx
+++ b/src/content/docs/features/commands.mdx
@@ -1,7 +1,389 @@
 ---
-title: Commands
+title: Calling Rust from the frontend
 ---
 
-import Stub from '@components/Stub.astro';
+Tauri provides a simple yet powerful `command` system for calling Rust functions from your web app.
+Commands can accept arguments and return values. They can also return errors and be `async`.
 
-<Stub />
+## Basic Example
+
+Commands are defined in your `src-tauri/src/lib.rs` file. To create a command, just add a function and annotate it with `#[tauri::command]`:
+
+```rust title="src-tauri/src/lib.rs"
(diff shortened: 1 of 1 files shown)


pewsheen:
Maybe we should also mention this https://beta.tauri.app/guides/plugins/develop-mobile/#adding-mobile-commands

vasfvitor:
> Maybe we should also mention this https://beta.tauri.app/guides/plugins/develop-mobile/#adding-mobile-commands

I haven't played with mobile, do you happen to know if this is the default pattern to invoke from mobile? If so, I agree that it could be linked. What bothers me is that it's in the context of "Mobile Plugin Development", otoh, could be copied and adapted, but I'm not so sure about it

Maybe if this section (mobile commands) could be moved to a partial file and imported both in this guide and on plugin development? Only partially solves the issue of being in "Plugin development" context.

What do you think?

Note:
https://discord.com/channels/616186924390023171/662624589037436928/1219455566985035898

pewsheen:
Hmm.. Ok, I have changed my mind lol
I think this guide should focus on "Frontend -> Rust", and the mobile plugin command is another direction.

vasfvitor:
Where are we at this? I think it's ok, later we can revisit to maybe add a tauri specta tip when it releases stable to v2

pewsheen:
> Where are we at this? I think it's ok, later we can revisit to maybe add a tauri specta tip when it releases stable to v2

I’m also ok with it. We can improve it later

vasfvitor (approved):

Issue #1462, opened 2023-09-03
Guide to manually create a Tauri project


lorenzolewis:
Add to https://beta.tauri.app/2/guide/create/ (although it could be split into it's own page depending on how long it is).

simonhyll:
Need to verify the linked PR's contents hasn't gotten lost somewhere in my restructuring, because I can't find the contents anymore.

vasfvitor:
It should be in https://beta.tauri.app/start/create-project/ but now it's gone

Rigidity:
Interested in picking this up, I'll move the content back into the page

PR #1994, opened 2024-03-25
add(docs): manual setup guide

dreyfus92:

- New or updated content

#### Description
Adds manual setup guide using the Tauri CLI


vasfvitor:
related #1462

vasfvitor (commented):
Do you mind adding `cargo` instructions too?

dreyfus92:
> Do you mind adding `cargo` instructions too?

oof, I missed cargo 😅. I will handle it later.

src/content/docs/guides/create/index.mdx
```diff
@@ -65,6 +64,77 @@ You'll now see a new window open with your app running.
...
+    1. First install Tauri's CLI tool globally using your package manager of choice:
+
+        <CommandTabs
+            npm="npm install -g @tauri-apps/cli"
+            yarn="yarn global add @tauri-apps/cli"
+            pnpm="pnpm add -g @tauri-apps/cli"
+            cargo="cargo install tauri-cli"
+        />
+
+    2. Create a new directory for your project and initialize a your app. You can use any frontend framework you prefer, like NextJS, Nuxt, Svelte, or just plaing HTML, CSS, and JavaScript.
```
vasfvitor:
```suggestion
    2. Create a new directory for your project and initialize a your app. You can use any frontend framework you prefer, such as Next.js, Nuxt, Svelte, Yew, Leptos, or just plain HTML, CSS, and JavaScript.
```

src/content/docs/guides/create/index.mdx
```diff
@@ -65,6 +64,77 @@ You'll now see a new window open with your app running.
...
+            npm="mkdir my-tauri-app
+                cd my-tauri-app
+                npm <command to bootstrap a template for your frontend app>"
+            yarn="mkdir my-tauri-app
+                cd my-tauri-app
+                yarn <command to bootstrap a template for your frontend app>"
+            pnpm="mkdir my-tauri-app
+                cd my-tauri-app
+                pnpm <command to bootstrap a template for your frontend app>"
+        />
```
vasfvitor: I think that instead we can focus on the end goal that is to have a parent folder 'my-tauri-app' and a folder for the frontend app that can be called anything else 'my-tauri-frontend'. This can be a use case for the starlight [file tree](https://starlight.astro.build/guides/components/#file-tree).

Not exactly this wording, just a thought
```suggestion
      In this directory setup the frontend of your choice ...
      
      The folder structure should be
      - my-tauri-app
               - my-frontend
                           ....frontendfiles
```

what do you think?

src/content/docs/guides/create/index.mdx
```diff
@@ -65,6 +64,77 @@ You'll now see a new window open with your app running.
...
+        ```bash
+        tauri init
+        ```
+
+        After running the command it will display a prompt asking you for different options. Here you have an example:
+
+        ```md
+        ✔ What is your app name? my-tauri-app
+        ✔ What should the window title be? my-tauri-app
+        ? Where are your web assets (HTML/CSS/JS) located, relative to the "<current dir>/src-tauri/tauri.con ✔ Where are your web assets (HTML/CSS/JS) located, relative to the "<current dir>/src-tauri/tauri.conf.json" file that will be created? ../build
```
vasfvitor:
```suggestion
        ✔ Where are your web assets (HTML/CSS/JS) located, relative to the "<current dir>/src-tauri/tauri.conf.json" file that will be created? ../build
```

vasfvitor (changes requested):
added Rust frontend to the list and fixed a few typos. Oh and a suggestion on step 2

src/content/docs/guides/create/index.mdx
dreyfus92: I think that we should just give the user the choice to manage their own project however they want. It could be creating a directory or just live in the root tho. I don't see no problem with this part.

vasfvitor: then perhaps I think it's needed to add the cargo command for the front end aswell

vasfvitor (approved):
Thanks for this

src/content/docs/guides/create/index.mdx
dreyfus92: Will cargo work on this part?

vasfvitor: I haven't used, but I believe there's no need to use only node cli
for example:
https://trunkrs.dev/ cargo install --locked trunk

And so does Sycamore, each will have different usage but are Rust based

vasfvitor: https://tauri.by.simon.hyll.nu/samples/simple/

dreyfus92: gotcha, makes sense.

vasfvitor (approved):

PR #1995, opened 2024-03-25
chore(submodule):Update tauri-v1

vasfvitor:
For this one do we need to update anything on the docs? I mean, this module is utilized to generate api reference, also considering that we are leaning into having two separated docs (the old v1 and this one for v2): https://discord.com/channels/616186924390023171/662624589037436928/1212714709485232148


simonhyll:
After the restructure we're just having it as `packages/tauri` since we're no longer gonna make an effort to make a single docs site that fits all versions. Closing this.

PR #1996, opened 2024-03-26
Improve `/release` page

vasfvitor:
Now it utilizes native Starlight TOC. 

This is open for discussion about how it should look and function


amrbashir:
I prefer to have each release has its own dedicated route so it can have better links for example https://beta.tauri.app/releases/tauri/v2.0.0-beta.13/#enhancements this will lead to the the `Enhancements` section in `v2.0.0-beta.13` release, also having each release in its own route, will make the page smaller in size and load faster for user.

While the TOC looks good, we shouldn't rely on it at all, I'd prefer a custom list implementation (potentially with a filtering input box, e.g. type `2.0.0-beta` to get only `2.0.0-beta` releases).

simonhyll:
@amrbashir You can still do links like this:
https://deploy-preview-1996--tauri-docs-starlight.netlify.app/releases/tauri/#enhancements
We'd just need to update the links that get generated a bit so they target a specific release

I kinda like that it's not separate pages, it's performant and easy to overview.

We could make a similar implementation for this as we did for the features searching list, then we'd just have to put the input field at the top of the page
https://deploy-preview-1996--tauri-docs-starlight.netlify.app/features/

amrbashir:
Even if links are fixed, the generated files will only get bigger as time goes on which means more unnecessary traffic and slower load times for users.

dreyfus92:
I think that we could implement pagination to fix that issue @amrbashir

amrbashir:
fine I guess, but is it really worth that much effort that just plain pages?

vasfvitor:
I'm thinking about this whole situation and long term I personally think that each release having it own route would be the best, no broken links, no need to worry about pagination, less js (edit: less js considering we need to work around starlight design).

Then of course we could have a page for quick reference with all releases for a pkg, possible with a search.

The way I tried to do in PR (one route to rule them all) was to experiment and see it live, gather more ideas. But after seeing the problems arise: broken anchor, possible need to paginate, and mostly important no canonical links. Who knows what else may arise as those pages will always be changing on each release. All of those can be solved, but at what cost?

Now I'm wondering if having a separated releases.tauri.app would be best. This would avoid having to regenerate all the pages in the docs, and because in the current live implementation (dedicated route) the build time doubled:

`Build time: 2m 14s. Total deploy time: 2m 14s` https://app.netlify.com/sites/tauri-docs-starlight/deploys/660123329fd8ff00088ab0b9

After /releases PR:
`Build time: 4m 44s. Total deploy time: 4m 46s` https://app.netlify.com/sites/tauri-docs-starlight/deploys/66016d026d31f20008fb3300

Seeing how this will be a time sink (and doesn't help moving tauri to rc?) I'll leave as draft for a while, if anyone reading this wants to use and work on, go ahead

This is still open for ideas and discussion, unless someone from the team says otherwise ofc. 

thanks for the feedback @amrbashir @simonhyll and @dreyfus92

amrbashir:
I don't mind having a subdomain `releases.tauri.app`, in fact I lean more towards it.

dreyfus92:
Also as side note, hopefully soon Astro will have content collections caching which will lead to faster builds.

vasfvitor:
closing considering @simonhyll did a really good job at improving it recently, there's still some requested changes to do (I remember someone mentioned a summary table of versions or something) but this can go into another PR.

https://v2.tauri.app/release/

PR #2001, opened 2024-03-27
Fix: Fixed a jump in top navigation bar in non-English languages

### LINKED ISSUE #1989: (next) navigation bar ignores language setting
In non-English state, click the navigation bar to jump to the address error.
1. switch language to zh_cn
2. The language is switched to English after click any one
image

GuoJikun:

#### Description

- Closes #1989


vasfvitor (approved):
Nice, do you think it's possible to solve (preferably without shipping js) that when clicking on /blog then navigating back and it keeps the locale?

As of now this looks ok, lemme now so I can merge already

thanks

GuoJikun:
It's possible to store the current language information in the cache, but that's the only way I can think of for now.

GuoJikun:
Maybe copy the 'blog' directory into your language directory.

vasfvitor:
> It's possible to store the current language information in the cache, but that's the only way I can think of for now.

Do you think it's worth doing it? I'm happy with this as is.

vasfvitor:
Oh, I forgot about the /releases page

vasfvitor:
here: https://github.com/tauri-apps/tauri-docs/blob/next/src/components/overrides/Header.astro

GuoJikun:
> Oh, I forgot about the /releases page


What does the '/releases' page do? I don't know much about this.


image
That doesn't seem to be a problem.

vasfvitor:
> What does the '/releases' page do? I don't know much about this.

It lists Tauri release notes for core packages.

> That doesn't seem to be a problem.

This is odd, here it goes to root locale when navigating into it

GuoJikun:
> This is odd, here it goes to root locale when navigating into it

You are right. Even if I add the language prefix, it's still the root locale for navigation.

vasfvitor:
thanks

GuoJikun:
I noticed a 404 when going to '/zh-cn/release' in dev, but deploying to 'netlify' worked fine. @vasfvitor 

dev
image
netlify
image

vasfvitor:
`/release` pages are generated with a build command, so if you didn't run it, it will give 404, same with a js api pages. Thanks for bringing it up nonetheless @GuoJikun

GuoJikun:
Ok, got it

PR #2008, opened 2024-03-29
fix: biometric and nfc js api and stubs

### LINKED ISSUE #1988: (next) Fix js-api-generator to include biometric and nfc js api
And possible other new plugins

Maybe even make something dynamic that doesn't need to be always manually adding, editing the plugin list

vasfvitor:
- fix js api generator
- add stubs for nfc and biometric

closes #1988


dreyfus92 (approved):
Thank you @vasfvitor, LGTM ✅

PR #2029, opened 2024-04-03
fix: hide releases from search

### LINKED ISSUE #2028: Disable pagefind for release pages
While in theory it's amazing to be capable of searching releases, in practice it has completely taken over the entire search engine more or less, most things you search for will end up being release page results, which isn't the primary part you wanna be searching through in the main page search.

This option should be disabled for release pages: https://starlight.astro.build/reference/frontmatter/#pagefind

Release pages should also be searchable, but we'll do it as a dedicated search component instead on the releases page with its own index. The exact implementation for this is undecided but I'm imagining something similar to what we did for features.

FabianLars:
i kept it enabled for the topmost index file so that you can still search for the releases overview page at least 🤷 

fixes #2028


dreyfus92 (approved):
Thank you @FabianLars, LGTM ✅

PR #2035, opened 2024-04-04
Add set RUST_BACKTRACE=1 when using powershell on windows

Giuliano1993:
- New or updated content

#### Description
The provided command for setting RUST_BACKTRACE=1 on Windows was valid only if using cmd; while, if using PowerShell, that  command wouldn't work.

Added the powershell version too

### DIFF
--- a/docs/guides/debugging/application.md
+++ b/docs/guides/debugging/application.md
@@ -18,12 +18,20 @@ Sometimes you may have an error in your Rust code, and the Rust compiler can giv
 RUST_BACKTRACE=1 tauri dev
 ```
 
-or like this on Windows:
+or like this on Windows if you're on cmd:
 
 ```shell
 set RUST_BACKTRACE=1
 tauri dev
 ```
+
+or like this on Windows if you're using PowerShell:
+
+```shell
+$env:RUST_BACKTRACE=1
+tauri dev
+```
+
 
 This command gives you a granular stack trace. Generally speaking, the Rust compiler helps you by
 giving you detailed information about the issue, such as:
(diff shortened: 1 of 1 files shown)


docs/guides/debugging/application.md
```diff
@@ -18,13 +18,21 @@ Sometimes you may have an error in your Rust code, and the Rust compiler can giv
 RUST_BACKTRACE=1 tauri dev
 ```
 
-or like this on Windows:
+or like this on Windows if you're on cmd:
```
FabianLars:
```suggestion
or like this on Windows if you're using cmd.exe:
``` adding the .exe makes it a bit clearer.

File: docs/guides/debugging/application.md [hunk 2]
```diff
@@ -18,13 +18,21 @@ Sometimes you may have an error in your Rust code, and the Rust compiler can giv
...
 
 ```shell
 set RUST_BACKTRACE=1
 tauri dev
 ```
 
+or like this on Windows if you're using PowerShell:
+
+```shell
```
FabianLars: i think you can use `powershell` instead of `shell` here. should fix the highlighting.

FabianLars (changes requested):
thank you! just 2 small things :)

docs/guides/debugging/application.md
Giuliano1993: Yes you're right! Fixed it right away!

Giuliano1993:
Thank you for the review @FabianLars ! Now it should be okay! :)

FabianLars (approved):
thanks!

PR #2038, opened 2024-04-04
fix: remove extra backtick in sql guide

vasfvitor:
page had extra \`, apparently it isn't causing any issues


dreyfus92 (approved):
Nice catch! LGTM ✅

PR #2041, opened 2024-04-05
Update next-js.mdx with correct bunx installation command.

adityabrahmankar:
Fixed the bunx create next-app command.

- New or updated content
- Translated content
- Something else!

#### Description

- Closes # 

### DIFF
--- a/docs/guides/getting-started/setup/next-js.mdx
+++ b/docs/guides/getting-started/setup/next-js.mdx
@@ -57,7 +57,7 @@ pnpm create next-app --use-pnpm
   <TabItem value="Bun">
 
 ```shell
-bunx create-next-app --use-bun
+bunx create next-app --use-bun
 ```
 
   </TabItem>


vasfvitor:
Hi, thanks for catching this, but Next.js docs says it is the way current it is `create-next-app`. I haven't tested https://nextjs.org/docs/pages/api-reference/create-next-app so maybe it's outdated?
image

If you have tested it to be different, let me know so I can reopen this and test myself. Then we can open a issue upstream to make sure it's not just a typo

adityabrahmankar:
image

bun create next-app is the correct way which works as shown in the official bun documentation.

FabianLars:
It has to be bun create instead of bunx create then though, right? Since bunx is an alias for bun x (the binary run command) I assume it will cause issues? I don't have bun installed / I'm not home so I can't test this rn

vasfvitor (approved):
Here what I get on windows, fresh bun out of the oven:
`bunx create-next-app --use-bun` error
`bunx create next-app --use-bun` works
same with `bun`

Oddly, yesteday, on next.js changed the docs to the current non working case
https://www.github.com/vercel/next.js/pull/64151

On bun cli it states it should be not dashed https://github.com/oven-sh/bun/blob/main/src/cli.zig#L1669

In a quick test, I noticed that `bunx create whatever-app` is an alias to `create-wahatver-app`, so `bunx create tauri-app` works to as `create-tauri-app`.

Again, thanks for bringing this up and citing the source docs which I should have looked at myself.

dreyfus92 (approved):
Thanks everyone, LGTM ✅

PR #2042, opened 2024-04-05
fix: NixOS setup -- shell.nix must also depend on librsvg

LeUlukai:

#### Description
The flakes.nix variant includes the dependency of librsvg, but the shell.nix variant does not. Following the shell.nix approach leads to 
```
Your system is missing dependencies (or they do not exist in $PATH):
│ rsvg2 │ Visit https://tauri.app/v1/guides/getting-started/prerequisites#setting-up-linux │
```
Thus, we have to add librsvg as dependency for shell.nix as well, which fixes the issue above.

### DIFF
--- a/docs/guides/getting-started/prerequisites.md
+++ b/docs/guides/getting-started/prerequisites.md
@@ -263,6 +263,7 @@ let
     glib
     dbus
     openssl_3
+    librsvg
   ];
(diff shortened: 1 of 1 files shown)


FabianLars (approved):
thanks :)

PR #2046, opened 2024-04-07
feat: restructured pages with autogenerate and badges

simonhyll:
Restructures the project.

- New or updated content

#### Description
Updated content layout, moved a component, some cleanup, badges, sidebar based on autogenerate, added partners page.

- Did you change something visual? 
Yes, the sidebar is substantially different now.


vasfvitor:
In a quick look, I think this could work, aside from the v2 js references that are unnecessarily nested
image

And what do you think about separating features into "plugins" section, and default/top level features?

simonhyll:
@vasfvitor Yea there's definitely some layers of depth too many, I'm just not sure how to fix it other than changing the folders around since autogenerate doesn't support setting depth.

> And what do you think about separating features into "plugins" section, and default/top level features?
Lorenzo and I talked very briefly about it in Discord and we do need to have bit of a thinker about how we wanna handle that. On the one hand I don't wanna put *too* many things at the top level and Features is more than just plugins. Maybe `Features > Plugins` and `Features > Recipes`? Something like that, so basically nesting most of the current features one layer deeper under Plugins

PR #2073, opened 2024-04-16
Docs: plugin development documentation fixes

WalrusSoup:
- New or updated content

#### Description
- Adds additional information to the plugin development area, specifically that users need tauri-cli to be on beta as well
- Fixes the commands to match the working version used with the latest beta release 
- Added prefix to folder name

### DIFF
--- a/src/content/docs/guides/plugins/index.mdx
+++ b/src/content/docs/guides/plugins/index.mdx
@@ -33,19 +33,31 @@ By default Tauri prefixes your plugin crate with `tauri-plugin-`. This helps you
 
 ## Initialize Plugin Project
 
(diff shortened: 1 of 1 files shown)


FabianLars:
Thank you! I personally don't think that the version note makes much sense right now, every page of the docs assume you're using the latest beta version.

Switching out init with new (init still exists btw) is good though. And while you at it, could you also update the folder structure? It webview-src and -dist don't exist anymore for example. no problem if not, we can do that tomorrow too :)

WalrusSoup:
Updated with suggestions :+1:

FabianLars (approved):

PR #2074, opened 2024-04-17
Use correct method in file selector dialog example

Oughie:
Corrects inconsistency in docs: Instead of suggesting to use `show()`, it should suggest `pick_file()`, which is actually used in the Rust example.

#### Description

Change `show()` to `pick_file()` when recommending the user which method to use.
The example actually uses `pick_file()`.


FabianLars (approved):
thanks

PR #2088, opened 2024-04-23
[i18nIgnore]docs: redirects broken links

vasfvitor:
There are probably more, I did a few that was easy to notice

`upgrade--migrate` because the directory is now named "Upgrade & Migrate", the & becomes a dash

PR #2093, opened 2024-04-30
Add redirect for Tauri v2 upgrade guide

oscartbeaumont:
Add a redirect from `/guides/upgrade-migrate/from-tauri-1` to `/start/upgrade--migrate/from-tauri-1/`.

If you google "tauri v1 to v2" or similar you'll end up with this as the top entry but it points to a 404 page.

Screenshot 2024-04-30 at 1 03 35 PM


vasfvitor:
when I ran into the same problem https://github.com/tauri-apps/tauri-docs/pull/2088

oscartbeaumont:
Oh my bad for the duplicate I might close this then cause yours has other redirects too.

Definitely hope Tauri can get this fixed soon!

PR #2115, opened 2024-05-03
[v2] document tauri::dev()

vasfvitor:
based on https://github.com/tauri-apps/tauri/blob/07ff78c2de74d3bd85328ce6536f8a858be89128/.changes/dev-fn.md


simonhyll:
I'm not sure if this belongs in the Debug category, it's more of a general development tool 🤔
Maybe we should make an entire article as e.g. `Develop > System Information` where we document how to grab a whole bunch of different variables that can be used to determine how the app is set up, both at runtime and compile time?

vasfvitor:
Yeah, placing here was just a way to put it out. I can create a System Information page as stub and place this there, but most system info guess it's retrieve from the os info plugin

simonhyll:
Lets merge it on the basis of it being an improvement, getting the information put out there *somewhere*, then move it to where it's appropriate in a followup PR.

PR #2117, opened 2024-05-06
docs: Update frontend configs to latest versions

FabianLars:
_still not reading the template_ 🕺

### DIFF
--- a/src/content/docs/start/Frontend Configuration/index.mdx
+++ b/src/content/docs/start/Frontend Configuration/index.mdx
@@ -12,12 +12,6 @@ Tauri is frontend agnostic and supports most frontend frameworks out of the box.
 
 If a framework is not listed then it may work with Tauri with no additional configuration needed or it could have not been documented yet. Any contributions to add a framework that may require additional configuration are welcome to help others in the Tauri community.
 
-:::tip[Framework Not Listed?]
-
-Don't see a framework listed? It may work with Tauri without any additional configuration required. Read the [configuration checklist](/start/frontend-configuration/#configuration-checklist) for any common configurations to check for.
-
-:::
-
 ## Configuration Checklist
 
 Conceptually Tauri acts as a static web host. You need to provide Tauri with a folder containing some mix of HTML, CSS, Javascript and possibly WASM that can be served to the webview Tauri provides.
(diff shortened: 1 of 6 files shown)


src/content/docs/start/Frontend Configuration/index.mdx
```diff
@@ -28,18 +22,24 @@ Below is a checklist of common scenarios needed to integrate a frontend with Tau
...
 {/* TODO: Concept of how to do a client-server relationship? */}
 
-- Use static site generation (SSG). Tauri doesn't officially support server based alternatives (such as SSR).
+- Use static site generation (SSG), single-page applications (SPA), or classic multi-page apps (MPA). Tauri does not natively support server based alternatives (such as SSR).
 - For mobile development, a development server of some kind is necessary that can host the frontend on your internal IP.
 - Use a proper client-server relationship between your app and your API's (no hybrid solutions with SSR).
 
 ## JavaScript
 
+{/* TODO: Help me with the wording here lol */}
```
FabianLars, line 31: 😢

src/content/docs/start/Frontend Configuration/index.mdx
```diff
@@ -28,18 +22,24 @@ Below is a checklist of common scenarios needed to integrate a frontend with Tau
...
 	<LinkCard title="Next.js" href="/start/frontend-configuration/nextjs" />
 	<LinkCard title="Nuxt" href="/start/frontend-configuration/nuxt" />
 	<LinkCard title="Qwik" href="/start/frontend-configuration/qwik" />
-	<LinkCard title="Svelte" href="/start/frontend-configuration/svelte" />
-	<LinkCard title="Vite" href="/start/frontend-configuration/vite" />
+	<LinkCard title="SvelteKit" href="/start/frontend-configuration/sveltekit" />
+	<LinkCard
+		title="Vite (recommended)"
+		href="/start/frontend-configuration/vite"
+	/>
```
vasfvitor, line 42: Since Vite is the recommended, I think placing it in first position makes sense

FabianLars: probably a good idea if we all agree to mark it as recommended. Just have to remember that the rest keeps being sorted alphabetically so we don't end up with an unsorted mess 😂

vasfvitor: well, CTA always will create a vite project if ts/js is chosen right? that's beyond recommended I'd say 😄. I personally agree in having it recommended

>Just have to remember that the rest keeps being sorted alphabetically so we don't end up with an unsorted mess

Oh, so... I was thinking about smth that fits this. Having link guide for vite separated in a section before those guides, but this is stuff for another PR. I think for now then let's no change its position then.

vasfvitor (approved):
In my opinion we can leave the wording to a follow up PR, considering there's more context to add. @dreyfus92 do you have any suggestion here?

dreyfus92 (approved):
This looks good to me, we can definitely save the JS section for later.

PR #2135, opened 2024-05-10
Add a guide on making a bespoke `Titlebar` component

dan-myles:
- New or updated content

#### Description
This PR adds a new guide that goes in-depth on making a custom titlebar for Tauri applications.
It also goes over conditional rendering using some plugins. I think it'll be a great addition to the
"learn" section, but can be moved around, of course.

I've also added a "bun" command to the command tabs component, needed it throughout the guide.


dan-myles:
Here is a link to the deployment preview and the *specific* article that was written:
https://deploy-preview-2135--tauri-v2.netlify.app/learn/custom-titlebar/

src/components/CommandTabs.astro
```diff
@@ -7,9 +7,10 @@ interface Props {
 	yarn?: string;
 	pnpm?: string;
 	cargo?: string;
+  bun?: string;
```
mhsnook: It's cool to see this article and the site preview. Looks nice :) But I think you will have an easier time getting the article merged if you don't also include this unrelated change in the same PR.

vasfvitor: you're right, and now it isn't necessary anymore as we already implemented bun support on CommandTabs

dan-myles: yeah, this change was not meant to be included in the original PR. Will update.

PR #2143, opened 2024-05-13
feat: app size documentation

simonhyll:
Adds some thoughts on Cargo.toml configuration for improving app size.

### DIFF
--- a/src/content/docs/concept/size.mdx
+++ b/src/content/docs/concept/size.mdx
@@ -7,11 +7,9 @@ sidebar:
     variant: caution
 ---
 
-import Stub from '@components/Stub.astro';
-
-<Stub />
-
 import { Tabs, TabItem } from '@astrojs/starlight/components';
+
+While Tauri by default provides very small binaries it doesn't hurt to push the limits a bit, so here are some tips and tricks for reaching optimal results.
 
 ## Cargo Configuration
 
@@ -24,14 +22,14 @@ Dependent on whether you use the stable or nightly Rust toolchain the options av
 ```toml
 # src-tauri/Cargo.toml
 [profile.dev]
-incremental = true
+incremental = true # Compile your binary in smaller steps.
 
 [profile.release]
-codegen-units = 1
-lto = true
-opt-level = "s"
-panic = "abort"
-strip = true
+codegen-units = 1 # Allows LLVM to perform better optimization.
+lto = true # Enables link-time-optimizations.
+opt-level = "s" # Prioritizes small binary size. Use `3` if you prefer speed.
+panic = "abort" # Higher performance by disabling panic handlers.
+strip = true # Ensures debug symbols are removed.
 ```
 </TabItem>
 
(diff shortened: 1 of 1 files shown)


src/content/docs/concepts/size.mdx
```diff
@@ -10,3 +10,93 @@ sidebar:
...
+opt-level = 3
+rustflags = ["-Zthreads=8"]
+
+[profile.release]
+codegen-units = 1
+lto = true
+opt-level = s
+panic = "abort"
+strip = true
+rpath = false
```
FabianLars: false is the default iirc. for the explanations below, i also doubt this will have an effect on compile times.

edit: to be clear, this comment is about rpath

simonhyll: Some texts may need updating, copy-pasted some of them around and it made me forget which ones I had to update 😅 
And you're right, it's the default, I probably have it in my profiles for the sake of being explicit, which is a practice I like doing but won't be forcing upon the community lol

src/content/docs/concepts/size.mdx
```diff
@@ -10,3 +10,91 @@ sidebar:
...
+- [trim-paths:](https://rust-lang.github.io/rfcs/3127-trim-paths.html) Removes potentially privileged information from binaries.
+  - `dev` `omit`: Irrelevant during development.
+  - `release` `"all"`: Removes paths from the built binary which may in some cases be seen as privileged data.
+- [rustflags:](https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#profile-rustflags-option) Sets Rust compiler flags on a profile by profile basis.
+  - `"-Cdebuginfo"`: Whether debuginfo symbols should be included in the build.
+    - `dev` `omit`: You want debuginfo during development.
+    - `release` `0`: Depending on your Rust version this may or may not have an effect on the symbols that get stripped.
+  - `"-Zthreads"`: Increases the number of threads used during compilation.
+    - `dev` `8`: Adjust as is appropriate based on developer and CI/CD hardware.
+    - `release` `8`: Has no effect on the resulting binary beyond it compiling faster.
```
amrbashir: Tbh, I don't like this section at all, it seems bloated and also needs to understand a special syntax.

I think I'd prefer the simplicity of the snippet from v1 docs which just have a small explanation next to each option

```toml
[profile.release]
panic = "abort" # Strip expensive panic clean-up logic
codegen-units = 1 # Compile crates one after another so the compiler can optimize better
lto = true # Enables link to optimizations
opt-level = "s" # Optimize for binary size
strip = true # Remove debug symbols
```

We can then replace this bloated section with just links to the official cargo docs about it:
```md

# References
- [lto](link-to-cargo-docs)
..etc
```

vasfvitor (commented):
welp, it took me a while to understand the features section, I glanced over the note, then I read it, and only later after checking the config in rust docs was that I understood the note.

config codeblock

take a time to read about the impact of these options at profile.release as they may impact in so and so... make sure .... and done

simonhyll:
I'll restructure it to be a shorter and sweeter version I guess.

src/content/docs/concepts/size.mdx
simonhyll: Restructured to what I think is closer to what you want.

amrbashir (approved):

PR #2151, opened 2024-05-14
Update typedoc

lorenzolewis:

#### Description

- Updates us to be on the stable release of typedoc
- Remove custom patched version of `typedoc-markdown-plugin`, `typedoc`, and `typedoc-plugin-mdn-links`
- Here are the changes you can expect to see rendered:

1. A `<a>` tag is now added beneath all headings. This is due to the https://www.typedoc-plugin-markdown.org/docs/options#usehtmlanchors options being used. This enables us to link to things that are within tables (such as enum variants). These are not rendered, only present in the HTML.
2. Constructor headings are now named more specifically. For example, `##### constructor()` is now `##### new Authenticator()`
3. Links to MDN have a `↗️` symbol in them. This is set to be removed in https://github.com/typedoc2md/typedoc-plugin-markdown/commit/8962a9f98abc74d512ccc1e1fe8b5c36af95c661
4. Horizontal breaks between sections (`—`) are no longer rendered between headings greater than an `h4`
5. Various whitespace adjustments
6. `Returns` and `Parameters` sections are now rendered below the definition and above any `Example` headings (used to be below the `Since` heading at the very bottom)

Since the API files aren't committed you can't easily see the changes in the diff view. I've created a separate gist that includes a full diff of the before and after: https://gist.github.com/lorenzolewis/ebecfc28c3dd5ca8372f520f6ad26557

The only thing that is slightly different is an error I caught in `source` where it was referencing the local filesystem after the diff was taken. This has been fixed but was after the diff was taken.


FabianLars (approved):
❤️

lorenzolewis:
Nothing like a 17217 line diff 😅

PR #2174, opened 2024-05-21
CSP Minimization

tweidinger:

- CSP adaption

#### Description

Testing how minimal the CSP could be defined.

Found no errors after minimal tweaking.
One thing left to do is setup the CSP plugin in Netlify to automatically nonce the inline scripts.


simonhyll (changes requested):
The reasoning behind having it in both the headers and the meta tag is two-fold, firstly it makes it harder to bypass it (headers are quite easy to get rid of with a proxy, CSP in the content makes it more annoying to scrub) so it's basically enforced better if it's in both. Secondly it's because the meta tag version applies during development, making it easier to detect issues while you're developing. While the headers are the cooler place, if we're going to pick only one place for it I'd probably wanna go with the meta tag version instead to get that at development time detection handled

~~Also, for some unknown reason, the mobile menu is affected by these changes, no idea how~~ Actually the white bar at the bottom on mobile is related to the Netlify preview bar thingie not being able to render properly anymore. The mobile menu is fine, Netlify is not.

tweidinger:
> firstly it makes it harder to bypass it (headers are quite easy to get rid of with a proxy, CSP in the content makes it more annoying to scrub) 

I don't understand this point from a threat model perspective to be honest. The CSP is a browser security feature on a TLS encrypted connection and the only proxy able to break this can also just modify the content of the page.


>Secondly it's because the meta tag version applies during development, making it easier to detect issues while you're developing. 

That's a very good point. The only thing we can't test in meta tag is the `frame-ancestor` policy, which should be fine to not have for development.
I saw that astro supports setting custom headers in dev and preview. I'll check if we can use this to serve the same netlify headers defined in the `public/_headers` file. Otherwise I would re-add the meta tag CSP. Worst thing is desync and too strict CSP (more strict setting overrides weaker setting), which will be noticed.

tweidinger:
So it works locally with just the header set and serving them also in `astro dev` and `astro preview` from the same `public/headers` file.

I haven't written js in ages and just made it work. If it is not a good idea to read values of a file during config setup I am up for reverting things 😅 otherwise we have the single source headers in dev and production then.

simonhyll (approved):
LGTM :) I would consider possibly adding them to the meta tag as well since it is a slight security bump with (now) no added maintenance, but I'm fine with merging this as-is. Good job!

PR #2186, opened 2024-05-23
fix: broken links by adding default append slashes to internal redirects

### LINKED ISSUE #2185: [bug] Broken links on Core Concepts Page
### Question you want answered

Core Concepts Page has some broken links

### Where did you look for an answer?

I think is the same problem as before. Missing trailing slashes.

### Page URL

 https://v2.tauri.app/concept/

### Additional context

When I click on this card the this happens:
image
image

### Are you willing to work on this yourself?

- [X] I want to work on this myself

gabenddos:

#### Description

- Closes #2185
- Only fixes links with missing append slashes. 

For the dummy links like the `/start/create-project/` that points to itself and not an actual link I fixed the redirect to prevent the error but also added a TODO so we can more easily track these dummy links in the future.

TL;DR: It was broken and a dummy link. Now it's just a dummy link with a TODO hint on the code = )
image


gabenddos:
@simonhyll hi just fixed a toon of broken links in this, for content itself nothing should be changed = )

vasfvitor:
_eae,_ I can't seem to reproduce this issue, tried Firefox and Chrome. Could you elaborate?

gabenddos:
Shure, basically I was using the docs and I noticed that always I used a link that didn't have an append slash the redirect would break.
When I was looking for more broken links I noticed a strange behavior that sometimes it would auto-complete the trailing slash and not break. But if I use the go back in the browser and try to access the same link again it breaks 100% of time.

I'll try to record a nimbus to show what I'm saying

gabenddos:
Here it is!

https://github.com/tauri-apps/tauri-docs/assets/36337762/ef281876-0f0f-4572-b9f4-f8f866dd2672

> I'm sorry for the awful sounds, please mute the video, it's 6 AM in my timezone, just woke up, it's cold AF and I'm having terrible rhinitis 💀

FabianLars:
We had that issue in the v1 docs (or earlier v2 docs can't remember) a while back so i can at least confirm that this _can_ be an issue x)

gabenddos:
As in the video above, I tried with Chrome and Safari and sometimes the trailing slash don't apply and the page breaks 😢 
If it helps, I also noticed that the missing trailing slash does not cause any issues while running on localhost.
That's why I think that the problem is also related to some deployment configurations like a NGINX conf or something like that.

@vasfvitor

Rigidity:
I've always removed trailing slashes from my work's docs site 😅

FabianLars:
Ideally we can make both variants behave the same... Can't find the conversations about the last time so idk what we did there.

vasfvitor:
now I'm wondering if isn't there a  linting setting that will do this automagically

I see no reason to not merge this once reviewed, would be a good idea to set config to always to enforce?

https://docs.astro.build/en/reference/configuration-reference/#trailingslash

plus, https://github.com/tauri-apps/tauri-docs/blob/v2/public/_redirects would need to be updated  if so

simonhyll (approved):

simonhyll:
We can fix trailing slash management in a followup PR, the updates LGTM, shipping.

Thanks for the hard work!! ❤️

PR #2193, opened 2024-05-23
fix: i18n pages structure

simonhyll:

- New or updated content
- Translated content
- Something else!

#### Description

- Closes # 


vasfvitor (approved):
if it builds, then it's approved. Joking, but considering no new page will be rendered, it just fixes the structure so people can easily come back and translate right?

locales.json
```diff
@@ -11,6 +11,14 @@
...
     "lang": "es"
   },
+  "it": {
+    "label": "Italiano",
+    "lang": "it"
+  },
+  "ko": {
+    "label": "한국어",
+    "lang": "ko"
+  },
```
vasfvitor: those _ko and _it only exists because they were in v1 docs, but I don't think they should be enabled, there's next to zero content

simonhyll: Ok I guess we'll remove it completely then for now

PR #2202, opened 2024-05-24
feat: d2 diagrams, process model

simonhyll:
Requires that d2 is installed on the machine, there's no NPM dependency we can add for that. I wouldn't say that's too big a deal however seeing as we're currently pushing images manually anyway, now we'd just have a nice tool for doing so.

<https://astro-d2.vercel.app/>


vasfvitor (approved):
do you think it's worth to add a note to CONTRIBUTING.md mentioning the process of adding new diagrams?

vasfvitor:
we could even open a new good first issue to replace our current diagrams with d2

simonhyll:
> do you think it's worth to add a note to CONTRIBUTING.md mentioning the process of adding new diagrams?

Yea that's probably a good idea, I'll add a note about it to the PR

> we could even open a new good first issue to replace our current diagrams with d2

Mmmmmmaybe, depends on how quick I get done updating some of the diagrams x) If I get bored we can for sure make good first issues for it

PR #2204, opened 2024-05-24
feat: added diagram for Tauri Architecture

simonhyll:

- New or updated content
- Translated content
- Something else!

#### Description

- Closes # 


vasfvitor (approved):
great, now upstream is really at the top, and not below

dreyfus92 (approved):
Thank you @simonhyll, LGTM ✅

FabianLars:
With the refactorings to make multiwebview happen the upstream part isn't 100% right, runtime-wry now pulls in both wry in tao directly.

The diagram still makes sense considering that wry will be used inside a tao window but just in case this change was missed.

PR #2212, opened 2024-05-27
feat: bun command tab and lower case commands

vasfvitor:
add bun to command tabs and change yarn and cargo to lowercase


dreyfus92 (approved):
Thank you @vasfvitor, LGTM ✅

PR #2256, opened 2024-06-06
feat: turns theme selector into a button instead of a dropdown

simonhyll:
Stole this from one of my other projects because I think it's much nicer than having a dropdown for theme selection when we only have 2 themes, dark and light.


vasfvitor (approved):
oh yeaahhh

PR #2265, opened 2024-06-08
fix(prerequisites): remove duplicated env path of `NDK_HOME`

zennolux:

remove duplicated env path of `NDK_HOME`

#### Description

When setting the Android environment of  `NDK_HOME` on Windows, the variable `$VERSION` returns the full path already, so there is no need to add the duplicated path before  `$VERSION`.

Before:
image

After:
image


FabianLars:
Hmm, it does only print the version for me so the current docs is what's required on my end, i wonder what's going on 🤔 
What does make it work me like it works for me is changing the env set command to this:
```pwsh
[System.Environment]::SetEnvironmentVariable("NDK_HOME", $VERSION.FullName, "User")
```

Not sure what the best path forward here is...

zennolux:
> ```powershell
> [System.Environment]::SetEnvironmentVariable("NDK_HOME", $VERSION.FullName, "User")
> ```

This also works for me.

And replacing `-Path` with `-Name` makes sense in my test.

image

FabianLars (approved):

FabianLars:
thank you

PR #2267, opened 2024-06-09
fix: Tray click type

Levminer:
- New or updated content

#### Description
* The tray had changes made in: https://github.com/tauri-apps/tauri/pull/9869, but the migration docs does not reflect that.
* The french translation uses the old window getter


FabianLars (approved):
thanks :)

PR #2273, opened 2024-06-11
WIP: Align Status and Minor Fixes

tillmann-crabnebula:
This is a WIP PR to align some understanding of where we are at.

* Add some more labels like `1.x` or `Stub`
* Remove WIP from things
* Add missing pages
* Generally minor fixes

Provides a base understanding for @FabianLars to continue on after our sync call.


simonhyll:
Note that pages are still considered to be WIP if they're either
A) Not marked as i18n ready, which tonnes of the pages in this PR aren't
B) Not written yet
C) Haven't been verified to be upgraded and working with v2

PR #2287, opened 2024-06-13
feat: State Management page

### LINKED ISSUE #676: [docs] Add a guide about `tauri::State`.
Managing state of the application is a very common problem. Consider adding a section on managing state of the application in rust on [this page](https://tauri.app/v1/guides/features/).  

The documentation is available in [docs.rs](https://docs.rs/tauri/latest/tauri/struct.Builder.html#method.manage) but linking this page in features section of the main website will be more helpful.

Rigidity:
This is a first draft, so let me know if you have any feedback :)

#### Description

Resolves https://github.com/tauri-apps/tauri-docs/issues/676 by adding a State Management page to the docs.


FabianLars:
I only skimmed through but one thing i'd like to have added is that you can use the `state()` method to get the state if you can't rely on the command injection (for example if you need to move it into a thread, where using an AppHandle is easier, or if you're simply not in a command but in idk on_window_event for example)

src/content/docs/develop/state-management.mdx
```diff
@@ -2,32 +2,144 @@
...
+#[tauri::command]
+fn increase_counter(state: State<'_, Mutex<AppState>>) -> u32 {
+  let mut state = state.lock().unwrap();
+  state.counter += 1;
+  state.counter
+}
+```
+
+For more information on commands, see [Calling Rust from the Frontend](/develop/calling-rust).
+
```
dreyfus92, line 106: Dunno if this small section would do the work @FabianLars, let me know what you think.
```suggestion
### Accessing State Outside Commands

Sometimes you may need to access the state outside of commands, such as in a different thread or in an event handler like on_window_event. In such cases, you can use the state() method of the AppHandle to get the state:

<!-- code snippet -->
rust
use tauri::Manager;

// In an event handler:
fn on_window_event(app: &AppHandle, event: WindowEvent) {
  let state = app.state::<Mutex<AppState>>();
  let mut state = state.lock().unwrap();
  state.counter += 1;
}

This method is useful when you cannot rely on command injection. For example, if you need to move the state into a thread where using an AppHandle is easier, or if you are not in a command context.
```

Rigidity:
> I only skimmed through but one thing i'd like to have added is that you can use the `state()` method to get the state if you can't rely on the command injection (for example if you need to move it into a thread, where using an AppHandle is easier, or if you're simply not in a command but in idk on_window_event for example)

Ok I did mention how you can read state directly but I'll add the section on using it in practice like that

src/content/docs/develop/state-management.mdx
Rigidity: I implemented this, and made some changes after testing it locally

dreyfus92 (approved):
Thank you for this @Rigidity, LGTM ✅

src/content/docs/develop/state-management.mdx
```diff
@@ -2,32 +2,180 @@
...
+    app.manage(AppData {
+      welcome_message: "Welcome to Tauri!",
+    });
+    Ok(())
+  })
+  .run(tauri::generate_context!())
+  .unwrap();
+```
+
+You can later access your state via the `App` instance:
```
amrbashir: This is partially true, you can access the state through any type that implements the `Manager` trait

src/content/docs/develop/state-management.mdx
```diff
@@ -2,32 +2,180 @@
...
+let data = app.state::<AppData>();
+```
+
+For more info, including accessing state in commands, see the [Accessing State](#accessing-state) section.
+
+## Mutability
+
+In Rust, values which can be shared between multiple owners (for example, sent to different threads) cannot be mutable. This is to prevent multiple owners from modifying the value at the same time, which can lead to issues such as data races.
+
+To work around this, you can use the standard library's `Mutex` to wrap your state. This allows you to lock the value when you need to modify it, and unlock it when you are done.
```
amrbashir: I would make this a bit general like "you can use one of the various techniques to achieve mutability" and link to these Rust book sections https://doc.rust-lang.org/book/ch15-05-interior-mutability.html  and https://doc.rust-lang.org/book/ch16-00-concurrency.html

After that we can make a suggestion like "for this example, we will use `Mutex` which locks the value...etc"

src/content/docs/develop/state-management.mdx
```diff
@@ -2,32 +2,180 @@
...
+state.counter += 1;
+```
+
+At the end of the scope, or when the `MutexGuard` is otherwise dropped, the mutex is unlocked automatically so that other parts of your application can access and mutate the data within.
+
+### Do you need `Arc`?
+
+It's common to see `Arc` used in Rust to share ownership of a value across multiple threads. However, in Tauri, you typically don't need to use `Arc` to wrap your state's `Mutex` because the state is already internally managed and shared across threads.
+
+The only time in which it would make sense to use `Arc` for your state is when you need to share the state across multiple parts of your application that are not managed by Tauri itself.
```
amrbashir: I would honestly remove this section; I don't see much value in it or maybe the first paragraph is enough.

src/content/docs/develop/state-management.mdx
```diff
@@ -2,32 +2,180 @@
...
+
+## Accessing State
+
+You are now storing state in your application, but the obvious next question would be how to access the state in commands.
+
+This is how you can access state (and optionally modify it) within commands:
```
amrbashir: Maybe make these into sections with heading:

```md
#### Access state in Command

#### Access state using `Manager` trait
```
and so on

src/content/docs/develop/state-management.mdx
```diff
@@ -2,32 +2,180 @@
...
+```rust
+use tauri::{async_runtime::Mutex, State};
+
+#[tauri::command]
+async fn increase_counter(state: State<'_, Mutex<AppState>>) -> Result<u32, ()> {
+  let mut state = state.lock().await;
+  state.counter += 1;
+  Ok(state.counter)
+}
+```
```
amrbashir: This section seems irrelevant as accessing state inside normal command vs async command, is basically the same.

We also shouldn't overload the readers with knowledge about a huge crate like tokio at least for explaining state management. Also you can get pretty far in Rust and Tauri without ever knowing anything about tokio.

I would remove this and just quickly mention an async example after the sync example.

src/content/docs/develop/state-management.mdx
```diff
@@ -2,32 +2,180 @@
...
+
+Note that the return type must be `Result` if you use asynchronous commands.
+
+### Outside of Commands
+
+Sometimes you may need to access the state outside of commands, such as in a different thread or in an event handler like `on_window_event`. In such cases, you can use the `state()` method of the `AppHandle` to get the state:
```
amrbashir: This is again partially true,  you can access the state through any type that implements the `Manager` trait

src/content/docs/develop/state-management.mdx
```diff
@@ -2,32 +2,180 @@
...
+    let state = app_handle.state::<Mutex<AppState>>();
 
-    // retrieve the store later
-    let store = app.state::<Store>();
+    // Lock the mutex to mutably access the state.
+    let mut state = state.lock().unwrap();
+    state.counter += 1;
+}
+
+// In `fn main()`:
```
amrbashir: just write the `fn main()`, would make the example more readable

src/content/docs/develop/state-management.mdx
```diff
@@ -2,32 +2,180 @@
...
+
+```rust
+use std::sync::Mutex;
+
+#[derive(Default)]
+struct AppStateInner {
+  counter: u32,
+}
+
+type AppState = Mutex<AppStateInner>;
```
amrbashir: Love this part, well done

src/content/docs/develop/state-management.mdx
```diff
@@ -2,32 +2,180 @@
...
 ```
 
-</Stub>
+However, make sure not to wrap it in a `Mutex` twice if you do this, otherwise you will run into the same issue.
```
amrbashir:
```suggestion
However, make sure to use the type alias as is, and not wrap it in a `Mutex` a second time, otherwise you will run into the same issue.
```

src/content/docs/develop/state-management.mdx
Rigidity: It's a common confusion

amrbashir: Alright how about this?

```suggestion
### Do you need `Arc`?

It's common to see `Arc` used in Rust to share ownership of a value across multiple threads usually paired with a `Mutex` in the form of `Arc<Mutex<T>>`. However, in Tauri, you don't need to use `Arc` because Tauri will wrap your type internally inside an `Arc` before storing it.
```

I dropped the following
> The only time in which it would make sense to use `Arc` for your state is when you need to share the state across multiple parts of your application that are not managed by Tauri itself.

Because places where you need to access the state across threads should be through cloning an app handle and moving it to that thread. If you need to use `Arc` manually then you probably shouldn't be storing it inside tauri state.

FabianLars: "However, in Tauri, you don't need to use `Arc` _**for State**_ because" (something like this, otherwise people _will_ misinterpret this as a general thing - source: already happened to me)

Rigidity: Thanks

Rigidity:
Sorry, been on a trip - will finish this in a few days

tweidinger (approved):

PR #2077, opened 2024-04-19
docs: fix command to create next app using bun

zeybek:
#### What kind of changes does this PR include?

- Minor content fixes (broken links, typos, etc.)
- Changes to the docs site code

#### Description
- What does this PR change? Give us a brief description.
  It will fix typo in command to create next app with bun 
- Did you change something visual?
  No

PR #2311, opened 2024-06-18
Typo change in Next.js getting started docs

quiint:
#### Description

Typo change - bunx create next-app to bunx create-next-app
image

### DIFF
--- a/docs/guides/getting-started/setup/next-js.mdx
+++ b/docs/guides/getting-started/setup/next-js.mdx
@@ -57,7 +57,7 @@ pnpm create next-app --use-pnpm
   <TabItem value="Bun">
 
 ```shell
-bunx create next-app --use-bun
+bunx create-next-app --use-bun
 ```
 
   </TabItem>


FabianLars:
Why does everyone say something different 😭 https://github.com/tauri-apps/tauri-docs/pull/2041 / https://github.com/tauri-apps/tauri-docs/pull/2077

quiint:
@FabianLars maybe it's on different operating systems? this is what it shows for my linux board:
image

the original issue came from my m1 mac running 1.0.36

simonhyll:
If ran with `bunx` it should be `bunx create-next-app`. If ran with `bun` it should be `bun create next-app`. Anything else is a user side issue

What `bun` does is check in npmjs for a package called `create-next-app`, which is made available in a prettier way by doing `bun create next-app`. This is the same way all JS package managers handle the create functionality

What `bunx` does is the same as e.g. `npx`, which is download the specified package, in other words `create-next-app`, and runs it.

So if anybody raises more PR's regarding "fixing" this, don't merge anything other than what's actually correct. If they're able to run `bunx create next-app` it's something weird on their system, it's actively against the official patterns and intended usages.

simonhyll (approved):
Using `bunx` this updated version is indeed the correct syntax. If anyone says otherwise, they're wrong.

PR #2316, opened 2024-06-19
Adding RPM doc

Sir-Thom:
#### Description
- Added to distribute directory  rpm.mdx
- Added section about how to sign manually

#### TODO

- [x] add section for how to sign with tauri 
- [x] make doc more "clean"
- [x] fixing grammar


src/content/docs/distribute/rpm.mdx
```diff
@@ -0,0 +1,397 @@
...
+  order: 1
+---
+
+import CommandTabs from '@components/CommandTabs.astro';
+
+# Packaging for RPM
+
+This guide covers how to distribute and manage RPM packages, including retrieving package information, configuring scripts, setting dependencies, and signing packages.
+
+## Configuring the RPM package
```
FabianLars, line 17: Can we add a note saying that the sections in here are optional, or at least those that are optional like the script? I'm afraid people would take it literally otherwise.

src/content/docs/distribute/rpm.mdx
```diff
@@ -0,0 +1,397 @@
...
+To add a license to the package, add the following to the `src-tauri/cargo.toml` file:
+
+```toml title="src-tauri/cargo.toml"
+[package]
+name = "tauri-app"
+version = "0.0.0"
+description = "A Tauri App"
+authors = ["you"]
+edition = "2021"
+license = "MIT" # add the license here
```
FabianLars, line 166: If we mention this, i think we should also mention bundle.license and/or bundle.licenseFile in tauri.conf.json 🤔 what do you think?

src/content/docs/distribute/rpm.mdx
```diff
@@ -0,0 +1,397 @@
...
+### Getting information about the package
+
+To get information about your package, such as the version, release, and architecture,
+use the following command:
+
+```bash
+rpm -qip package_name.rpm
+```
+
+### query specific information about the package
```
FabianLars:
```suggestion
### Query specific information about the package
```

FabianLars (changes requested):
Thanks for contributing! I annotated a few minor things. Also, could you remove the .idea files again?

src/content/docs/distribute/rpm.mdx
Sir-Thom: yeah that make sense to mention both  to show bundle.license  but for bundle.licenseFile  maybe but i do not know the difference with bundle.license. But  for what i understand this will find the license file if you have multiple config file like `tauri.linux.conf.json`.    I have included it  anyway in the rpm doc

FabianLars: bundle.license is the same as license in cargo.toml for those who don't want to touch the cargo.toml file.
bundle.licenseFile is a seperate license file that some bundles will include (not sure which ones, may be windows-only). What you did is good enough imo

FabianLars:
Thanks for applying my feedback. Another thing the changes made me notice is that as far as i know, our convention is to keep the tauri.conf.json minimal to only what's actual changed, for example like here https://v2.tauri.app/develop/resources/ (to keep pages a bit easier on the eyes)

FabianLars (approved):
thank you!

FabianLars:
don't mind the failing format check now. It can't handle mdx files very well. I'll fix it (not really) in another PR i'm already working on.

PR #2340, opened 2024-06-27
webdriver: WebKitWebDriver: set correct browsername to avoid `Failed to match capabilities`

bukowa:
fixes https://github.com/tauri-apps/tauri/issues/8828

This is really annoying, with previous `wry` example:

1. `Edgedriver` doesn't care.
2. `WebKitWebDriver` fails with `Failed to match capabilities`.
3. `Selenium` won't allow `undefined`.

P.S `WebKitWebDriver` can run with bundled `MiniBrowser` if you set `browserName` to `MiniBrowser`

### DIFF
--- a/docs/guides/testing/webdriver/example/selenium.md
+++ b/docs/guides/testing/webdriver/example/selenium.md
@@ -165,7 +165,7 @@ before(async function () {
 
   const capabilities = new Capabilities()
   capabilities.set('tauri:options', { application })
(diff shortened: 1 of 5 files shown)


bukowa:
This only happens in `v1` but it doesn't matter as `v1` won't work (prove me wrong) with newest webkitdriver anyway.
In `v2` `wry` is required to match capabilities.

PR #2349, opened 2024-06-29
Blog Post Sidebar Sorts By Date

jbolda:
## Description

Blog posts were sorting alphabetically in the sidebar partly due to a combination of plugins. This adjusts the config to use a sort and order in the `autogenerated` sidebar config.

_Note: It is currently using a patch to accomplish this. We may want to consider whether this is acceptable upstream, and push it up if so._


vasfvitor (approved):
let's use this while we find out if upstream will support it

PR #2377, opened 2024-07-07
AS-001: Updated next-js app create command for bunx

Asim2021:
The bunx craete-next-app is not a command.

#### Description

- Closes #

### DIFF
--- a/docs/guides/getting-started/setup/next-js.mdx
+++ b/docs/guides/getting-started/setup/next-js.mdx
@@ -57,7 +57,7 @@ pnpm create next-app --use-pnpm
   <TabItem value="Bun">
 
 ```shell
-bunx create-next-app --use-bun
+bunx create next-app --use-bun
 ```
 
   </TabItem>


FabianLars:
@simonhyll make it stop 😭

simonhyll:
Thanks for the PR @Asim2021 ! Unfortunately, the issues is in your end. The correct command is `bunx create-tauri-app`.

Long story short, what `bun create tauri-app` does is search for a package called `create-tauri-app`, downloads it then runs it, it's effectively an alias of `bunx create-tauri-app`. `bunx` is a utility that downloads packages by their exact name and runs them. So `bunx create-tauri-app` would download the `create-tauri-app` package and run it like a binary. There is no `create` package for it to download  that takes an argument of `tauri-app`.

If you are getting issues with running `bunx` I suggest you either upgrade your Bun version or try running the command in a different shell. If it still says the command isn't found then it's an issue in either your end, the shells end, or the Bun end, the command is correct as-is.

PR #2381, opened 2024-07-08
Adding an example of how to trigger a notification from rust code.

charlesschaefer:
#### Description

Adding an example of how to send notifications from rust code. The docs show how to do it from javascript, but was still lacking how to do it from rust side.


FabianLars (approved):
Thank you!

PR #2383, opened 2024-07-08
fix nextjs beforeBuildCommand

AurimarL:
#### Description
- What does this PR change? Nextjs use build and not generate,change to fix.


FabianLars (approved):
Thank you

PR #2384, opened 2024-07-09
Add Description To Store Plugin

jbolda:
#### Description

Clarify in the plugin `store` guide the use cases for this plugin.

### DIFF
--- a/src/content/docs/plugin/store.mdx
+++ b/src/content/docs/plugin/store.mdx
@@ -1,10 +1,6 @@
 ---
 title: Store
 description: Persistent key value storage.
-sidebar:
-  badge:
-    text: WIP
-    variant: caution
 ---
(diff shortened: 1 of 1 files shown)


src/content/docs/plugin/store.mdx
```diff
@@ -82,6 +80,8 @@ Install the store plugin to get started.
 
 ```js
 import { Store } from '@tauri-apps/plugin-store';
+// when using `"withGlobalTauri": true`, you may use
+// const { Store } = window.__TAURI_PLUGIN_STORE_;
```
jbolda:
```suggestion
// const { Store } = window.__TAURI_PLUGIN_STORE__;
```

simonhyll (approved):

PR #2385, opened 2024-07-09
[i18nIgnore] Import Examples When `"withGlobalTauri": true`

jbolda:
#### Description

To better support those venturing into Tauri with vanillajs, we include (commented out) import statements for each plugin. The imports are used from the `window` object instead of from the package, so this will improve visibility and discoverability.

Additionally, normalized on code blocks noting code type using `javascript` instead of `js` which improves searching for the code blocks while writing docs.


dreyfus92 (approved):
Thank you for this @jbolda, LGTM ✅

PR #2394, opened 2024-07-11
fix config plugin snipet

flapili:
fix config snipet for plugin config


dreyfus92 (approved):
Thank you @flapili, LGTM ✅

PR #2400, opened 2024-07-12
Add vscppdbg guide

Legend-Master:
#### Description

- Closes #


Legend-Master:
I'm terrible at writing really, please take a look and see if we can word this better 🙏

src/content/docs/develop/Debug/vscode.mdx
```diff
@@ -82,3 +84,69 @@ Note that it does not use the Tauri CLI, so exclusive CLI features are not execu
...
+      // (to debug release builds, change `target/debug` to `release/debug`)
+      "program": "${workspaceRoot}/src-tauri/target/debug/your-app-name-here.exe",
+      "cwd": "${workspaceRoot}",
+      "preLaunchTask": "ui:dev"
+    }
+  ]
+}
+```
+
+`tasks.json` is the same as with `lldb`, except you need to add a config group and target your `preLaunchTask` from `launch.json` to it if you want it to always compile before launching.
```
FabianLars: I'm not sure whether we should depend on the lldb section here or instead basically copy paste the description from there into here.

Legend-Master: I'm fine with both, maybe should paste it here since the reader might skip that section and going straight for vscppdbg instead

FabianLars (approved):

PR #2411, opened 2024-07-15
Update window customization guide

StepanZagray:
Guide had a mistake. To get appWindow you need to use `getCurrentWindow()` and not `new Window('main')`. I tested code with leptos and svelte frameworks. Both `import { getCurrentWindow } from '@tauri-apps/api/window';` and `const { Window } = window.__TAURI__.window;` worked with both frameworks. I want to note also that previous way of getting appWindow was working only with javascript frontend.

### DIFF
--- a/src/content/docs/plugin/window-customization.mdx
+++ b/src/content/docs/plugin/window-customization.mdx
@@ -143,7 +143,7 @@ import { getCurrentWindow } from '@tauri-apps/api/window';
 // when using `"withGlobalTauri": true`, you may use
 // const { getCurrentWindow } = window.__TAURI__.window;
 
-const appWindow = getCurrentWindow()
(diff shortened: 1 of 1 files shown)


src/content/docs/plugin/window-customization.mdx
```diff
@@ -155,6 +156,8 @@ document
   ?.addEventListener('click', () => appWindow.close());
 ```
 
+Note that if you are using rust for frontend, you can copy code above into `<script>` tag into your `index.html`.
```
FabianLars:
```suggestion
Note that if you are using a Rust-based frontend, you can copy the code above into a `<script>` element in your `index.html` file.
```

FabianLars (approved):
That was quick, Thank you very much!

PR #2413, opened 2024-07-15
Nix/NixOS Requirements and Documentation

### LINKED ISSUE #8588: [feat] upstream nix requirements to nixpkgs
### Describe the problem

Not sure if this is a bug or a feature, it's borderline imo.

Basically what I expect is that when i use the `cargo-tauri` derivation from `nixpkgs` that i don't also have to copy and paste a bunch of additional prerequisites from the tauri docs in order to get things half working.

I say half working because I'm still running into bugs with `cargo-tauri`, especially on mac, one of which i already pushed a fix for https://github.com/NixOS/nixpkgs/pull/279771#event-11446551011

I also see that one of the dependencies, `webkitgtk` is marked as broken in `nixpkgs` on mac, but listed as a dependency on the tauri docs. I understand the nix docs say nixos rather than `nix` but it's pretty common use case for people working on linux/mac to be using nix rather than committing to a full nixos setup.

Additionally, I noticed that the prerequisites in the docs only work for a `nix develop` style shell, they don't work for `nix run`, say if I wanted to lift some standard tauri tasks into a package, the instructions don't cover that. This is because the additional steps of setting up the library etc. are limited to a `shellHook`.

Further, the prerequisites are missing required steps in the `shellHook`. We found that when tauri is built via nix there is a need to disable webkit compositing mode like `export WEBKIT_DISABLE_COMPOSITING_MODE=1` in the `shellHook`. This seems to be some known tribal knowledge, as it appears in snippets from other people such as https://github.com/tauri-apps/tauri/issues/8535 but it's not listed in the prerequisites in the docs.

There are other issues like https://github.com/tauri-apps/tauri/issues/8535 that I expect are probably not issues with tauri itself, but some additional work that needs to be done on the nix side. At the moment it isn't clear where to put or discuss such a fix, would it just result in more documented boilerplate? or is there a way to codify it?

Stepping back a bit, I feel like the meta issue is that fixes to issues are being pushed into the tauri docs (or not) rather than some derivation that is directly maintainable and reusable downstream.

### Describe the solution you'd like

Tauri maintains the derivation in `nixpkgs` with the same dependencies that it outlines in the documentation, so then there are no manual steps for consumers.

### Alternatives considered

Another alternative would be that tauri maintains their own nix code separate to nixpkgs, such as an overlay or flake.

This would be fine too, the main problem to solve is to lift dependencies and other issues from docs and into nix-compatible code somewhere.

### Additional context

_No response_

Eveeifyeve:
#### Description
**Note This Is English Only! Translation will be needed for i18n support**

This PR Adds the necessary requirements for [Nix(Package Manager) & NixOS(Linux Distro)](https://nixos.org)

- Closes https://github.com/tauri-apps/tauri/issues/8588


Eveeifyeve:
continue more tomorrow.

Eveeifyeve:
Working on a builder for tauri at https://github.com/eveeifyeve/buildtauriapp.

Eveeifyeve:
So I am working on the builder to make it easy to build tauri apps

Eveeifyeve:
Working on converting it all to stdenv from rustplatform.

Eveeifyeve:
In my roadmap https://github.com/Eveeifyeve/buildTauriApp/issues/1

Eveeifyeve:
today I am going to push more documentation and see if I can get it working on my end and provide a build template to get started to build tauri for desktop only I haven't gotten around mobile.

Eveeifyeve:
I am looking to add more documentation about packaging for nix/nixos using the hook: https://github.com/NixOS/nixpkgs/pull/335751

Eveeifyeve:
Is it okay to rebase @simonhyll to fix this merge confict?

FabianLars:
Fix it however you prefer. rebase or "normal" merge, we don't really care :)

Eveeifyeve:
I am waiting for this: https://github.com/NixOS/nixpkgs/pull/335751 to be on all unstable channels before I make docs about building for nix/nixos

Eveeifyeve:
I am rethinking of this pr I am going to do a nix overhaul.

Eveeifyeve:
I think it's ready it, but the android and ios parts haven't been tested.

src/content/docs/distribute/index.mdx
getchoo, line 1: Not entirely sure why this is here. This hook isn't a tool for developers to distribute their Tauri apps -- in fact the binaries built with this will not work on other platforms -- it is only meant for packagers in nixpkgs

src/content/docs/start/prerequisites.mdx
```diff
@@ -28,6 +28,10 @@ Tauri requires various system dependencies for development on Linux. These may b
 {/* Note: These are the officially supported linux distributions. */}
 {/* If you wish to add another please open an issue to discuss prior to opening a PR */}
 
+:::note
+Using nix/nixos? we have moved nix/nixos to a dedicated section [here](#nixnixos)
```
getchoo:
```suggestion
Using Nix/NixOS? we have moved Nix/NixOS to a dedicated section [here](#nixnixos)
```

src/content/docs/start/prerequisites.mdx
```diff
@@ -202,6 +181,34 @@ Install WebView2 by visiting the [WebView2 Runtime download section](https://dev
...
+  buildInputs = [
+    libiconv
+  ] ++ lib.optionals stdenv.isLinux [ gtk3 libsoup_3 webkitgtk_4_1] ++
+	lib.optionals stdenv.isDarwin [
+		# Only required if your using the npm package `@tauri-apps/cli` inside of your `package.json`
+		darwin.apple_sdk.frameworks.WebKit
+		# cargo-tauri # only if you want to cargo-tauri 
+	];
+}
+```
```
getchoo: Fixed a few problems here:

- This is Nix code, not `sh`
- We should be using a [reproducible nixpkgs configuration](https://nix.dev/guides/best-practices#reproducible-nixpkgs-configuration)
- `gtk3` is for Tauri v1, not v2
- `libsoup3` is no longer used
- `glib-networking` is required for a vast majority of apps
- The comment on only requiring the WebKit framework when using the prebuilt binary from NPM is misleading. All Tauri projects need it to build
- `libiconv` isn't required on macOS or Linux to my knowledge -- unless this was introduced in Tauri v2
- `pkg-config` was missing
- A Rust toolchain should be included so you can actually build an app
- `openssl` was missing

The comment next to `cargo-tauri` has also been clarified. It isn't required per-se since it seems the pre-built NPM binary for Linux is completely static, but if this ever changes NixOS *will* have issues

> Downloading and attempting to run a binary on NixOS will almost never work. This is due to hard-coded paths in the executable.
>
> -https://wiki.nixos.org/wiki/Packaging/Binaries

```suggestion
```nix
let
  pkgs = import <nixpkgs> {
    config = { };
    overlays = [ ];
  };
in

pkgs.mkShell {
  nativeBuildInputs =
    with pkgs;
    [
      rustc
      cargo
      # Uncomment if `@tauri-apps/cli` doesn't work on your system
      # cargo-tauri
      pkg-config
    ]
    ++ lib.optionalAttrs stdenv.hostPlatform.isLinux [ wrapGAppsHook4 ];

  buildInputs =
    with pkgs;
    [ openssl ]
    ++ lib.optionals stdenv.hostPlatform.isLinux [
      glib-networking
      webkitgtk_4_1
    ]
    ++ lib.optionals stdenv.hostPlatform.isDarwin [
      darwin.apple_sdk.frameworks.WebKit
    ];
}
```
```

File: src/content/docs/start/prerequisites.mdx [hunk 3]
```diff
@@ -202,6 +181,34 @@ Install WebView2 by visiting the [WebView2 Runtime download section](https://dev
...
+
+:::caution[Experimental]
+Just to note flakes are experimental in nix and may change any time.
```
getchoo:
```suggestion
Flakes are [experimental](https://nix.dev/manual/nix/stable/contributing/experimental-features.html) in Nix and may change at any time.
```

src/content/docs/start/prerequisites.mdx
```diff
@@ -202,6 +181,34 @@ Install WebView2 by visiting the [WebView2 Runtime download section](https://dev
...
+:::
+
+Using `flake.nix`: checkout [eveeifyeve/Tauri-Flake-Templates](https://github.com/Eveeifyeve/Tauri-Flake-Templates) on github to use a community maintained nix flake.
```
getchoo:
```suggestion
If you prefer to use Flakes, [Eveeifyeve/Tauri-Flake-Templates](https://github.com/Eveeifyeve/Tauri-Flake-Templates) on GitHub is a community maintained template.
```

src/content/docs/start/prerequisites.mdx
```diff
@@ -202,6 +181,34 @@ Install WebView2 by visiting the [WebView2 Runtime download section](https://dev
...
+
+Feel free to follow along if your using the flake template as this is the same thing
+
+Next: [Install Rust](#rust)
+
```
getchoo: This doesn't make sense...the template should have everything ready, including Rust and Tauri's dependencies
```suggestion
```

src/content/docs/start/prerequisites.mdx
```diff
@@ -227,6 +234,16 @@ Here is the file as a plain script: [rustup.sh](https://sh.rustup.rs/)
...
+  <TabItem label="Nix/NixOS"> 
+  Add the following in your buildinputs from [Previous Section](#nixnixos) below to install Rust on nix/NixOS.
+  ```nix
+  cargo
+  rustc
+  rustfmt
+  ```
+
+</TabItem>
+
```
getchoo: Again, we should just have Rust in the above dev shell and Flake template. It makes no sense to do this manually when every Tauri projects requires it

This is also *very* confusing since the previous section says to follow along with this guide, but this is specifically only referring to the shell.nix above and *not* the Flake?

```suggestion
```

src/content/docs/start/prerequisites.mdx
```diff
@@ -239,6 +256,8 @@ Next: [Configure for Mobile Targets](#configure-for-mobile-targets) if you'd lik
 Only if you intend to use a JavaScript frontend framework
 :::
 
+<Tabs>
+<TabItem label="Linux/macOS/Windows">
```
getchoo:
```suggestion
```

src/content/docs/start/prerequisites.mdx
```diff
@@ -254,6 +273,19 @@ It's important to restart your Terminal to ensure it recognizes the new installa
...
+  ```nix
+  nodejs # feel free to change the version
+  npm
+  ```
+
+    While npm is the default package manager for Node.js, you can also use others like pnpm or yarn. To install them look for that package on [nixpkgs](https://search.nixos.org/packages) and add the package above.  This step is optional and only needed if you prefer using a package manager other than npm.
+
+  </TabItem>
+</Tabs>
+
```
getchoo: I don't think this is worthy of a dedicated section. It would be much better to just have comments in the dev shell shown above again, as projects will always need a version of Nodejs and a package manager
```suggestion
```

src/content/docs/start/prerequisites.mdx
```diff
@@ -265,7 +297,13 @@ If you'd like to target your app for Android or iOS then there are a few additio
...
 ### Android
 
-1. Download and install [Android Studio from the Android Developers website](https://developer.android.com/studio)
+:::caution[Nix/NixOS]
+Android Studio only supports Linux Nix/NixOS.
+If you are using Nix on darwin you could try the [homebrew route through nix-darwin](https://daiderd.com/nix-darwin/manual/index.html#opt-homebrew.brews) to install Android Studio.
+:::
+
+1. Download and install [Android Studio from the Android Developers website](https://developer.android.com/studio) or if you are using Nix/NixOS by adding the package to your system `android-studio`.
+
```
getchoo: This *really* doesn't need Nix/NixOS specific instructions. If you're using either, I am sure you know how to install a package. We don't need to explain that here

src/content/docs/start/prerequisites.mdx
```diff
@@ -301,7 +339,7 @@ export JAVA_HOME="/Applications/Android Studio.app/Contents/jbr/Contents/Home"
 - Android SDK Build-Tools
 - Android SDK Command-line Tools
 
-Selecting "Show Package Details" in the SDK Manager enables the installation of older package versions. Only install older versions if necessary, as they may introduce compatibility issues or security risks.
+/electing "Show Package Details" in the SDK Manager enables the installation of older package versions. Only install older versions if necessary, as they may introduce compatibility issues or security risks.
```
getchoo:
```suggestion
Selecting "Show Package Details" in the SDK Manager enables the installation of older package versions. Only install older versions if necessary, as they may introduce compatibility issues or security risks.
```

src/content/docs/start/prerequisites.mdx
```diff
@@ -340,6 +377,15 @@ $VERSION = Get-ChildItem -Name "$env:LocalAppData\Android\Sdk\ndk"
...
 
+# Nix/Nixos
+
+Add the following to `buildInputs` to install the tools/Android Studio(Linux only):
+
+```nix
+android-studio # Linux only
+android-tools
+```
+
```
getchoo: See above

I also don't think this is the only thing that's required? The regular instructions tell you to add multiple targets to rustup for cross compilation; only installing Android Studio does not do this.
```suggestion
```

src/content/docs/start/prerequisites.mdx
```diff
@@ -372,6 +420,17 @@ Next: [Create a project](/start/create-project/).
 
 If you run into any issues during installation be sure to check the [Troubleshooting Guide](/develop/debug/) or reach out on the [Tauri Discord](https://discord.com/invite/tauri).
 
+</TabItem>
+<TabItem label="Nix">
+Add the following to `buildInputs` under darwin section to install
```
getchoo: This section is a *but* more informative since the docs don't already mention a specific package to install for Nix,, so it's fine IMO

```suggestion
Add the following to your development shell's `buildInputs`
```

src/content/docs/start/prerequisites.mdx
```diff
@@ -154,6 +130,10 @@ Next: [Install Rust](#rust)
 
 ### macOS
 
+:::note
+Using Nix? we have a dedicated section on getting started on nix [here](#nixnixos)
```
getchoo:
```suggestion
Using Nix? We now have a dedicated section on getting started with it [here](#nixnixos)
```

src/content/docs/start/prerequisites.mdx
```diff
@@ -202,6 +181,34 @@ Install WebView2 by visiting the [WebView2 Runtime download section](https://dev
 
 Next: [Install Rust](#rust)
 
+### Nix/NixOS
+
+Using `nix-shell`:
```
getchoo:
```suggestion
Using a [`shell.nix`](https://nix.dev/tutorials/first-steps/declarative-shell):
```

getchoo (changes requested):
Some of this is really good, but a lot of it feels like overkill. Many of the instructions here already apply to Nix/NixOS if you know how to use it (like what dependencies to install, how to install them, etc) and others are so exclusive to NixOS that I don't think upstream documentation is the place for it

I would recommend moving much of this to wiki.nixos.org, as it will have a lot more support and recognition from its target audience there. Nix/NixOS should be treated mostly the same as other distros here IMO

src/content/docs/start/prerequisites.mdx
Eveeifyeve: Ops my mistake.

src/content/docs/distribute/index.mdx
Eveeifyeve: Well it's for Distributing to to the nix ecosystem.

src/content/docs/start/prerequisites.mdx
Eveeifyeve: Rust by default adds these targets but It seems like android studio is not required in this case.

In nixpkgs there is something called `androidenv` which is under `pkgs/development/mobile/androidenv` which has the required stuff needed.

src/content/docs/start/prerequisites.mdx
Eveeifyeve: I guess.

src/content/docs/start/prerequisites.mdx
getchoo: > Rust by default adds these targets

It does not. These are cross compile targets, which you would need to use `pkgsCross` for

> In nixpkgs there is something called androidenv which is under pkgs/development/mobile/androidenv which has the required stuff needed.

It supplies the Android SDK and NDK. Nothing else

src/content/docs/distribute/index.mdx
getchoo: Then it should probably only be documented in the Nix ecosystem. It doesn't make sense to include documentation on building .deb packages here for example, as that isn't meant for a whole platform. This would be something for wiki.nixos.org and the nixpkgs manual

src/content/docs/start/prerequisites.mdx
Eveeifyeve: Which is needed for to build android?

src/content/docs/distribute/index.mdx
Eveeifyeve: Well initially I was going to add a link directly to nixpkgs for distrobuting to nixpkgs but it seems like astro is limiting me to do that because of the autogenerated directory’s.

src/content/docs/start/prerequisites.mdx
getchoo: The Android SDK, NDK, and Rust cross compilation targets are all required

Eveeifyeve: Which androidenv supplies with SDK & NDK is there but adding the targets would a. require something like fenix or b. require rustup.

getchoo: Which is why I said

> I also don't think this is the only thing that's required?

But again, this section can probably just be removed. Nix(OS) users should know how to install Android Studio and add Rust targets. They can also follow along with the rustup instructions verbatim anyways

src/content/docs/distribute/index.mdx
Eveeifyeve: > Well initially I was going to add a link directly to nixpkgs for distrobuting to nixpkgs but it seems like astro is limiting me to do that because of the autogenerated directory’s.

@FabianLars or @simonhyll Do you know perhaps?

src/content/docs/start/prerequisites.mdx
Eveeifyeve: True, I guess your right.

src/content/docs/start/prerequisites.mdx
Eveeifyeve: Your right, Resolved.

src/content/docs/distribute/index.mdx
Eveeifyeve: It's not really needed since I am going to link to a wiki page for more info about tauri on nix/nixos.

Eveeifyeve:
Also would it be okay to squash the commits in this pr?

FabianLars:
We always squash merge (almost) all PRs :)

Eveeifyeve:
> We always squash merge (almost) all PRs :)

Okay good cause this pr has too many commits.

src/content/docs/start/prerequisites.mdx
```diff
@@ -202,6 +181,51 @@ Install WebView2 by visiting the [WebView2 Runtime download section](https://dev
...
+      cargo
+      # Uncomment if `@tauri-apps/cli` doesn't work on your system
+      # cargo-tauri
+      pkg-config
+    ]
+    ++ lib.optionalAttrs stdenv.hostPlatform.isLinux [ wrapGAppsHook4 ];
+
+  buildInputs =
+    with pkgs;
+    [ openssl ]
```
Eveeifyeve, line 201: @getchoo for some reason `libiconv` is required maybe for plugins?

getchoo: What plugins specifically? It'd be nice to note it

I also imagine that would be a macOS exclusive dependency; libiconv is hardly ever used on Linux

Eveeifyeve: I would have to test to see which plugins require it.

Eveeifyeve: I will investigate and create a minimal environment.

vasfvitor:
@Eveeifyeve is this ready?

src/content/docs/start/prerequisites.mdx
```diff
@@ -115,43 +115,7 @@ sudo apk add \
...
-    librsvg
-    libsoup_3
-    pango
-    webkitgtk_4_1
-    openssl
-  ];
-}
-```
-
+    This section has been moved Nix/NixOS to a dedicated section [here](#nixnixos)
```
vasfvitor, line 118: I'd prefer if you wrap it into a :::info aside, as of now it kinda mixed with the text content right after it. Ofc if you have a better idea than aside it could work.
image

Eveeifyeve:
> @Eveeifyeve is this ready?

This is not finished as I got to resolve https://github.com/tauri-apps/tauri-docs/pull/2413#discussion_r1788461138.

I am also going to mark this pr as a draft in a sec because of this.

src/content/docs/start/prerequisites.mdx
Eveeifyeve: I will test it once I get a new mac since my system is screwed by liquid damage luckily I use nix so no problems of reinstall apps.

src/content/docs/start/prerequisites.mdx
Eveeifyeve: I will do this once I clone it today, I forgotten this pr by accident.

PR #2434, opened 2024-07-19
feat: automations for project management

simonhyll:
> There seems to be an authentication issue for the workflow. When executed locally it works fine. May require a PAT with more permissions than what the workflow permissions can grant.

## Assign

Auto assigns PR's to the one that opened the PR. We use assigning primarily to indicate which people are intended to be actively working on pushing code to resolve something, which includes the person that opened the PR. The reason why we assign the author is also because it's an indication of whether the author is still interested in pushing the issue across the finish line, or if they unassign themselves it tells us they will no longer be actively working on finishing their work.

## Status

Automates handling of the `Status` field by checking the following:

**Issues:**
- Backlog: upstream or discussion is set, milestone missing, no labels set
- Ready: has labels and milestone
- In progress: Is assigned
- In review: Has a linked PR that's merged

**PRs:**
- In progress: Is in draft mode or not in a mergeable state
- In review: Open and mergeable

## Priority

Simply adds the `low` priority when issues or PR's are opened. Higher priorities are set manually for the time being. Potential future improvement is increasing priority based on how long the issue/pr has been opened.

### DIFF
--- a/.github/workflows/assign.yml
+++ b/.github/workflows/assign.yml
@@ -1,0 +1,21 @@
+name: 'Auto Assign'
+
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+  auto_assign:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Auto-assign PR to author
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          pr_number=${{ github.event.pull_request.number }}
+          pr_author=${{ github.event.pull_request.user.login }}
+          gh api -X POST \
+            -H "Accept: application/vnd.github+json" \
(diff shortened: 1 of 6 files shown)


vasfvitor:
> There seems to be an authentication issue for the workflow. When executed locally it works fine. May require a PAT with more permissions than what the workflow permissions can grant.

so the failed status is because of this?

simonhyll:
> > There seems to be an authentication issue for the workflow. When executed locally it works fine. May require a PAT with more permissions than what the workflow permissions can grant.
> 
> so the failed status is because of this?

Yea. After talking to Jacob I think what I'll do is just merge this and see if that makes it work, since the issue is likely due to me working in a fork which doesnt get the proper permissions.

Worst case scenario I'll just undo it and fix whatever statuses get set incorrectly

PR #2440, opened 2024-07-20
Updating dialog documentation to list the supported platforms

charlesschaefer:
#### Description

A small patch to list the supported platforms in the dialog documentation.


charlesschaefer:
Closing the pull request because I just realized that maybe it is not all the dialog plugin that isn't supported on Android, but maybe only the save() dialog.

PR #2449, opened 2024-07-22
Add Explanation on Dev Server Security

tillmann-crabnebula:
This adds a section to the lifecycle security explaining how development servers for the frontend can be insecure and that the default vanilla dev server has no authentication or encryption.

It is also mentioned in the develop section.


chippers (approved):
i have no problems with the wording

PR #2467, opened 2024-07-27
feat: update configuration files documentation

lucasfernog:


src/content/docs/develop/configuration-files.mdx
```diff
@@ -0,0 +1,166 @@
+---
+title: Configuration Files
+sidebar:
+  order: 1
+---
+
+Since Tauri is a toolkit for building applications there can be many files to configure project settings. Some common files that you may run across are `tauri.conf.json`, `package.json` and `Cargo.toml`. We briefly explain each on this page to help point you in the right direction for which file to modify.
```
vasfvitor:
```suggestion
Since Tauri is a toolkit for building applications there can be many files to configure project settings. Some common files that you may run across are `tauri.conf.json`, `package.json` and `Cargo.toml`. We briefly explain each on this page to help point you in the right direction for which files to modify.
```

src/content/docs/develop/configuration-files.mdx
```diff
@@ -0,0 +1,166 @@
...
+
+[[app.windows]]
+title = "MyApp"
+
+[plugins.updater]
+pubkey = "updater pub key"
+endpoints = ["https://my.app.updater/{{target}}/{{current_version}}"]
+```
+
+Note that JSON5 and TOML supports comments and TOML can use kebab-case for config names with are more idiomatic.
```
vasfvitor:
```suggestion
Note that JSON5 and TOML supports comments, and TOML can use kebab-case for config names which are more idiomatic.
```

vasfvitor (approved):
Note that I used a diff tool as GitHub shows that everything was changed, which is not the case.

Quickly revised only grammar and readability and other than that it's ok

src/content/docs/develop/configuration-files.mdx
```diff
@@ -0,0 +1,166 @@
...
+[build-dependencies]
+tauri-build = { version = "2.0.0-beta", features = [ "config-json5" ] }
+
+[dependencies]
+tauri = { version = "2.0.0-beta", features = [  "config-json5" ] }
+```
+
+The structure and values are the same across all formats, however, the formatting should be consistent with the respective file's format:
+
+```json5 title=tauri.conf.json or tauri.conf.json5
```
tweidinger, line 29: It should only be `tauri.conf.json5` or we should remove comments to make it `.json` compatible.

lucasfernog: we also support json5 in a .json file (though maybe that doesn't even need to be explicit here to avoid confusion)

tweidinger (approved):

PR #2468, opened 2024-07-27
feat: fill calling-frontend documentation

lucasfernog:


src/content/docs/develop/calling-frontend.mdx
```diff
@@ -2,11 +2,387 @@
...
-{/* TODO: Add Events Page from v1 */}
-{/* TODO: Show rust->rust events somewhere */}
-{/* TODO: Also mention window.eval() (the rust method) */}
+The Rust side of your Tauri application can call the frontend by leveraging the Tauri event system,
+using channels or directly evaluating JavaScript code.
+
+## Event System
+
+Tauri ships a simple event system you can use to communicate between Rust and your frontend.
+The [AppHandle] and [WebviewWindow] types all implement the event system traits [Listener] and [Emitter].
```
tweidinger: We should add some paragraph to make a distinction between commands and events.

Something like:

```suggestion
Tauri ships a simple event system you can use to (bi-directional) communicate between Rust and your frontend.
The [AppHandle] and [WebviewWindow] types all implement the event system traits [Listener] and [Emitter].

The event system was designed for situations where small amounts of data need to be streamed or you need to implement a multi consumer multi producer pattern (e.g. push notification system).
The event system is not designed for low latency or high throughput situations.

The major differences between a [Tauri command] and a Tauri event are that events have no strong type support, event payloads are always json (de)serialized making them not suitable for bigger messages and there is no support of the [capability] system to fine grain control event data and channels.
```

tweidinger (approved):

src/content/docs/develop/calling-frontend.mdx
```diff
@@ -2,11 +2,404 @@
...
+const appWebview = getCurrentWebviewWindow();
+appWebview.once('ready', () => {});
+```
+
+:::note
+Events emitted in the frontend also triggers listeners registed by these APIs.
+For more information see the [Calling Rust from the Frontend] documentation.
+:::
+
+#### Listening to Events on Rust
```
FabianLars, line 208: Someone wanna bet on how long it'll take for someone to interpret this as instructions for a Rust frontend? :D

lucasfernog: indeed we should enhance the wording here

PR #2471, opened 2024-07-28
[i18n-ignore] show version on release cards

vasfvitor:
just to see where this could go: showing version of package

one possible issue is that it's not always the latest as may be a delay between the release and docs update. But could be labelled as "up to: v1.0.0" or something
image


FabianLars (approved):

dreyfus92 (approved):
LGTM ✅

PR #2514, opened 2024-08-08
ci: Fix token permissions

FabianLars:
Gonna triage it myself too now. If i can't make it work with GITHUB_TOKEN or the bot pat, i'll create a pat on my account for triage so we know what to change in the bot token.

Please don't merge this one (i'll do it myself when i'm done)


FabianLars:
I'll have to merge this to test if it works for forks (because of how pull_request_target works)

PR #2517, opened 2024-08-08
temp: Test status ci from fork

FabianLars:


FabianLars:
ayo can't believe it actually (almost) worked - don't ping me for gh actions stuff in the future because of this though 😂

FabianLars:
should all work now (don't want to create a new pr so i'll wait xd)

PR #2519, opened 2024-08-09
feat: add collapse to the table of content items

ahkohd:
#### Description

- Closes #

### DIFF
--- a/astro.config.mjs
+++ b/astro.config.mjs
@@ -96,6 +96,7 @@ export default defineConfig({
         Footer: 'src/components/overrides/Footer.astro',
         ThemeSelect: 'src/components/overrides/ThemeSelect.astro',
         PageFrame: 'src/components/overrides/PageFrame.astro',
+        TableOfContents: 'src/components/overrides/TableOfContents.astro',
       },
       head: [
         {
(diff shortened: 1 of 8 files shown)


lucasfernog:
can we automatically collapse above the max heading level (instead of removing them from the list)? would make this feature neat

simonhyll (approved):
This is definitely an improvement! 🥳 We might wanna consider adding an option to the frontmatter of pages so that we can configure a `tableOfContents.collapseLevel` on a page by page basis, because I think on most pages it'll work great with default collapsing above the second level, but on e.g. the config reference pages we'd likely want to collapse already at the top level, at least in my opinion

But that's a bunch of extra effort, so I'm totally fine with merging as-is, since it's effectively currently not changing the appearance, it just adds a cool feature

Good job! 🙂

ahkohd:
> This is definitely an improvement! 🥳 We might wanna consider adding an option to the frontmatter of pages so that we can configure a `tableOfContents.collapseLevel` on a page by page basis, because I think on most pages it'll work great with default collapsing above the second level, but on e.g. the config reference pages we'd likely want to collapse already at the top level, at least in my opinion
> 
> But that's a bunch of extra effort, so I'm totally fine with merging as-is, since it's effectively currently not changing the appearance, it just adds a cool feature
> 
> Good job! 🙂

Good day,

I have added support for configuring the collapse level via the document's frontmatter.

By default the collapse level is set to 1, counting starts from 0.

Commit 4476a67b40c36232b462fbff38b44afbfe86e3e3 shows an example usage.

PR #2525, opened 2024-08-09
Docs/develop in flatpak sandbox

DadiBit:
#### Description
- What does this PR change? Give us a brief description. <br> This adds the two steps required to develop and build a tauri app within a Flatpak IDE (or sandboxed app more generically)
- Closes #


FabianLars:
Hi, thanks for the PR! I think this needs more explanation (context of when this is needed for example). I only know why you need it because I received a mail notification for your stack overflow question 😅

Edit: I'm on the road rn so can't give proper suggestions for a while, sry

DadiBit:
> Hi, thanks for the PR! I think this needs more explanation (context of when this is needed for example).

I actually avoided the VSCodium example to avoid "advertising", however, if you consider it could be beneficial I can update the docs in little time.

I usually run it like this: `flatpak run --runtime=org.gnome.Sdk//45 com.vscodium.codium` 

> I only know why you need it because I received a mail notification for your stack overflow question 😅

That's wild! :laughing: 

As a sidenote: you must also install the node.js SDK and enable it (not straightforward), however, node.js is not present in the prerequisites of other environments.

Perhaps I could write a dedicated paragraph instead of a tab near Debian/Fedora/etc. Wdyt?

PR #2526, opened 2024-08-09
:memo: Update haptics and geolocation plugins link

homersimpsons:
#### Description

Fix links for those plugins. They were pointing to an inexistent page.


FabianLars:
Ahh good catch, the blog post was drafted before I upstreamed the plugins in the main repo (the links were 404 before that too though - right branches, wrong repo).

Thank you!

PR #2535, opened 2024-08-11
feat: autogenerate CLI command docs

### LINKED ISSUE #1985: Need generator for CLI reference page
Need a generator for CLI reference pages to prevent missing commands. Currently CLI reference pages are static pages and will prone to error as Tauri updates.

lucasfernog:


lucasfernog:
Due to https://github.com/pnpm/pnpm/issues/7002 it's tricky to automatically update the CLI before generating docs.. I suggest we let renovate handle that for us for now.

packages/cli-generator/build.ts
```diff
@@ -0,0 +1,100 @@
...
+    if (line === 'Commands:') {
+      readingSubcommands = true;
+    }
+  }
+
+  return subcommands;
+}
+
+function generateCommandDoc(command: string, level: number, subcommandList: Command[]): string {
+  const output = execSync(`pnpm tauri ${command} --help`)
```
tweidinger, line 35: This approach is kinda fragile whenever we change the layout of the help output but I don't think that will happen more than once.
Also running the binary just for the help output sounds weird when we have access to the docs/source code itself since we have tauri as a git submodule.

Is tauri auto generating the rustdocs/scheme and we can consume them the same way we do for the permissions and configs? If not is it not feasible to do so?

tweidinger (approved):

PR #2536, opened 2024-08-12
Add SolidStart to list of Frontend guides

birkskyum:
This PR adds SolidStart to the list of framework guides.

Screenshot 2024-08-12 at 16 00 04


tweidinger (changes requested):
Hey, thanks for the incentive. While I think Solid is a good thing to have in our docs the document does not contain any direct instructions or guides to set up the configuration etc. which the other frontend guides contain.

I don't think we should have our own page unless it contains all necessary information to get started.

birkskyum:
Okay, keeping it consistent makes sense. cc @atilafassina

PR #2590, opened 2024-08-26
chore: Clarify FS Plugin Rust Side Exposure

### LINKED ISSUE #2567: [request] File System guide missing rust instructions
### Question you want answered

Do i need the fs plugin on the rust side?

### Where did you look for an answer?

in our docs lol

Next time i step around the issue template again 😂 

Anyway, we just need to document at the top or somewhere that you should use std or tokio on the rust side, and if you don't need the js apis then don't add the plugin at all.

### Page URL

https://v2.tauri.app/plugin/file-system/

### Additional context

https://discord.com/channels/616186924390023171/1275176522885627929

### Are you willing to work on this yourself?

- [ ] I want to work on this myself

tweidinger:
Fixes #2567 by adding an explanation note to the plugin usage docs.

### DIFF
--- a/src/content/docs/plugin/file-system.mdx
+++ b/src/content/docs/plugin/file-system.mdx
@@ -86,8 +86,8 @@ The fs plugin is available in both JavaScript and Rust.
 
 :::note[Rust Exposure]
 
-The exposed API on the Rust side is only intended to manipulate the [Scope](#scopes) of the plugin but
-does not expose any functional API to access the file system. Please use [`tokio::fs`](https://docs.rs/tokio/latest/tokio/fs/index.html) or the
+The exposed API on the Rust side is mostly intended to manipulate the [Scope](#scopes) of the plugin.
+It does expose some APIs to access the file system on mobile, but the recommendation is to use [`tokio::fs`](https://docs.rs/tokio/latest/tokio/fs/index.html) or the
 [`std::fs`](https://doc.rust-lang.org/std/fs/index.html) for accessing the file system.
 
 :::
--- a/src/content/docs/plugin/file-system.mdx
+++ b/src/content/docs/plugin/file-system.mdx
@@ -84,6 +84,14 @@ Install the fs plugin to get started.
 
 The fs plugin is available in both JavaScript and Rust.
 
+:::note[Rust Exposure]
+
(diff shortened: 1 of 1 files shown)


src/content/docs/plugin/file-system.mdx
```diff
@@ -84,6 +84,14 @@ Install the fs plugin to get started.
 
 The fs plugin is available in both JavaScript and Rust.
 
+:::note[Rust Exposure]
```
lucasfernog, line 87: on the next release we'll include a few APIs for Android support https://github.com/tauri-apps/plugins-workspace/pull/1658
for now we have fs open(), read() and read_to_string() iirc - but we'll add more
this is important to resolve content URIs (returned by the dialog open() and save() APIs for instance) and resources on Android

tweidinger: Added a mention that there is some exposure for mobile.

src/content/docs/plugin/file-system.mdx
```diff
@@ -84,6 +84,14 @@ Install the fs plugin to get started.
...
+
+The exposed API on the Rust side is mostly intended to manipulate the [Scope](#scopes) of the plugin.
+It does expose some APIs to access the file system on mobile, but the recommendation is to use [`tokio::fs`](https://docs.rs/tokio/latest/tokio/fs/index.html) or the
```
lucasfernog, line 90: For Android the tokio or std APIs won't work if you're using e.g. the dialog API (which returns a content:// URI instead of a path to the file). We might also need more changes for iOS to properly request for permissions to access some files (for instance photos, or the Files location).
I think in the end when the FS plugin is marked as stable, we'll recommend using it even for Rust code unless your app is desktop-only.

tweidinger: :flip-table: 😂 
Alright, so we recommend std/tokio for desktop and our api for mobile?

lucasfernog-crabnebula: we'll fallback to std for desktop, so if you need cross platform support for e.g. reading resources and dialog-selected files, you need to use our own API wrapper

lucasfernog-crabnebula: so I think we should skip this note and also document the Rust API instead

PR #2597, opened 2024-08-26
feat: plugin setup docs consistency, fix desktop & mobile only setup

lucasfernog:
- makes the setup docs consistent across all plugins
- adjust the manual installation steps for desktop or mobile only plugins

### DIFF
--- a/src/content/docs/develop/Plugins/develop-mobile.mdx
+++ b/src/content/docs/develop/Plugins/develop-mobile.mdx
@@ -21,8 +21,7 @@ The default plugin template splits the plugin's implementation into two separate
 
 The desktop implementation uses Rust code to implement a functionality, while the mobile implementation sends a message to the native mobile code to execute a function and get a result back. If shared logic is needed across both implementations, it can be defined in `lib.rs`:
 
-```rust
-// lib.rs
+```rust title="src/lib.rs"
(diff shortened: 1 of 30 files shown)


vasfvitor (approved):
great, minor note is that in some places we're using 
```
```sh frame=none
```
image


and others
```
```sh
```
image

so that's may be a consistency point we need to tackle, do you have an opinion on this? if anything, I can do it as it is just manual labor.

I'm now looking at the live preview, but so far so good

vasfvitor (commented):
some plugins have the compatibility and minimum rust version listed. some not, about this there is this PR https://github.com/tauri-apps/tauri-docs/pull/1881 that it may well also fit the plugin minimum version (but still needs to be manually updated, at least in one place). 

Going further, join PluginLinks with the compatibility to create a PluginHeading

lucasfernog:
we should really start merging PRs, second time I've touched something already covered somewhere else.

lucasfernog:
ahh i see you only added the component and not the references, so i guess it's easier to manage this

lucasfernog:
let's go with ` frame=none` for now, though the other one is just a side effect of using CommandTabs, which uses the Code component directly instead of triple ticks.

PR #2599, opened 2024-08-27
i18n(zh-cn): Fix /plugin/barcode-scanner.mdx

Niavana97:
fix missing command

#### Description
-i18n(zh-cn): Add the missing cargo add command.


lucasfernog:
We just pushed some changes to the installation method for all plugins (the cargo add command are now in a code block so it can easily be copied) but I'll take this change for now :)

lucasfernog (approved):

PR #2600, opened 2024-08-27
docs: fix some fn run miss pub

simonwong:
#### Description

- Closes # 

`lib.rb` should be `pub fn run()`. In the setup of plugin, many documents miss `pub`


FabianLars (approved):
thank you

PR #2601, opened 2024-08-27
docs:fix the caution  in distribute/signing/linux 

Sir-Thom:
#### Description
The caution bubble was not showing correctly  in `distribute/signing/linux/ `
 because of the The signature is not verified! after `:::caution`


tweidinger:
Thanks for the catch!

PR #2610, opened 2024-08-29
fix: flatpak runtimes names

vjousse:
#### Description

Fix the flatpak runtimes names. They should be lower case or flatpak will not find the corresponding packages.


vasfvitor (approved):
thanks for this

PR #2618, opened 2024-08-30
register TAURI_WINDOWS_SIGNTOOL_PATH environment variable

vasfvitor:
initially I imported directly the content, but this way we lose TOC and a bit of customization (the purple aside)
so for now it just adds a new env var

https://github.com/tauri-apps/tauri/blob/5ec74456b9ae45eaa63e225856de1b7eb83abe6f/crates/tauri-cli/ENVIRONMENT_VARIABLES.md?plain=1#L23C4-L23C31


dreyfus92 (approved):
Nice catch @vasfvitor, LGTM ✅

PR #2624, opened 2024-09-01
[Docs]: Update window customization plugin guide

knownasnaffy:
#### Description

Updated the documentation for the plugin [window customization](https://v2.tauri.app/plugin/window-customization/). 
- Added a section for manual implementation of `data-tauri-drag-region` as suggested in #539.
- Updated code highlight for capabilities config
- Added a note about `data-tauri-drag-region` stating that the attribute works only on the parent and will need to be applied individually to the child elements.


tweidinger (approved):

tweidinger:
don't know to much about the drag behavior but looks like a good more in detail explanation to me

PR #2638, opened 2024-09-03
fix(v1): Correct Debian Bundle Example Path

tweidinger:
Updates the Debian example bundling code block to use the correct path as mentioned in the code comment.


FabianLars (approved):

FabianLars:
i always thought that's intentional but it really made no sense, mb lol

PR #2645, opened 2024-09-05
[i18n-ignore] Reorder key-value pairs in "tauri.conf.json" in "leptos.mdx"

just-another-notion:
Changes the order of the key-value pairs in "tauri.conf.json" in "leptos.mdx" to be consistent with the order used for Next.js, Nuxt, SvelteKit and Trunk.


vasfvitor:
Its ok, just have to make sure all locales are updated, I think only zh-cn has the same content

PR #2663, opened 2024-09-09
docs(plugins/global-shortcut): add ShortcutState usage

jason89521:
#### Description

Add `ShortcutState` in the usage for making users know that there is an api to know whether the shortcut is pressed or released.

https://github.com/tauri-apps/plugins-workspace/issues/1748#issue-2511535398


FabianLars (approved):
thx!

PR #2667, opened 2024-09-10
feat(plugins): document required Info.plist values

lucasfernog:


FabianLars:
Do we also need something for the fs plugin? https://github.com/ionic-team/capacitor-plugins/blob/aaa8658eb65b6c95ae231a3ed1547f3f78cd2cac/filesystem/README.md?plain=1#L47

(Does the plugin even work on iOS?)

~~Also, would you mind adding a stub geolocation page with NSLocationWhenInUseUsageDescription ? (I can open an issue instead too to add it when we actually add the guide)~~ Actually dw about this part, i'll try to add the guide this week.

tweidinger (approved):

tweidinger:
I think we can merge these docs first and then iterate on the other plugins as mentioned by Fabian. Looks good to me.

lucasfernog:
it does work on iOS (and some parts for android) btw

PR #2668, opened 2024-09-10
docs: Add Info.plist localization guide

FabianLars:
I still need to actually test the resources config, couldn't remember how it behaves after https://github.com/tauri-apps/tauri/pull/10293

I also moved the Sign section further down.


FabianLars:
ok resources config seems to work in my test

tweidinger (approved):

PR #2670, opened 2024-09-10
Update system-tray.mdx

timuric:
Note for those who are migrating from v1
https://github.com/tauri-apps/tauri/issues/10912


tweidinger:
Imho I think this should be part of the migration guide in https://v2.tauri.app/start/migrate/from-tauri-1/#migrate-to-tray-icon-module as the plugin documentation itself is for v2. Wdyt @FabianLars ?

FabianLars:
Yes, but even then the comment is a bit too extreme. It's not necessarily obsolte, it just works differently now.

timuric:
@FabianLars My intention was to prevent people from creating two tray icons; 

> It's not necessarily obsolteIt's not necessarily obsolte

Perhaps the note could point to the configuration alternative and why it would be useful.

PR #2673, opened 2024-09-12
Fix dead links in release notes

rmburg:
#### Description

- What the title says.


FabianLars:
Thanks for contributing! In my opinion the v1 docs should only show v1 patch notes for now (because v1 release notes aren't merged into the v2 branch anymore). For that, the old links were correct (except create-tauri-app) but requires the branch to be changed from `dev` to `1.x`.

rmburg:
Thanks, I'll update the PR accordingly.

FabianLars (approved):

PR #2678, opened 2024-09-12
feat: enhance documentation for desktop-only plugins

lucasfernog:
usage should actually be behind the #[cfg(desktop)] flag

### DIFF
--- a/src/content/docs/plugin/single-instance.mdx
+++ b/src/content/docs/plugin/single-instance.mdx
@@ -74,9 +74,9 @@ The closure has three arguments:
 
   <Steps>
 
-1. **`app` :** The [AppHandle](https://docs.rs/tauri/latest/tauri/struct.AppHandle.html) of the application.
-2. **`args` :** The list of arguments, that was passed by the user to initiate this new instance.
-3. **`cwd` :** The Current Working Directory denotes the directory from which the new application instance was launched.
+1. **`app`:** The [AppHandle](https://docs.rs/tauri/2.0.0-rc/tauri/struct.AppHandle.html) of the application.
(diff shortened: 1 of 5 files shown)


src/content/docs/plugin/single-instance.mdx
```diff
@@ -96,10 +96,15 @@ By default, when you initiate a new instance while the application is already ru
 use tauri::{AppHandle, Manager};
 
 pub fn run() {
-    tauri::Builder::default()
-        .plugin(tauri_plugin_single_instance::init(|app, args, cwd| {
+    let mut builder = ;tauri::Builder::default()
```
FabianLars:
```suggestion
    let mut builder = tauri::Builder::default();
```

lucasfernog: just checking if you're actually reviewing or just clicking 🤣

FabianLars:
ngl looking at this i kindaaa want to change the platform specific plugins to be compilable on all platforms, then you wouldn't have to have that many cfg flags (unless you want maximum optimization ig). But that can be done later too if we actually want to at some point 🤷

FabianLars (approved):

dreyfus92:
Thank you both! 😁

lucasfernog:
> ngl looking at this i kindaaa want to change the platform specific plugins to be compilable on all platforms, then you wouldn't have to have that many cfg flags (unless you want maximum optimization ig). But that can be done later too if we actually want to at some point 🤷 

i remember chip also wanted to do something similar for the entire codebase.. have less cfg and more no-ops would be beneficial

PR #2694, opened 2024-09-16
Add dts dtx keys to info plist

rtmotiondev:
#### Description

Adds DTS and DTX keys to Info.plist. Without them it's impossible to submit a test flight build for review. 
Referenced in: https://github.com/tauri-apps/tauri/issues/4415


rtmotiondev:
This should be automated but @FabianLars recommended handling it manually for the time being.

src/content/docs/distribute/app-store.mdx
```diff
@@ -157,6 +157,16 @@ Create a Info.plist file in the src-tauri folder:
 <dict>
 	<key>ITSAppUsesNonExemptEncryption</key>
 	<false/> # or `true` if your app uses encryption
+	<key>DTPlatformName</key>
```
lucasfernog, line 160: are these values just referencing the system that built the app bundle? or does `DTSDKName` reflect in the minimum supported version? i think it's just metadata right?

lucasfernog: also are you sure we do not need `DTPlatformBuild`? research online pointed we might need it

rtmotiondev: These are the values that worked for my case when submitting a TestFlight app for review. There might be other edge cases where different values are needed, but I haven't encountered them. These values are specific to macOS. The DTPlatformName indicates the platform for which the build is intended. It may be helpful to add comments explaining that this should only be included in macOS builds. For iOS, this may not be necessary since it uses an Xcode project.

rtmotiondev: DTSDKName specifies the SDK used, and as far as I can tell, it is just metadata.

PR #2727, opened 2024-09-23
docs: Add further `resources` syntax explanations

### LINKED ISSUE #2708: [bug] Nested resources not being bundled correctly
### Describe the bug

When creating installers, tested only on Windows, nsis and msi targets do not install all folders recursively specified under `resources` with the glob pattern. In the reproduction, the `library` folder is excluded from the installers.

### Reproduction

Folder layout:
```
root/
├─ resources/
│  ├─ data.db
│  ├─ library/
│  │  ├─ static.json
├─ src/
├─ src-tauri/
│  ├─ tauri.conf.json
```

Relevant config for `tauri.conf.json`
```json5
  "bundle": {
    "active": true,
    "targets": ["nsis", "msi"],
    "resources": ["../resources/*"],
    // ...
  },
```

Run `tauri build` to generate the installers.

### Expected behavior

I expect `../resources/*` to recursively install all files and folders during installation. However, only top-level files are installed, e.g. `_up_/resources/data.db`.

**Files expected to be installed:**
- `_up_/resources/data.db`
- `_up_/resources/library/static.json`

### Full `tauri info` output

```text
[✔] Environment
    - OS: Windows 10.0.19045 x86_64 (X64)
    ✔ WebView2: 128.0.2739.67
    ✔ MSVC: Visual Studio Community 2022
    ✔ rustc: 1.81.0 (eeb90cda1 2024-09-04)
    ✔ cargo: 1.81.0 (2dbb1af80 2024-08-20)
    ✔ rustup: 1.27.1 (54dd3d00f 2024-04-24)
    ✔ Rust toolchain: stable-x86_64-pc-windows-msvc (default)
    - node: 20.17.0
    - npm: 10.8.1
    - bun: 1.1.27

[-] Packages
    - tauri 🦀: 2.0.0-rc.11
    - tauri-build 🦀: 2.0.0-rc.10
    - wry 🦀: 0.43.1
    - tao 🦀: 0.30.0
    - @tauri-apps/api : 2.0.0-rc.4
    - @tauri-apps/cli : 2.0.0-rc.13

[-] Plugins
    - tauri-plugin-shell 🦀: 2.0.0-rc.3
    - @tauri-apps/plugin-shell : not installed!
    - tauri-plugin-updater 🦀: 2.0.0-rc.3
    - @tauri-apps/plugin-updater : 2.0.0-rc.2

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:1420/
    - framework: Vue.js (Quasar)
    - bundler: Vite
```

### Additional context

This used to work on older versions of v2, I'm not sure which version broke this change since my last app release was with version v2-beta.23.

### Workaround
```json5
  "bundle": {
    // ...
    "resources": ["../resources/*", "../resources/library/*"]
  },
```

FabianLars:
fixes #2708

### DIFF
--- a/src/content/docs/develop/resources.mdx
+++ b/src/content/docs/develop/resources.mdx
@@ -10,7 +10,7 @@ To bundle the files of your choice, you can add the `resources` property to the 
 
 See more about `tauri.conf.json` configuration [here][tauri.bundle].
 
-`resources` expects a list of strings targeting files or folders either with absolute or relative paths. It supports glob patterns (bash-style) in case you need to include multiple files from a directory.
+`resources` expects a list of strings targeting files or directories either with absolute or relative paths. It supports glob patterns (bash-style) in case you need to include multiple files from a directory.
 
 Here is a sample to illustrate the configuration. This is not a complete `tauri.conf.json` file:
 
--- a/src/content/docs/develop/resources.mdx
+++ b/src/content/docs/develop/resources.mdx
@@ -51,11 +51,11 @@ In Tauri's [permission system](../reference/acl/), absolute paths and paths cont
 In the following explanations "target resource directory" is either the value after the colon in the object notation, or a reconstruction of the original file paths in the array notation.
 
 - `"dir/file.txt"`: copies the `file.txt` file into the target resource directory.
+- `"dir/"`: copies all files **and directories** _recursively_ into the target resource directory. Use this if you also want to preserve the file system structure of your files and directories.
 - `"dir/*"`: copies all files in the `dir` directory _non-recursively_ (sub-directories will be ignored) into the target resource directory.
 - `"dir/**`: throws an error because `**` only matches directories and therefore no files can be found.
 - `"dir/**/*"`: copies all files in the `dir` directory _recursively_ (all files in `dir/` and all files in all sub-directories) into the target resource directory.
 - `"dir/**/**`: throws an error because `**` only matches directories and therefore no files can be found.
-- `"dir/"`: copies all files **and directories** _recursively_ into the target resource directory. Use this if you also want to preserve the file system structure of your files and directories.
 
 ## Accessing files in Rust
 
--- a/src/content/docs/develop/resources.mdx
+++ b/src/content/docs/develop/resources.mdx
@@ -10,7 +10,7 @@ To bundle the files of your choice, you can add the `resources` property to the 
 
 See more about `tauri.conf.json` configuration [here][tauri.bundle].
 
-`resources` expects a list of strings targeting files either with absolute or relative paths. It supports glob patterns in case you need to include multiple files from a directory.
+`resources` expects a list of strings targeting files or folders either with absolute or relative paths. It supports glob patterns (bash-style) in case you need to include multiple files from a directory.
 
 Here is a sample to illustrate the configuration. This is not a complete `tauri.conf.json` file:
 
@@ -21,6 +21,7 @@ Here is a sample to illustrate the configuration. This is not a complete `tauri.
       "/absolute/path/to/textfile.txt",
       "relative/path/to/jsonfile.json",
       "resources/*"
+      "resources/**/*"
     ]
   }
 }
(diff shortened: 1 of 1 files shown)


src/content/docs/develop/resources.mdx
```diff
@@ -34,12 +35,22 @@ Alternatively the `resources` config also accepts a map object if you want to ch
...
     }
   }
 }
 ```
 
+The syntax for resource source paths works like this:
+
+- `"dir/file.txt"`: Will copy the `file.txt` file into the target resource directory.
+- `"dir/*"` / `dir/**`: Will copy all files in the `dir` directory _non-recursively_ (sub-directories will be ignored) into the target resource directory.
+- `"dir/**/*"`: Will copy all files in the `dir` directory _recursively_ (all files in all sub-directories) into the target resource directory. Note, this does **not** include the files from `dir/*`, but only those in sub-directories.
```
amrbashir: Not sure what you mean by does not include files from `dir/*`

src/content/docs/develop/resources.mdx
```diff
@@ -34,12 +35,22 @@ Alternatively the `resources` config also accepts a map object if you want to ch
...
+- `"dir/"`: Will copy all files **and directories** _recursively_ into the target resource directory. Use this if you also want to preserve the file system structure of your files and directories.
+
+The `target resource directory` is either the right value of the object notation, or a reconstruction of the source path in the array notation.
```
amrbashir: This was a bit confusing at first and I had to read it multiple times to know what it meant, maybe using `"` instead of a backtick is better here. A better solution would be to add intralinks where "target resource directory" is mentioned.

I also think this section should have its own header

src/content/docs/develop/resources.mdx
FabianLars: This is what was confusing me so much in #2708 - this comment was based on 2 things - 1) you saying we're following bash's behavior and 2) the build script fucking with my head.

Since globstar is not enabled by default in bash `*` and `**` behave the same which i didn't notice because i didn't check the config, and didn't look at the output of `**/*` closely enough to notice it missing files 3 levels down.

I'll go through all of it again.

amrbashir: It is not that we are following bash's behavior and more like `glob` crate is similar to bash globbing but it is actually more similar to libc's `glob` function

FabianLars:
@amrbashir Please check again. I tried to phrase everything in a way that works for both `[]` and `{}` configs (having in mind that `dir/**/*` works differently for both due to the path reconstruction)

src/content/docs/develop/resources.mdx
```diff
@@ -34,18 +34,29 @@ Alternatively the `resources` config also accepts a map object if you want to ch
...
+## Source path syntax
+
+In the following explanations "target resource directory" is either the value after the colon in the object notation, or a reconstruction of the original file paths in the array notation.
+
+- `"dir/file.txt"`: Will copy the `file.txt` file into the target resource directory.
+- `"dir/*"`: Will copy all files in the `dir` directory _non-recursively_ (sub-directories will be ignored) into the target resource directory.
+- `"dir/**`: Will throw an error because `**` only matches directories and therefore no files can be found.
+- `"dir/**/*"`: Will copy all files in the `dir` directory _recursively_ (all files in `dir/` and all files in all sub-directories) into the target resource directory.
+- `"dir/**/**`: Will throw an error because `**` only matches directories and therefore no files can be found.
+- `"dir/"`: Will copy all files **and directories** _recursively_ into the target resource directory. Use this if you also want to preserve the file system structure of your files and directories.
```
amrbashir:
```suggestion
- `"dir/file.txt"`: copies the `file.txt` file into the target resource directory.
- `"dir/*"`: copies all files in the `dir` directory _non-recursively_ (sub-directories will be ignored) into the target resource directory.
- `"dir/**`: throws an error because `**` only matches directories and therefore no files can be found.
- `"dir/**/*"`: copies all files in the `dir` directory _recursively_ (all files in `dir/` and all files in all sub-directories) into the target resource directory.
- `"dir/**/**`: throws an error because `**` only matches directories and therefore no files can be found.
- `"dir/"`: copies all files **and directories** _recursively_ into the target resource directory. Use this if you also want to preserve the file system structure of your files and directories.
```

amrbashir: I think the `dir/` case should be explained right after the first one

FabianLars: ngl i had this first but it looked ugly in vscode x) will see how it looks in the website preview (can't build locally)

FabianLars: yeah, path syntax wise that'd make more sense. Just felt like this order because of its effect being the "highest" basically. I'll re-order it.

amrbashir (approved):

PR #2754, opened 2024-09-29
fix: commas in JSON code block

RealHinome:
#### Description
This PR addresses a comma-related issue in the V2 documentation that causes JSON formatting errors. Specifically: - a missing comma has been added; - an extraneous comma at the end of the JSON has been removed.
These changes ensure that the documentation's JSON syntax is valid and can be correctly parsed.

### DIFF
--- a/src/content/docs/plugin/stronghold.mdx
+++ b/src/content/docs/plugin/stronghold.mdx
@@ -213,9 +213,9 @@ See the [Capabilities Overview](/security/capabilities/) for more information an
 		"stronghold:allow-create-client",
 		"stronghold:allow-load-client",
 		"stronghold:allow-save",
-		"stronghold:allow-save-store-record"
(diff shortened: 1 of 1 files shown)


RealHinome:
The error is due to a formatting error; the elements are not in the right order. I'm changing that.

vasfvitor (changes requested):
Thanks for this!

run `pnpm format` and it will fix whatever is breaking the formatting. I believe it wasn't due to the order, and they make more sense as they were before (e.g. `save-`, `get-`, `remove-` `store-record` in order), so could you please restore the original order and leave just the comma fix.

RealHinome:
I reverted my previous changes, sorry for inconvenience. As you said, it will be more idiomatic and checks will pass.

tweidinger:
Also outdated by #2759

PR #2765, opened 2024-10-01
feat: update dialog Rust API, add note on path format for mobile

lucasfernog:
I'm not totally happy with how the dialog docs layout is completely different from the others, but I'm not willing to change it right now - especially not without discussing with the other docs maintainers

this PR clarifies the return path format for Android and iOS, and also updates the Rust examples following the change from https://github.com/tauri-apps/plugins-workspace/pull/1842


vasfvitor (approved):

goenning:
image

I was trying to follow the example on the docs, but it doesn't seem to works. I'm being forced to do a `"Yes".to_string()`. 
Are the docs wrong or was this supposed to accept a `&str`?

PR #2771, opened 2024-10-01
feat: add Node.js as a sidecar guide

lucasfernog:


tweidinger:
I think a good start but should be improved later on with more explicit copy paste ready files (or instructions in which files and where to add the code blocks).

Changed a bit of layout and prettier complains even though the write fix breaks it even more on my end 🤷 

LGTM

tweidinger (approved):

lucasfernog:
> Changed a bit of layout and prettier complains even though the write fix breaks it even more on my end 🤷

just add some newlines and it'll stop complaining 😂

PR #2773, opened 2024-10-01
refactor: change NPM dependencies to install v1 instead of latest

lucasfernog:
we will launch v2 soon which will take over the `latest` tag, so the v1 instructions should properly use the v1 versions instead of latest

### DIFF
--- a/src/theme/Command.js
+++ b/src/theme/Command.js
@@ -71,7 +71,7 @@ export const InstallTauriCli = () => {
     <Tabs groupId="package-manager">
       <TabItem value="npm" default>
         <CodeBlock className="language-shell" language="shell">
-          npm install --save-dev @tauri-apps/cli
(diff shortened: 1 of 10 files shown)


vasfvitor (approved):
ah ok, v1 docs

src/theme/Command.js
```diff
@@ -116,22 +116,22 @@ export const InstallTauriApi = () => {
     <Tabs groupId="package-manager">
       <TabItem value="npm" default>
         <CodeBlock className="language-shell" language="shell">
-          npm install @tauri-apps/api
+          npm install @tauri-apps/api@"&gt;1.0.0"
```
tweidinger, line 119: intended html encoding?

lucasfernog: yep

tweidinger (approved):

PR #2776, opened 2024-10-02
feat: add NFC usage documentation

lucasfernog:


lucasfernog:
the Rust API needs some love.. that's for sure

maybe we even delay its stable release a bit

vasfvitor (approved):

PR #2782, opened 2024-10-02
Improved macOS Signing section

0PandaDEV:
#### Description
- Added more details regarding code signing with free apple developer accounts.
- Added GitHub Actions example for code signing
- Dependency updates in package.json to latest versions


vasfvitor:
> Dependency updates in package.json to latest versions

hey, if it's to be done (not the case) it would be best in a dedicated PR. But as of now, renovate bot handles the packages updates to the latest versions. Keep in mind the even if `package.json` doesn't reflect it, it is updated in the lockfile

0PandaDEV:
i reverted the package changes

vasfvitor (approved):

PR #2802, opened 2024-10-05
definitely fix (hopefully) trailing slash and broken urls

### LINKED ISSUE #2812: [docs] clicking link in docs brings to 404
image

https://tauri.app/concept/architecture/

Edit: actually its all 3 of the links:

`custom menus` - 404
`tray-type-interfaces` - 404
`updated` - 404

vasfvitor:
disabled italian locale because it had only 1 page

note that trailiing slash: always forces the blog plugin to use trailing slash


astro.config.mjs
```diff
@@ -64,8 +64,7 @@ const site = 'https://v2.tauri.app';
 // https://astro.build/config
 export default defineConfig({
   site,
-  // TODO: Fix trailing slashes throughout the docs
-  // trailingSlash: 'always',
```
dreyfus92, line 68: should we remove this?

dreyfus92 (commented):
So far it looks good to me 🫡

astro.config.mjs
vasfvitor: this (setting `trailingSlash: 'always'`) is what fixed the missing blog posts trailing slashes because it is handled by the starlight-blog, plus enabling it, makes the starlight-links-validator checks for missing trailing slashes.

Here is what I could guess after looking to the [commits history](https://github.com/tauri-apps/tauri-docs/commits/v2/astro.config.mjs): @simonhyll was doing docs work and enabled it, then "OH NO!", hundreds broken links that was too much to handle at the moment and wasn't priority....
image

it's just a guess, but this makes sense doesn't it?

dreyfus92 (approved):
let's ship and see what breaks 😉

vasfvitor:
> let's ship and see what breaks 😉

noooooooo 😭😭 

it just broke due to merges without trailing slash

vasfvitor:
ok, should be good to go once built

PR #2810, opened 2024-10-07
[i18n-ignore] Updated `src-tauri/capabilities/main.json` to `src-tauri/capabilities/default.json`

patelka2211:
#### Description

Updated `src-tauri/capabilities/main.json` to `src-tauri/capabilities/default.json` for the Security/Plugin permissions docs.

Does Tauri use `default.json` and `main.json` for different purposes? `default.json` is created by default, so I thought that maybe in stable releases, Tauri uses `default.json`.

Should I make this change to every instance?


FabianLars:
> Does Tauri use default.json and main.json for different purposes? default.json is created by default

We may have changed the file name or just messed up the docs, but functionality wise the file name doesn't matter, tauri will read all json files in that directory by default.

Edit: If the templates use default.json then ig it makes sense to have the docs match that.

patelka2211:
Thank you for the clarification, @FabianLars. Please let me know if you’d like me to update all instances of `main.json`. If not, I hope the changes I made [here](https://github.com/tauri-apps/tauri-docs/pull/2810/files#diff-ed9e30c72c213b61576fcb784d234b919396443a58dde10f05c2fd4704b5ddc5) are acceptable. Thanks!

FabianLars:
Without knowing where we use main.json right now (i'm on my phone), I think that it makes sense to update all of them, yes.

patelka2211:
Tauri template uses `default.json`. https://github.com/tauri-apps/tauri/blob/dev/crates/tauri-cli/templates/app/src-tauri/capabilities/default.json

vasfvitor (approved):
thanks. I wonder how didn't the docs broke with those `"` missing that you fixed

PR #2813, opened 2024-10-07
corrected description about createNew of open in file-system.mdx

HeveraletLaidCenx:
#### Description

    The original description about `createNew` of `open` is exactly the opposite of the actual effect. Correct it after test.

- Closes #


src/content/docs/plugin/file-system.mdx
```diff
@@ -330,7 +330,11 @@ Always call `file.close()` when you are done manipulating the file.
 
 - createNew
 
-  `createNew` works similarly to `create`, but if the file does not exist, the operation fails.
+  `createNew` works similarly to `create`, but as its name suggests, it is only takes effect when **New**: 
+
+  That is, only allowed to be executed when the file to be opened does not exist. 
+
+  Otherwise, the operation failed and raise an error like: `failed to open file at path: *the file path* with error: The file exists. (os error 80)`.
```
vasfvitor, line 337: I saw your (?) comment somewhere about that it's quite the opposite way it was written before. I think we can simplify by just matching that "it fails if the file already exists" or something. No need to mention the error or explain "in other words".

vasfvitor (changes requested):
thanks for this, I left a suggestion

PR #2823, opened 2024-10-08
(i18n-ignore) docs: restore the window if it is minimized

RSS1102:
#### Description

image

In actual situations, when the app main is minimized to the tray, left-click the tray icon and you need to cancel the minimization before it can be displayed and the window focus can be set.

---
`D:\GithubDesktop\tauri-docs\packages\tauri\examples\api\src-tauri\src\tray.rs`
`D:\GithubDesktop\tauri-docs\packages\plugins-workspace\examples\api\src-tauri\src\tray.rs`

I don't know if this also needs to be changed for the example?
image


Legend-Master (approved):
Thanks, sorry for the delay

PR #2861, opened 2024-10-14
fix: wrong JavaScript example in upload plugin

LightQuanta:
#### Description
upload plugin uses [ProgressHandler](https://github.com/tauri-apps/plugins-workspace/blob/7e1c17a635f10d69e10ea9f35081a68c71276203/plugins/upload/guest-js/index.ts#L7) as a parameter, which needs destructuring assignment in order to get its two parameters, `progress` and `total`.


vasfvitor:
thanks, you need to check formatting `pnpm run format`.

vasfvitor (approved):

PR #2885, opened 2024-10-19
Update calling-frontend.mdx to fix missing match arm

luis-prates:
#### Description

- What does this PR change? 
It fixes a missing match arm in the Webview Event emit_filter example


vasfvitor:
I'm not sure it is strictly necessary but is more readable this way, thanks

PR #2893, opened 2024-10-22
Fix hover color for links outside the header

olivierlemasle:
#### Description
Fix hover color for links outside the header

Using the light theme, all links with `:hover` are grey, which renders with a very poor contrast. This gray color should be only for links in the header, as it was most certainly intended.

Before (home page):
image

After (home page): 
image

Before (navbar):
image

After (navbar):
image


vasfvitor:
great ty

PR #2897, opened 2024-10-23
Fix import statement for invoke in Next.js setup tutorial

lukasb:
update import location for invoke

#### Description

- Closes #


FabianLars:
You're editing the v1 branch which is for tauri v1. In this context the current import is correct, it was changed to core in v2.

lukasb:
> You're editing the v1 branch which is for tauri v1. In this context the current import is correct, it was changed to core in v2.

In that case someone else should probably edit this tutorial so it installs Tauri v1, right now it installs v2.

PR #2911, opened 2024-10-27
TypeScript error fix for NextJS 15

dcodesdev:
#### Description

Documentation updated that fixes a TypeScript error.

`assetPrefix` requires type `string | undefined` giving `null` is a TypeScript error.


vasfvitor (approved):
hey thanks, would you also update the same page on other locales if they have the same issue?

PR #2914, opened 2024-10-28
Update vite.mdx

BSD-Yassin:
Corrected the withGlobalTauri mention in the configuration, which switches places depending on V1 or V2

#### Description

- Simply added the JSON for the app object, so that anyone using V2 will not be confused about the option `    "withGlobalTauri":true` not working in build.

Here's the reference for the change I made : [Appconfig](https://v2.tauri.app/reference/config/#appconfig)


FabianLars:
The v1 branch is for Tauri v1. For v2 instructions there's the v2 branch. Thanks tho

PR #2917, opened 2024-10-28
docs: Update 'npm create-' to 'npm init' for install

jmcscript:
Change `npm create tauri-app` to `npx create-tauri-app` as it does not require an npm install beyond the desired result.


FabianLars:
Sorry for the long delay here. I have to admit that the proposed doesn't make sense to me (as someone who doesn't use npm).
You say that this is done to not require an npm install but now you need npx installed (both of which usually are installed together)? Sorry if i'm dense but i don't get it 😅

jmcscript:
> Sorry for the long delay here. I have to admit that the proposed doesn't make sense to me (as someone who doesn't use npm).
> You say that this is done to not require an npm install but now you need npx installed (both of which usually are installed together)? Sorry if i'm dense but i don't get it 😅 

My apologies for the lack of specificity. NPM must be installed regardless. But the create tauri app package only has to be installed if you use the NPM command. If you use NPX it just executes from that package. That's my understanding.

FabianLars:
Ah that's what you mean, got it. The create/init command is basically a nicer alias for npx though: https://docs.npmjs.com/cli/v6/commands/npm-init?v=true#description

jmcscript:
> Ah that's what you mean, got it. The create/init command is basically a nicer alias for npx though: https://docs.npmjs.com/cli/v6/commands/npm-init?v=true#description

I agree, that is a pleasant alternative. Shall I update the Pull Request with that instead?

jmcscript:
> Ah that's what you mean, got it. The create/init command is basically a nicer alias for npx though: https://docs.npmjs.com/cli/v6/commands/npm-init?v=true#description

If you guys are not interested in this one, just let me know, and I'll close it out! Have a great day.

FabianLars:
Hmm, my comment wasn't send (probably cause i didn't click on send again).

npm create and npm init are the same commands (create is an actual alias for init). No idea why npm doesn't mention this anywhere in the docs. But yeah, i think we should keep `npm create` to match the other package managers and to match what other projects seem to prefer.

Thanks though :)

jmcscript:
> Hmm, my comment wasn't send (probably cause i didn't click on send again).
> 
> npm create and npm init are the same commands (create is an actual alias for init). No idea why npm doesn't mention this anywhere in the docs. But yeah, i think we should keep `npm create` to match the other package managers and to match what other projects seem to prefer.
> 
> Thanks though :)

Thank you, kindly!

PR #2932, opened 2024-11-02
docs: update app-store.mdx

dionysuzx:
#### Description

I noticed that after running `cargo tauri ios init` and then following the docs to do `cargo tauri icon <path> --ios-color #fff` is that:

1. I get an error `error: a value is required for '--ios-color <IOS_COLOR>' but none was supplied` (fish shell)
2. It doesn't overwrite the icons that are generated under `apple/` from `cargo tauri ios init`; meaning the user needs to manually copy it

This PR improves the docs by fixing the command by adding quotes around `#fff` and a step to manually copy.

### DIFF
--- a/src/content/docs/distribute/app-store.mdx
+++ b/src/content/docs/distribute/app-store.mdx
@@ -20,14 +20,14 @@ Additionally, you must setup code signing for [macOS][macOS code signing] and [i
 
 ## Changing App Icon
 
-After running `tauri ios init` to setup the Xcode project, you can use the `tauri icon` command to update the app icons.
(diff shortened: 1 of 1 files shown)


FabianLars:
This is actually not supposed to be required: https://github.com/tauri-apps/tauri/blob/12ffc19ce08eeafdedc65207f312432aa2cf9c84/crates/tauri-cli/src/icon.rs#L448

dionysuzx:
> This is actually not supposed to be required: https://github.com/tauri-apps/tauri/blob/12ffc19ce08eeafdedc65207f312432aa2cf9c84/crates/tauri-cli/src/icon.rs#L448

thank you @FabianLars. I will convert this PR into a bug report.

PR #2948, opened 2024-11-06
Update: Flatpak Distribution Docs

RoseBlume:
#### Description

-  
- Closes # 

This PR makes some modifications to the original one. This introduces a new method which is preferred by the flathub staff. It will make it easier for people to submit their apps to flathub without being asked to make as many changes.


FabianLars:
> This introduces a new method which is preferred by the flathub staff. It will make it easier for people to submit their apps to flathub without being asked to make as many changes.

Are you talking from experience? == did you talk with them, or did you see discussions around this?

I mean, i know that this is preferred and wanted to see this in the docs anyway, but i'd also be interested in seeing what they said.

FabianLars:
oh and thanks for the PR! i'm quite busy this week (was out of office the last month so tons of things to catch up on) so i may not check in again until next week :/

RoseBlume:
> > This introduces a new method which is preferred by the flathub staff. It will make it easier for people to submit their apps to flathub without being asked to make as many changes.
> 
> Are you talking from experience? == did you talk with them, or did you see discussions around this?
> 
> I mean, i know that this is preferred and wanted to see this in the docs anyway, but i'd also be interested in seeing what they said.

This is stuff that they wanted me to fix when I published my app. I am also updating the gnome version and I would like to completely get rid of the old stuff.

FabianLars:
> This is stuff that they wanted me to fix when I published my app.

yeah, just found the pr.

> I am also updating the gnome version and I would like to completely get rid of the old stuff.

don't remove the old stuff, closed source apps will still need it.

RoseBlume:
That makes sense. It might be better to change the tabs to Open Source and Close Source Options

src/content/docs/distribute/flatpak.mdx
```diff
@@ -6,6 +6,182 @@ sidebar:
...
+      - node-sources.json
+    build-commands:
+      - echo -e 'yarn-offline-mirror "/run/build/your-module/flatpak-node/yarn-mirror"\nyarn-offline-mirror-pruning true' > /run/build/your-module/.yarnrc
+      - mkdir -p src-tauri/.cargo && echo -e '[source.crates-io]\nreplace-with = "vendored-sources"\n\n[source.vendored-sources]\ndirectory = "/run/build/your-module/cargo/vendor"' > src-tauri/.cargo/config.toml
+      - yarn install  --offline --immutable --immutable-cache --inline-builds
+      - yarn run tauri build -- -b deb
+      - ar -x src-tauri/target/release/bundle/deb/*.deb
+      - tar -xf src-tauri/target/release/bundle/deb/your-app/data.tar.gz
+      - install -Dm755 src-tauri/target/release/bundle/deb/your-app/data/usr/bin/your-command /app/bin/your-command
+      - install -Dm644 src-tauri/target/release/bundle/deb/your-app/data/usr/share/applications/Rosary-Music.desktop /app/share/applications/org.your.id.desktop
```
RayJW: `src-tauri/target/release/bundle/deb/your-app/data/usr/share/applications/Rosary-Music.desktop` should be `src-tauri/target/release/bundle/deb/your-app/data/usr/share/applications/your-app.desktop`

RoseBlume: Thank you. I appreciate the feedback. This will be fixed soon.

thegreatGreenstar:
As reference, it may be worth looking into: https://github.com/flatpak/flatpak-github-actions 
(Found per https://github.com/flathub/flathub/issues/3281)

src/content/docs/distribute/flatpak.mdx
```diff
@@ -2,11 +2,186 @@
...
+          "/usr/share/metainfo/org.your.id.metainfo.xml": "relative/path/from/your/tauri.conf.json/to/your/org.your.id.metainfo.xml"
+        }
+      }
+    }
+```
+4. Create your manifest
+```yaml
+id: org.your.id
+
+runtime: org.gnome.Platform
```
thegreatGreenstar, line 118: We shouldn't be assuming GNOME here.
Instead, we should default to Freedesktop.   
https://docs.flatpak.org/en/latest/available-runtimes.html#freedesktop
Both GNOME and KDE are big enough we cannot assume either as a default. With stuff like SteamOs and personal use: KDE has a massive userbase,. Meanwhile on the more corporate side via stuff like Ubuntu: GNOME tends to win out. However we shouldn't assume either of those two without significant reason. If an application desires to target one; that should be a decision they have to make, not are defaulted into.
 
Ditto for the other instances where we are assuming Gnome, but to avoid flood this shall be my sole mention thereof.

src/content/docs/distribute/flatpak.mdx
```diff
@@ -2,11 +2,186 @@
...
+    </supports>
+
+    <description>
+        <p>
+            Your Description
+        </p>
+    </description>
+    <screenshots>
+        <screenshot type="default">
+            <image>https://site.com/your-image.png</image>
```
thegreatGreenstar, line 86: Haven't checked Flatpak side of this but if possible the more modern WebP should be recommended. (RE:https://developers.google.com/speed/webp/faq) 
Impact is small but, modernization is made with a thousand small steps.

thegreatGreenstar: Not possible, https://specifications.freedesktop.org/icon-theme/latest/
Found per https://docs.flatpak.org/en/latest/conventions.html

src/content/docs/distribute/flatpak.mdx
```diff
@@ -2,11 +2,186 @@
...
+id: org.your.id
+
+runtime: org.gnome.Platform
+runtime-version: '47'
+sdk: org.gnome.Sdk
+
+command: tauri-app
+finish-args:
+  - --socket=wayland # Permission needed to show the window
+  - --socket=fallback-x11 # Permission needed to show the window
```
thegreatGreenstar:
```suggestion
  - --socket=fallback-x11 # Permission needed to show the window on legacy windowing systems
```

src/content/docs/distribute/flatpak.mdx
thegreatGreenstar: Created an issue to move towards this becoming possible: https://gitlab.freedesktop.org/xdg/xdg-specs/-/issues/244

RoseBlume: That is not an icon. It acts as a preview on the chosen software store. An example is shown below


image

thegreatGreenstar: My bad, commented on the wrong spot. I believe that for icons this is otherwise correct, though needs to be addressed upstream instead of here still.
As for preview... before we even consider modern stuff there the icon itself must support said modern formats.

src/content/docs/distribute/flatpak.mdx
RoseBlume: Using the Gnome Platform does not prevent it from running on KDE devices. The Gnome Platform simply includes the required libraries such as webkitgtk2-4.1 so that it does not need to be bundled with the app. This reduces the size of the bundle

RoseBlume:
> As reference, it may be worth looking into: https://github.com/flatpak/flatpak-github-actions (Found per [flathub/flathub#3281](https://github.com/flathub/flathub/issues/3281))

This does properly build a bundle, but the guide is intended to distribute via flathub as it is typically the default source, and will allow it to appear on SteamOS

src/content/docs/distribute/flatpak.mdx
thegreatGreenstar: Yes, however this leads to us preferring GTK over QT. I do not know enough about Tauri to know if this is something needed in which case that's fine, however if not...
KDE prefers QT, GNOME prefers GTK. Ideally once more neutrality should be maintained if possible.
Edit: I see. GTK is preferred here, thus so is GNOME. https://v2.tauri.app/start/prerequisites/

RoseBlume: GTK is required by tao for windowing on linux. As such the decision to use the Gnome runtime for simplicity was made

FabianLars: It's also required for the webview we currently use. It's not about preference here but requirements.

ghostdevv:
Is there a way we can help move this forward? I run into Tauri apps in flatpaks every now and then, so I know it's possible - but we're of course missing docs currently and subsequently an easy way to help get apps from projects I'm not involved with to publish a flatpak

dreyfus92 (approved):
thanks @RoseBlume for sticking with this, and thanks to everyone who reviewed. We're going to ship this as-is. it's a solid baseline, and we can iterate in follow-ups as things need updating or improving.

PR #2963, opened 2024-11-11
Move non-plugins guides into learn section

FabianLars:
https://discord.com/channels/616186924390023171/662624589037436928/1305535070500814890


FabianLars:
why did i even check for links beforehand if i miss them anyway

Issue #2966, opened 2024-11-11
[request] Include list of all core permissions in documentation


Simon-Laux:
### Question you want answered

where can I find list of all core permissions?

### Where did you look for an answer?

I started locking down my app with permissions, but was unable to find a list of all core permissions on the documentation page.

Basically I started browsing through the sites in roughly this order:
- Guides/Security/Capabilities
- Guides/Develop/ConfigurationFiles
- Guides/Security/Command Scopes
- Guides/Security/Permissions
- Reference/Security/*
- Reference/Configuration

At first I had a feeling that the documentation about permissions would be geared more towards plugin and tauri(core) developers than users, but then again just a feeling, I must admit that I just skimmed over the content and didn't study it.

### Additional context

Then I asked on discord and got this link from @FabianLars: https://github.com/tauri-apps/tauri/tree/dev/crates/tauri/permissions

He asked me to create this issue here.

secdude suggested running:
```
tauri permission list
```
Even though it is not that readable in console it is still a good to know option IMO, so documentation site should also have it.

I think a mention in https://tauri.app/security/capabilities/ makes sense and additionally a page with all permissions on in the References tab.

PS: At first it was also not very clear to me how the capability / permission system works, probably I haven't read enough, so I didn't understand where to put the permissions and also didn't understand that core permissions need to be prefixed with `core:`. But that's not enough info for a concrete issue yet, also I don't have any suggestions yet.

### Are you willing to work on this yourself?

- [ ] I want to work on this myself

PR #2970, opened 2024-11-12
docs: add list of all core permissions [issue #2966]

514sid:
Related to issue https://github.com/tauri-apps/tauri-docs/issues/2966


FabianLars:
Thanks! I think for the future automating this may make sense (like we do for the plugins) 🤔 But for now this is totally fine.

@tweidinger idk if you care but are you okay with this being on the `security/capabilities/` page or would you prefer some other location?

tweidinger:
I think this kind of reference should be part of a [`reference/acl/core-capabilities.mdx`](https://github.com/tauri-apps/tauri-docs/tree/v2/src/content/docs/reference/acl) file.
Agreed on it is okay to have it manually for now but should be automatically generated from the already generated markdown of the crate (https://github.com/tauri-apps/tauri/blob/dev/crates/tauri/permissions/app/autogenerated/reference.md for example).

This shouldn't live in the guide section where the concept of a capability is explained.

514sid:
> I think this kind of reference should be part of a [`reference/acl/core-capabilities.mdx`](https://github.com/tauri-apps/tauri-docs/tree/v2/src/content/docs/reference/acl) file.

If the list of capabilities is added to this file, I think it would be nice to include a link to it at `/security/capabilities/`

src/content/docs/reference/.gitignore
```diff
@@ -1,5 +1,4 @@
 config.md
-acl/*.md
 cli.mdx
```
FabianLars, line 2:
```suggestion
cli.mdx
acl/*.md
!acl/core-permissions.mdx
```

idk if the "un-ignore" line is really needed with the different file extensions but the auto-generated .md files should still be ignored.

514sid: I think this change wasn't needed, you're right

src/content/docs/security/capabilities.mdx
```diff
@@ -227,3 +227,7 @@ Everything can be inlined into the `tauri.conf.json` but even a
 little more advanced configuration would bloat this file and
 the goal of this approach is that the permissions are abstracted
 away whenever possible and simple to understand.
+
+## Core Permissions
+
+List of all core permissions can be found on the [Core Permissions](/reference/acl/core-permissions/) page.
```
FabianLars:
```suggestion
A list of all core permissions can be found on the [Core Permissions](/reference/acl/core-permissions/) page.
```

FabianLars (approved):
thanks :)

PR #1854, opened 2024-02-13
feat: Migrates `i18ntracker` to `Lunaria`

dreyfus92:
#### What kind of changes does this PR include?

- Changes to the docs site code

#### Description

- What does this PR change? Give us a brief description. 
Migrates `i18ntracker` package to use `lunaria` as default. This will offload maintenance of the `i18ntracker` and enhance a bit more the the current state of the tracker's UI.
> [!NOTE]
> Encountered an issue where Lunaria couldn't track the git history of certain files used in generating Tauri documentation. This is due to these files being fetched via a git submodule, which isn't reflected in the project's main git history.
> 
> Implemented a temporary fix in the Lunaria config to bypass this issue:
> 
> ``` json
> "ignore": [
>   "src/content/docs/references/{v1,v2}/(js/**/*|config).{md,mdx}"
> ]
> ```
> This change prevents Lunaria from tracking these files, thus avoiding the error. We'll explore a more permanent solution that maintains the necessary git history for seamless documentation generation in the future.
- Did you change something visual? A before/after screenshot can be helpful.

| Before | After |
| ------ | ----- |
| Before | After |

PR #2976, opened 2024-11-13
migrate i18n-tracker to Lunaria, closed #2853

### LINKED ISSUE #2853: migrate i18n-tracker to Lunaria
context: iirc the i18n-tracker we use was based off the one was used at starlight official docs. It evolved to [Lunaria](https://lunaria.dev/getting-started).

A PR was opened a while ago, but was closed https://github.com/tauri-apps/tauri-docs/pull/1854 . Much, if not everything can be reused with minor adjustments

Hopefully it will work without the temporary fix mentioned there.

GuoJikun:
#### Description
closes #2853


GuoJikun:
@vasfvitor How to get `GITHUB_TOKEN` for environment in `packages/lunaria/renderer.config.ts`

vasfvitor:
related: https://github.com/tauri-apps/tauri-docs/pull/1854/files

> @vasfvitor How to get `GITHUB_TOKEN` for environment in `packages/lunaria/renderer.config.ts`

Do we really need that to setup Lunaria? I think we can get rid of the external package and use Lunaria directly into the docs main package as an astro integration and just make a few config to mantain the same route instead of `/lunaria`, or is there any reason you did that way?

...and I can't create and manage tokens. If you still need to to know more about token, please tag FabianLars, not tagging him now because I'm not sure we will really need. Either way you can tag me again so we can bring this live. Thanks

GuoJikun:
> I think we can get rid of the external package and use Lunaria directly into the docs main package as an astro integration and just make a few config to mantain the same route instead of `/lunaria`

The Lunaria-related configuration has been moved to the main package

image
this module need Github Token

dreyfus92 (commented):
I created a pr months ago to add Lunaria to this repo, dunno if it helps as [ref](https://github.com/tauri-apps/tauri-docs/pull/1854/files), btw you might also need to test this removing the `i18ntracker` package.

GuoJikun:
> I created a pr months ago to add Lunaria to this repo, dunno if it helps as [ref](https://github.com/tauri-apps/tauri-docs/pull/1854/files), btw you might also need to test this removing the `i18ntracker` package.

The "NeedReview" module has no content yet because I don't know how to get the 'GITHBB_TOKEN'.
Your PR has been referenced.
`i18ntracker` is removed.

lunaria/components.ts
```diff
@@ -0,0 +1,75 @@
...
+  }[];
+}
+interface Options {
+  githubRepo: string;
+  githubToken: string;
+}
+async function getPullRequests(options: Options) {
+  const pullRequests: PullRequest[] = await githubGet({
+    url: `https://api.github.com/repos/${options.githubRepo}/pulls?state=open&per_page=100`,
+    githubToken: options.githubToken,
```
vasfvitor:
```suggestion

```
would you test removing the token altogether?  I'm pretty sure it should work without it. If not I think we need to setup a pat with read-only on pull requests

https://api.github.com/repos/tauri-apps/tauri-docs/pulls?state=open&per_page=100

https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests

vasfvitor: even if it works, leave it commented out, not sure if we're gonna get rate limited or something

vasfvitor: oh, and I'm not so sure but this function would need to be async, maybe we can get away from it if we create a small script that will fetch the data on `pnpm dev:setup`  and create a static md with the necessary snippet, then when running `dev` or `build` it would be readily available, I haven't used Lunaria so I'm not sure this will be needed.

GuoJikun: > oh, and I'm not so sure but this function would need to be async, maybe we can get away from it if we create a small script that will fetch the data on `pnpm dev:setup` and create a static md with the necessary snippet, then when running `dev` or `build` it would be readily available, I haven't used Lunaria so I'm not sure this will be needed.

you are right. It's possible to generate static html up front, but you'll need to update the page every time there's a new i18n-related PR.

GuoJikun: Asynchrony in js is difficult to eliminate, and slots in Lunaria must not be asynchronous functions.

vasfvitor: > but you'll need to update the page every time there's a new i18n-related PR.


Ah, you're right, I forgot about that detail

vasfvitor: what if we make an webcomponent? It should be able to render sync and once the promise resolve, update the content.

vasfvitor: ughhh, this might not so straight forward as I'd thought... Do you have any ideas?

GuoJikun: > https://api.github.com/repos/tauri-apps/tauri-docs/pulls?state=open&per_page=100

`https://api.github.com/repos/tauri-apps/tauri-docs/pulls?state=open&per_page=100`

I tried to put the api in the html, but the api was throwing errors because of cross-origin issues.
I recommend dropping the "[Translations that need reviews](https://tauri.app/contribute/translate-status#needs-review)" module

GuoJikun: > ughhh, this might not so straight forward as I'd thought... Do you have any ideas?

`https://api.github.com/repos/tauri-apps/tauri-docs/pulls?state=open&per_page=100`

I tried to put the api in the html, but the api was throwing errors because of cross-origin issues.
I recommend dropping the "[Translations that need reviews](https://tauri.app/contribute/translate-status#needs-review)" module.
Or wait for other ideas to become available.

vasfvitor: we can drop it fow now and create a new issue so we can track

vasfvitor: >I tried to put the api in the html, but the api was throwing errors because of cross-origin issues.

how come?

GuoJikun: > we can drop it fow now and create a new issue so we can track

Agree, i will remove `Translations that need reviews` module.

GuoJikun: > how come?

image
There should be a restriction on csp in the documentation.

vasfvitor: ah indeed there's a headers file, I forgot about that. Assuming we're removing `Translations that need reviews` we don't need to update the headers right? If that's the only the issue then we probably can work around it. https://github.com/tauri-apps/tauri-docs/blob/v2/public/_headers

GuoJikun: > ah indeed there's a headers file, I forgot about that. Assuming we're removing `Translations that need reviews` we don't need to update the headers right? If that's the only the issue then we probably can work around it. https://github.com/tauri-apps/tauri-docs/blob/v2/public/_headers

yes.

it work if add `connect-src 'self' https://api.github.con` to file `public/_header`.
image

vasfvitor: so  you can add the necessary changes and for review we'll invite someone from security to make sure it's ok. Assuming you're using `fetch`, adding this line should fix `connect-src 'self' https://api.github.com`. Ofc if it makes sense you can make a separated package and it will not have any limitation by default.

vasfvitor:
@tweidinger would you approve the `_headers` modification? just to ensure there's no pitfall on this setting https://github.com/tauri-apps/tauri-docs/pull/2976/files#diff-36e1aa26d55b829e1d4d1cc909c9c9d34972deeedf4ad4d631e5bff6842887d4

FabianLars:
can we revert the formatting changes of the headers file? 😅 looks kinda broken even if it still works.

lunaria/components.ts
```diff
@@ -0,0 +1,68 @@
+import { html } from '@lunariajs/core';
+
+export const TitleParagraph = () =>
+  html`<p>
+      If you're interested in helping us translate
+      <a href="https://v2.tauri.app/">v2.tauri.app</a> into one of the languages listed below,
```
dreyfus92:
```suggestion
      <a href="https://tauri.app/">tauri.app</a> into one of the languages listed below,
```

dreyfus92 (commented):
i just left a tiny nit 😁

public/_headers
```diff
@@ -1,3 +1,3 @@
 /*
   X-Content-Type-Options: nosniff
-  Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'; img-src self data: http: https:; script-src 'self' 'unsafe-inline' 'wasm-unsafe-eval' ; script-src-elem 'self' 'unsafe-inline'; worker-src 'self' data:; frame-src 'self' https://app.netlify.com https://www.youtube-nocookie.com;  frame-ancestors 'none';
+  Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'; img-src self data: http: https:; script-src 'self' 'unsafe-inline' 'wasm-unsafe-eval' ; script-src-elem 'self' 'unsafe-inline'; worker-src 'self' data:; frame-src 'self' https://app.netlify.com https://www.youtube-nocookie.com; frame-ancestors 'none'; connect-src 'self' https://api.github.com;
```
vasfvitor:
```suggestion
  Content-Security-Policy: default-src 'self'; style-src 'self' 'unsafe-inline'; img-src self data: http: https:; script-src 'self' 'unsafe-inline' 'wasm-unsafe-eval' ; script-src-elem 'self' 'unsafe-inline'; worker-src 'self' data:; frame-src 'self' https://app.netlify.com https://www.youtube-nocookie.com; frame-ancestors 'none'; connect-src 'self' https://api.github.com/repos/tauri-apps/*;
```

let's play safe and limit calls only to tauri-apps. I think it needs a wildcard so it works

vasfvitor (approved):

public/_headers
GuoJikun: > let's play safe and limit calls only to tauri-apps. I think it needs a wildcard so it works

It`s not work.

Changed to `connect-src 'self' https://api.github.com/repos/tauri-apps/tauri-docs/pulls`

dreyfus92 (approved):
LGTM, hope this doesn't break anything 🫡

FabianLars:
> LGTM, hope this doesn't break anything 🫡

That would be super boring 😠

FabianLars (approved):

PR #2996, opened 2024-11-17
fix(docs): fix and simplify Nix devShell

max-privatevoid:
#### Description

I'm sick and tired of seeing people ask why they can't build their Tauri app on NixOS and showing up with this absolute monstrosity of a devShell. I call them out for using `LD_LIBRARY_PATH` (I hope I don't have to explain to you why) and their defense is "Tauri devs told me to do it". Enough.

This PR vastly simplifies the example devShell provided in the docs, along with some misc. fixes. Specifically:
- It now makes proper use of `nativeBuildInputs` to allow the correct setup hooks to be called, that way you don't need to set `PKG_CONFIG_PATH` yourself.
- It now also gives you `cargo`, `cargo-tauri` and `nodejs` from Nixpkgs, because why use horrible toolchain managers like `rustup` and abuse language package mangers for things they weren't meant for when you're already using the best package manager ever? A note has been added to the top of the section to inform users about this.
- Removed the superfluous explicit usage of `dev` outputs. `mkShell` (`mkDerivation` to be exact) does this automatically.
- Removed all the superfluous OpenSSL-specific environment variables. I think that was intended for the `openssl-sys` crate. Unnecessary, because now it will simply use `pkg-config` to find the library.

TODO:
- The devShell is also present in the Chinese translation of the docs. I've copied my fixed version over there, but I haven't included the note. Would it be fine to just have the note in English there, or should it be translated?

### DIFF
--- a/src/content/docs/start/prerequisites.mdx
+++ b/src/content/docs/start/prerequisites.mdx
@@ -116,6 +116,10 @@ sudo apk add \
   </TabItem>
   <TabItem label="NixOS">
 
+:::note
+This will also install Rust and Node.js as well as the `cargo-tauri` CLI for you, so you can skip those steps below.
+:::
+
 Using `nix-shell`:
 
 ```nix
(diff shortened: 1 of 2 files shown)


FabianLars:
Thanks for the PR!

> and their defense is "Tauri devs told me to do it"

Sounds like we should annotate the setup instructions with something explaining that this part of the docs is community maintained and not verified by the tauri team (none of us actively use nix).

> It now also gives you cargo, cargo-tauri and nodejs from Nixpkgs, because why use horrible toolchain managers like rustup and abuse language package mangers for things they weren't meant for when you're already using the best package manager ever? A note has been added to the top of the section to inform users about this.

i don't disagree but our docs make use of rustup in some places to install additional targets and people proved to us that they can't translate rustup instructions themselves (not a nix specific thing, also applies to homebrew for example). Is this something that could be added to the note? idk how easy/hard it is to add additional targets in this context.

> I think that was intended for the openssl-sys crate. Unnecessary, because now it will simply use pkg-config to find the library.

yes, the PR author based it on https://discourse.nixos.org/t/openssl-libdir-at-nix-store-id-openssl-3-0-13-dev-lib-does-not-contain-the-required-files-to-either-statically-or-dynamically-link-openssl/48582 
And i am not really familiar with nix but the non-working example in there looks similar to your PR now. I'd appreciate it if you could test whether openssl-sys still works if not done already.

> The devShell is also present in the Chinese translation of the docs. I've copied my fixed version over there, but I haven't included the note. Would it be fine to just have the note in English there, or should it be translated?

the english note is fine

max-privatevoid:
> i don't disagree but our docs make use of rustup in some places to install additional targets and people proved to us that they can't translate rustup instructions themselves (not a nix specific thing, also applies to homebrew for example). Is this something that could be added to the note? idk how easy/hard it is to add additional targets in this context.

Nixpkgs supports a variety of Rust targets. Using custom targets generally requires knowledge of how cross-compiling works in Nixpkgs. I don't know if you want to teach that.

> yes, the PR author based it on https://discourse.nixos.org/t/openssl-libdir-at-nix-store-id-openssl-3-0-13-dev-lib-does-not-contain-the-required-files-to-either-statically-or-dynamically-link-openssl/48582
And i am not really familiar with nix but the non-working example in there looks similar to your PR now. I'd appreciate it if you could test whether openssl-sys still works if not done already.

The fuckup there was probably a similar one. They set `OPENSSL_DIR` and `PKG_CONFIG_PATH` for no reason. The `openssl` crate docs document this behavior: https://docs.rs/openssl/latest/openssl/#automatic

The given devShell is capable of building a basic Rust project with the `openssl` crate as-is, and it produces a correct binary.

Also note that whatever project template I created with `cargo tauri init` doesn't seem to depend on OpenSSL, so it might not need to be included. I kept it there because the instructions for other systems also provide it.

> the english note is fine

Will copy.

FabianLars (approved):

PR #3023, opened 2024-11-26
i18n(zh-cn): translate `about` module

GuoJikun:
#### Description

- Closes #


GuoJikun:
image
I don't know what should be translated here.

Maybe you can explain it in English.

FabianLars:
> I don't know what should be translated here.
> Maybe you can explain it in English.

No idea either 😂 The last part of that section isn't really correct anymore either (in the english source)

GuoJikun:
> The last part of that section isn't really correct anymore either (in the english source)
image
here?

FabianLars:
No this part
{1F560F87-F5A9-4E9D-A476-00FD5F968CDE}

> Today, Tauri uses Rust for the backend - but in the not too distant future, other backends like Go, Nim, Python, Csharp, etc. will be possible. This is because we are maintaining the official Rust bindings to the [webview](https://github.com/webview) organization and plan to let you switch out the backend for your needs. Since our API can be implemented in any language with C interop, full compliance is only a PR away.

We very likely won't completely switch out rust, but have the second backend "attached" to the rust one. Also, we're not the webview bindings maintainers anymore.

FabianLars:
I assume with the "Merge branch v2" commits that you consider this ready? Since there were no reviews we'll do as said on discord :)

FabianLars (approved):

PR #3034, opened 2024-11-28
fix(docs): fix reference to tokio mutex

pebogi:
#### Description
The reference for tokio's Mutex implementation was pointing to the std implementation instead. The updated version points to the page that is implied by the sentence.

### DIFF
--- a/src/content/docs/develop/state-management.mdx
+++ b/src/content/docs/develop/state-management.mdx
@@ -106,7 +106,7 @@ For more information on commands, see [Calling Rust from the Frontend](/develop/
 
 #### Async commands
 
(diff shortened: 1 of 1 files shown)


vasfvitor (commented):
hey, thanks for this, note that you need to update the link at the bottom of the file. Usually it's better to have the links grouped down there instead of inline

pebogi:
Hi, updating the mutex link at the bottom of the document would break all references that right now correctly point to the std mutex implementation. Also there is already another use of inline linking ([interior mutability]) in the document, so it made sense for me to use that for this single link. If you still prefer moving this to the bottom over inline I could use something like [Tokio's async `Mutex`] as the anchor. Which option do you prefer?

vasfvitor (approved):
great, it's ok then

PR #3070, opened 2024-12-18
fix PluginLinks undefined in language zh-cn

lerte:
PluginLinks has undefined url path in language zh-cn
image


vasfvitor (approved):
thank you

PR #3087, opened 2024-12-26
i18n(zh-cn): fix chinese docs url

coffeemil:
fix zh-cn docs url


dreyfus92 (approved):
Nice catch and thank you for fixing this @coffeemil 😁. Welcome to `tauri-docs` 🫡

PR #3098, opened 2025-01-03
Correct path for `bundleMediaFramework` config

jonbng:
The "path" of the bundleMediaFramework config in tauri.conf.json is specified wrongly I think. This fixes that, it should be  tauri.conf.json > bundle > linux > appimage > bundleMediaFramework.


FabianLars (approved):
thank you, i messed up copying it from the v1 docs 🙃

PR #3100, opened 2025-01-04
Typo in npm installation command

ThomasBurgess2000:
#### Description
Changes `npm install --save-dev @tauri-apps/cli@">1.0.0"` to `npm install --save-dev @tauri-apps/cli@^1.0.0`

### DIFF
--- a/src/theme/Command.js
+++ b/src/theme/Command.js
@@ -116,7 +116,7 @@ export const InstallTauriApi = () => {
     <Tabs groupId="package-manager">
       <TabItem value="npm" default>
         <CodeBlock className="language-shell" language="shell">
-          npm install @tauri-apps/api@^1.0.0
+          npm install @tauri-apps/api@1
         </CodeBlock>
       </TabItem>
       <TabItem value="Yarn">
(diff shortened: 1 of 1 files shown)


FabianLars:
There's one issue with this. `^1.0.0` will make it actually install 1.0.0 (in the lockfile) while `">1.0.0"` will be replaced with `^1.6.0`. What should work better than both would be `npm install @tauri-apps/api@1` - it's short and behaves like `">1.0.0"`

ThomasBurgess2000:
> There's one issue with this. `^1.0.0` will make it actually install 1.0.0 (in the lockfile) while `">1.0.0"` will be replaced with `^1.6.0`. What should work better than both would be `npm install @tauri-apps/api@1` - it's short and behaves like `">1.0.0"`

at least on windows `">1.0.0"` creates a file named `1.0.0`, which is what I was trying to prevent

updated to your suggestion though

FabianLars:
> at least on windows ">1.0.0" creates a file named 1.0.0, which is what I was trying to prevent

not sure what you mean tbh, i'm on windows too. but i guess it doesn't matter anymore with the new syntax. thanks for contributing :)

FabianLars (approved):

PR #3108, opened 2025-01-07
fix(file-system):  change BaseDirectory.App to BaseDirectory.AppData

klren0312:
#### Description

- in the `path api`, `App` not in the `BaseDirectory`, https://github.com/tauri-apps/tauri/blob/dev/packages/api/src/path.ts#L19C1-L44C1, so need to change the doc


FabianLars (approved):
Whoops we copied that badly from v1. Thanks for the fix :)

PR #3118, opened 2025-01-13
docs(state): Add note about the use of apphandles to move state into threads

FabianLars:


FabianLars:
not too happy about the phrasing being so thread specific. also, i'm aware of the existing note in the section i actually link to but people miss it and i guess it also fits to the Arc topic.

PR #3129, opened 2025-01-22
add v1 flag to v1 docs cta

### LINKED ISSUE #3036: Is it necessary to add the "--tauri-version 1" flag for create-tauri-app on the v1 documentation websites?
I'm not sure if creating a Tauri app with `create-tauri-app` requires a specific version (it's also possible that the default version created is the latest （ tauri v2 and later）). If it does, I believe we should add a fixed version flag "--tauri-version 1" to the quick start and homepage prompts on the v1 documentation website. For example, "pnpm create tauri-app --tauri-version 1".
↓↓↓
https://v1.tauri.app/v1/guides/getting-started/setup

vasfvitor:
closes #3036

is the bash command correct?

### DIFF
--- a/src/theme/Command.js
+++ b/src/theme/Command.js
@@ -28,38 +28,38 @@ export const CreateTauriApp = () => {
     <Tabs groupId="package-manager">
       <TabItem value="Bash">
         <CodeBlock className="language-shell" language="shell">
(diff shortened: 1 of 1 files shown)


FabianLars:
> is the bash command correct?

well, it works at least 😅

FabianLars (approved):

PR #3158, opened 2025-02-05
Corrects read only example for open function

ChristianPavilonis:
#### Description

The current example will read nothing because the U8Array is initialized with length of 0 because `file.read()` reads up to the bytelength of the buffer.

This example works, or the example in the JavaScript doc block of `new Uint8Array(100)` would work.


vasfvitor (approved):
thank you

PR #3111, opened 2025-01-08
i18n(ja): "Quick Start" Section

J1-takai:
#### i18n (JP) Translation for "Quick Start" section
- What does this PR change? Give us a brief description. 
  This is a trial translation work for "Tauri Doc 2.0 Japanese Version" from the scratch. If there are any similar activities already on-going, please disregard this pull request. If not, my drafts would need technical proof-reading as I have almost no knowledge on programming -  my only programming experiences were introductory courses for FORTRAN and COBOL using punch cards data loading some 40 years ago, HA HA!  No rush - Please take your time. 
- Closes #

PR #3168, opened 2025-02-11
i18n(ja): japanese translate for top page

penginnu:
#### Description

- Added Japanese translation for the top page.

#### Note

- This is a duplicate of the PR #3111 , but is submitted because it was translated separately. You can have them integrated.
- I have confirmed in PR #3111 that the directory is changed from jp to ja, but we have created the directory as jp for confirmation.


vasfvitor:
> from jp to ja, but we have created the directory as jp for confirmation
It should be ja, it's now updated on the main repo. Thanks.

> This is a duplicate of the PR 

I'll see what can be done

Thanks again

PR #3184, opened 2025-02-27
Adding in the file-system plugin some notes regarding Rust/JS API differences

charlesschaefer:
#### Description

Just adding some comments on the file-system plugin, to clarify that tauri doesn't have an API for file manipulation in the Rust side, and the traditional libs should be used instead.


dreyfus92 (approved):
thank you @charlesschaefer for taking the time in adding these notes 😁, LGTM ✅

PR #3197, opened 2025-03-09
i18n(ja):  Add Japanese texts for Concept section

J1-takai:
#### Description

- 7 files in Concept section were translated into Japanese.
- JP Doc Status List file is also updated (but this may/can be replaced by a similar .xls file).
- Closes #


vasfvitor:
just need to fix one hash link, I'll come back to it

J1-takai:
@vasfvitor Thank you always for your kind support on my JA translation "project" :+1: :+1: :+1:  and typo corrections in links and html tags. I will be more careful on those mistakes.

PR #3207, opened 2025-03-14
feat: commands can be defined as pub in crate root on tauri 2.4

lucasfernog:
ref https://github.com/tauri-apps/tauri/pull/12979


lucasfernog:
this should wait for tauri 2.4 release

PR #3238, opened 2025-04-05
Mention 'wasm-unsafe-eval' in the CSP documentation.

TimNN:
I just spent a very frustrating hour trying to figure out why my Tauri app wasn't loading, because Safari apparently can't be bothered to log any kind of warning if the CSP blocks WebAssembly execution (it does log warnings for other CSP issues).

I'm not sure if this is the best way to communicate this information, but I decided to make the proposal.


vasfvitor (approved):
thanks for this

PR #3250, opened 2025-04-10
Update imports in base-level menu rust implementation

sebastinez:
#### Description
Fixes an import statement in the base-level menu example.
Instead of importing `tauri::menu::{Menu, MenuItem}` which aren't used in the example, it should `MenuBuilder` which is used.


dreyfus92 (approved):
Nice catch, thank you 😁

PR #3258, opened 2025-04-15
Add on old version notice

Legend-Master:
#### Description

- Closes https://github.com/tauri-apps/tauri/issues/13233

### DIFF
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -299,6 +299,7 @@ async function siteConfig() {
               current: {
                 label: unreleasedTauriVersion,
                 path: unreleasedTauriVersion,
+                banner: 'unmaintained',
               },
               // If there is a "latest" version, map url to version number
               ...(lastestReleasedVersion && {
--- a/docusaurus.config.js
+++ b/docusaurus.config.js
@@ -225,7 +225,7 @@ const footerLinks = [
 async function siteConfig() {
   const mdxMermaid = await import('mdx-mermaid')
   return {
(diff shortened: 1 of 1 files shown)


Legend-Master:
Well, this banner doesn't seem to work, as it still shows the latest to be v1

image

Guess we change the announcement instead

vasfvitor (approved):
image

vasfvitor:
maybe we could go a step further and update the title from "Tauri Apps" to "Tauri v1"

Legend-Master:
> maybe we could go a step further and update the title from "Tauri Apps" to "Tauri v1"

Changed

Also I don't have permission to merge to v1 branch, so maybe you could merge it instead 🙃

RazielMartelus96:
As a humble suggestion, whilst the PR has already been merged and closed, I felt it sensible to mention the light blue color does compliment Tauri's UI... a Deprecation warning is not designed to "fit in" in my humble opinion.

It's purpose is to make it abundantly clear it is the V1 version, and hence should not blend in with the UI color pallet. Rather, it should do the opposite and be direct and contrasting (For example, a shade of red).

I hope this is not seen as pedantic, and indeed does have a precedent, such as with old versions of unity documentation: https://docs.unity3d.com/2017.4/Documentation/Manual/index.html

Not too invested in dying on this hill, but thought it was at least worth mentioning as, on investigation of your link, I didn't initially notice the warning.

That said thank you for the contribution and it is very much appreciated!

PR #3260, opened 2025-04-15
Update environment-variables.mdx

inlancersystem4:
#### Description

- Closes #


inlancersystem4:
spelling mistake change

FabianLars:
this was not a spelling mistake. the env var is a dynamic version of https://doc.rust-lang.org/std/env/consts/constant.FAMILY.html so unix was correct, and yes it will show unix for both linux and macos targets.

PR #3281, opened 2025-05-02
i18n(ja) Add "Dev/plugins" section

J1-takai:
#### Description
i18n(ja): Add Japanese Translation to "dev/plugins" section
- Two files were translated into Japanese
- Closes #


src/content/docs/ja/develop/plugins/index.mdx
```diff
@@ -0,0 +1,474 @@
...
+  let denied = command_scope.denies();
+  todo!()
+}
+```
+
+###### グローバル・スコープ
+
+アクセス権に、「許可」または「拒否」するコマンドが定義されていない場合は「スコープ権限」とみなされ、プラグインにはグローバル・スコープのみが定義されます：
+
+> > > 《訳注》 **スコープ権限**　正規訳不明。原文 it’s considered a scope permission（文意不詳）。
```
vasfvitor: could you elaborate on that comment? is it supposed to be displayed? or do you need help to translate?

if it helps: scope is where (is permitted) and permission is what (is permitted or not)

J1-takai: @vasfvitor 
This "《訳注》translator's note" section is intended to show why the term "scope permission" is translated as shown. It seems no "fixed terminology" for the term "Scope Permission" in Japanese according to the Microsoft's Terminology Search site, and I'm not sure if my "direct" translation makes any sense for the readers/programmers. So this part is intended to suggest them that other translations may be available. I am OK to hide this translator's note as a comment in the .mdx file as the future review point for translation contributors so that the readers may not be disturbed by the note. Thanks for your comments.

vasfvitor: Leave the note visible then, seems reasonable. I'll just find a way to improve how it appears visually, I think it could be a muted color and a tad smaller font

image
image

J1-takai: @vasfvitor 
THX & :+1:

vasfvitor: @J1-takai what do you think?
image
image

J1-takai: @vasfvitor 
Thank you for this improvement on the "TranslationNote" format. It looks nicer in the Tauri Docs. I will start applying this method (ie. use of "import TranslationNote ..." &  "TranslatioNote" tags.

One Note: in the index.mdx, two tags are required; on line 252 (the ending tag \</TranslationNote>) and on line 255 (the beginning tag \<TranslationNote>), because the line 253 is the original Tauri text.

vasfvitor: ok thx, fixed

J1-takai:
@vasfvitor 
Thank you always for your careful verifications of my files. :+1: :+1: :+1:

J1-takai:
@vasfvitor 
I just noticed that these two newly merged files are not shown on the **Menu column** and also the **file navigation** boxes (ie "to the previous file" and "to the next file") at the bottom of each file are missing. How can I correct these?

vasfvitor:
@J1-takai thanks for the heads up, I'll check it and fix on the main branch

vasfvitor:
fixed. The folder casing matters "plugins" should be "Plugins"

J1-takai:
@vasfvitor 
Thanks for your quick actions. :smiley:

J1-takai:
@vasfvitor 
Thanks for your quick actions. :smiley_cat:

PR #3289, opened 2025-05-08
Describe Ad-hoc signing on MacOS

matthewCmatt:
### Description

Adds a section to the MacOS Signing guide describing "ad-hoc" signing.

Also tweaks the markup for a few other code snippets to add `sh` coloring.

Relates to issue: https://github.com/tauri-apps/tauri/issues/8763

### Background

I ran into lots of confusion with this, as an older build of my app built by GitHub worked fine when downloaded, but when I rolled back to the same revision, the build output from GitHub failed to start on MacOS. This was because the original build was carried out on an Intel MacOS runner, and the newer build ran on an Apple Silicon runner.

On opening the newly generated build, I got the "App Is Damaged And Can’t Be Opened" error, even though my app wasn't signed. Turns out MacOS on ARM _requires_ a signature for apps to run, and my app was quarantined as it lacked this signature. I haven't set up the identity with my organization's Apple Developer account yet, but I wanted the build to still be runnable.

Setting the identity to the pseudo-identity "-" allowed the app to run on an M3 MBP, though I still needed to whitelist the app in Privacy & Security settings. This is consistent with the prior behavior for "unsigned" apps built by GitHub.

### DIFF
--- a/src/content/docs/distribute/Sign/macos.mdx
+++ b/src/content/docs/distribute/Sign/macos.mdx
@@ -195,6 +195,7 @@ Ad-hoc code signing does not prevent MacOS from requiring users to
 :::
 
 To configure an ad-hoc signature, provide the pseudo-identity `-` to Tauri, e.g.
+
 ```json
 "signingIdentity": "-"
 ```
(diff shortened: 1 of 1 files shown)


FabianLars:
thank you very much :) a simple `pnpm format` and this is good to go 👍

FabianLars (approved):

PR #3293, opened 2025-05-12
adjust doc

Kalmaegi:
#### Description
Consolidate all Serde renaming into a single top-level annotation (rename_all = "camelCase" + rename_all_fields = "camelCase") to remove redundant per-variant attributes and reduce boilerplate – keeps behavior identical while making the code more concise and maintainable

No change to the serialized JSON output; only internal attribute placement is modified


FabianLars (approved):

FabianLars:
thanks :)

PR #3302, opened 2025-05-26
Changed NixOS related parts in prerequisites

ukizet:
#### Description
Changed NixOS tab in Linux heading (moved from nix-shell to devenv). Changed Android heading, added there subheading NixOS (added devenv code).

https://devenv.sh/

### DIFF
--- a/src/content/docs/start/prerequisites.mdx
+++ b/src/content/docs/start/prerequisites.mdx
@@ -442,7 +442,7 @@ Using [devenv](https://devenv.sh/):
   enterShell = ''
     echo NDK_HOME=$NDK_HOME
     git --version
-    echo To create emulator lust launch "emulator"
+    echo To create emulator just launch "emulator"
   '';
 }
 ```
--- a/src/content/docs/start/prerequisites.mdx
+++ b/src/content/docs/start/prerequisites.mdx
@@ -120,35 +120,51 @@ sudo apk add \
 This will also install Rust and Node.js as well as the `cargo-tauri` CLI for you, so you can skip those steps below.
 :::
 
-Using `nix-shell`:
+Using [devenv](https://devenv.sh/):
 
 ```nix
-let
-  pkgs = import <nixpkgs> { };
-in
-pkgs.mkShell {
-  nativeBuildInputs = with pkgs; [
-    pkg-config
-    gobject-introspection
-    cargo
-    cargo-tauri
-    nodejs
-  ];
+{ pkgs, config, ... }:
 
-  buildInputs = with pkgs;[
+{
+  env = {
+    PATH = "$HOME/.cargo/bin:$PATH";
+  };
+
+  languages = {
+    javascript.enable = true;
+  };
+
+  packages = with pkgs; [ 
+    git
+    pkg-config
     at-spi2-atk
     atkmm
     cairo
     gdk-pixbuf
     glib
     gtk3
+    webkitgtk_4_1
     harfbuzz
     librsvg
     libsoup_3
     pango
-    webkitgtk_4_1
     openssl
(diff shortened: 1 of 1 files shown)


FabianLars:
is devenv a part of nix? if not then imo this not a good change, but i also don't use nix so i wouldn't mind more opinions on this matter.

ukizet:
@FabianLars 

What do you mean by "part of nix"? It uses nix underhood

FabianLars:
What i mean is that people who installed nix already have nix-shell but i assume they have to install devenv themselves first to follow your PR. And unless devenv is considered the ecosystem standard (i don't know whether it is) i don't think this change here is a good idea.

ukizet:
With my experience in nix, NixOS and NixOS + Tauri. I can say that it's a lot simpler to install everything that you need to use Tauri, using devenv than using nix-shell or flake.devShells (this is especially true for Android development). For nix users, there is no struggle in installing devenv (https://devenv.sh/getting-started/#installation).

FabianLars:
Yeah, i totally believe that it's easier. But in my opinion it makes no sense to require another tool (that's not close to a standard tool) to get started at all, even if it's no struggle to install it.

I'm not totally against having it as an option but it should not be the only thing we offer.

ukizet:
I can make it the second thing that you offer to nix users.

I do not understand this philosophy with "standard tool". Why do you care? Why would others care? If others would want to use Tauri they would install everything that they need, even non "standard tool's", cause they want to use Tauri. With "standard tools" that nix provide you can't do much. Like Tauri itself isn't "standard tool". Tauri doesn't even use nix underhood, even smallest flake.nix is not there. So why nix users should use it? Sounds strange right?

ukizet:
Why in prerequisites there is Android Studio? It's not "standard tool" in Windows, Linux or MacOS. So probably users should not use it, right?

FabianLars:
> I can make it the second thing that you offer to nix users.

So we agree on that. I don't understand why the rest of your response turned out so toxic then to be honest.

I know i didn't explain it well so i try it one last time (with the words of a colleague).

> Tauri doesn't even use nix underhood, even smallest flake.nix is not there.

Tauri also doesn't use fedora and we still provide prerequisites for it. Consider the `nix` tab a `nix and nixOS` tab if that makes it clearer.

> Why in prerequisites there is Android Studio? It's not "standard tool" in Windows, Linux or MacOS. So probably users should not use it, right?

Android Studio is considered _the_ standard tool for developing Android apps. But that doesn't even matter since it's a direct prerequisite of Tauri.

So, Android Studio is required for Tauri Android development while devenv is **not**. It may be nice to use, but you can develop tauri on nix/nixos without it.
If it was either required by tauri or used by the **majority** of nix users then i would be totally fine with your PR here. But if most nix users don't use that tool, and may not even heard of it, it's not a good idea to completely remove the plain nix instructions.

Also, another point to consider is that the nix prereqs docs are community maintained so there's a probability that we'd end up with less contributions (because of less people with experience).

----

Apologies that i can't explain it better. Seeing that you also reached out on Discord this is my last response on the matter and i will let others take over.

vasfvitor:
as I can see, having both options is possible, I can imagine the intro with a summary more or less:

"You can setup prerequisites using `nix-shell` or , optionally with `devenv`. For devenv follow this instrcutions (link), for nix-shell:

...current nix-shell instructions

---



And the devenv instruction would be a new page you can create, more or less copying what you already done into a new nixos-prereq.md file. I'm not fan of having this because no other system have (maybe they could use more detailed instructions?), and it would be nice to keep it simple and less friction possible, like, to avoid decision paralysis, but also I think the more details/options, the better.

PR #3311, opened 2025-05-27
docs: shortcut is registered but does not goes to handler

gabrielmar:
#### Description
- Fix shortcut Rust example.


FabianLars:
are you sure the old example doesn't work? i used that in one of my test apps not so long ago and it worked. maybe something broke in the meantime but it's still supposed to work.

gabrielmar:
I tested the current example and it is correct, the problem I found was that after downloading the plugin it registers itself, if you copy only the setup method the shortcut is registered but doesn't invoke the handler

PR #3330, opened 2025-06-04
Contributing guide

kandrelczyk:
#### Description

Adding contributing guide to the documentation. The content is a mix of contribution guides from main tauri and tauri-docs repos.


kandrelczyk:
Looks like I didn't sync my fork before creating the branch but it's all new content so I guess it will merge without issues.

vasfvitor (changes requested):
Hi, thanks for this, it's great to have this well doc

I haven't gone through everything but I think issue reporting and pr guidelines should be each to the respective project. I can think of two possibilities: make it specific for the docs repo, or make a list/table for each repo own instruction on the matter / not copying to avoid having mixed information, so whatever doesn't originally belongs here, must be referenced. 

While I'm specifying this about issue-reporting.mdx and pr.mdx file, others may have something I missed, but consider it as whole: if I recall we already have one issue guideline for the docs that could be used/linked, and on the docs we don't do releases nor versioning, so some instructions doesn't make sense here and may not make sense for the main repo, or the plugins-workspace repo and others.

Could you elaborate more as the source of these contents? Just so I have a reference about accuracy? I imagine you may have changed a few stuff here and there to adapt but it would be nice to have the original to ensure nothing is lost/changed by mistake

Thanks again, I've been thinking of having something like this for a while

kandrelczyk:
Hi @vasfvitor ,

The info comes from here:

https://github.com/tauri-apps/tauri/blob/dev/.github/CONTRIBUTING.md

and

https://raw.githubusercontent.com/tauri-apps/tauri-docs/refs/heads/v2/.github/CONTRIBUTING.md

I will add some comments marking the main changes.

As for PR and Issue reporting guidelines I don't really see points that need more context. For PRs I think all rules are pretty basic and applicable to all repos (at least I'm following them for tauri, plugin-workspace and tauri-docs. For Issue reporting I remove the first point which was confusing. I think other points are valid for all repos. The are not as relevant for tauri-docs but I don't think there's anything confusing or incorrect there.

vasfvitor:
@dreyfus92 any suggestion here? what do you think?

dreyfus92 (commented):
Huge thanks @kandrelczyk  for putting this comprehensive guide together. This is exactly the kind of contribution that makes our ecosystem stronger! And special thanks to @vasfvitor for looping me in on this discussion.

I don't think we need the `coc.mdx` file here, the Code of Conduct should be part of the main Tauri repo or handled at the organization level through `.github` files.

More broadly, I think we should focus this contribution guide specifically on documentation contributions rather than trying to cover all Tauri repositories. The main repo already has its own contributing guidelines, and trying to maintain duplicate/overlapping guidance across repos creates potential for inconsistency.

I'd suggest we take inspiration from Astro's approach with their dedicated docs contribution site: https://contribute.docs.astro.build/welcome/ - they keep it focused specifically on documentation contributions with clear, actionable guidance for docs contributors.

FabianLars:
> More broadly, I think we should focus this contribution guide specifically on documentation contributions rather than trying to cover all Tauri repositories. The main repo already has its own contributing guidelines, and trying to maintain duplicate/overlapping guidance across repos creates potential for inconsistency.

I agree with this when it comes to what's shown on the website but please don't discard the other guides here and instead upstream them into the respective repositories 🙏 

If we then think we should still show those guides on the website we can pull them from the repos automatically so we don't have to sync them manually.

kandrelczyk:
I removed the coc.mdx file and lined to the github repo instead, no issues here.

Now for other points. Astro also has this: https://docs.astro.build/en/contribute/ and this is more what I was aiming for. If we prefer to keep guides in specific repos I will move my changes there and only link to the repositories in docs. 

As for  https://contribute.docs.astro.build/welcome/  I see it as a separate initiative. For now tauri-docs repo has it's own CONTRIBUTING.md file and I will link to it. To create separate guide a new astro page would have to be created with separate deployment plan, right? If someone with proper permissions can prepare a basic template (just something that will be compiled and deployed under sub-domain) I can work on the content.

kandrelczyk:
This is new version. I will upstream changes to the core repo and plugins-workspace.

kandrelczyk:
Changes upstreamed in https://github.com/tauri-apps/plugins-workspace/pull/2753 and https://github.com/tauri-apps/tauri/pull/13594

dreyfus92 (approved):
Thank you @kandrelczyk, LGTM 😁. I'd like @FabianLars and @vasfvitor to give this a last review 🤘🏻

package-lock.json
vasfvitor, line 1: seems like you used npm? you can remove this file

kandrelczyk:
Remove some unused files and fixed formatting

vasfvitor (approved):

FabianLars (approved):

PR #3334, opened 2025-06-06
Add warning about keeping Single instance initialization as the first one

charlesschaefer:
#### Description

Adds a warning about keeping single instance init() call as the first one, before other plugins' initialization.

### DIFF
--- a/src/content/docs/plugin/single-instance.mdx
+++ b/src/content/docs/plugin/single-instance.mdx
@@ -70,6 +70,13 @@ Install the Single Instance plugin to get started.
 
 </Tabs>
 
+
+:::info
+
+The Single Instance plugin must be the first one to be init()'ed to work well. Doing so prevents other plugins to initiate and eventually interfering with the app's instance that is already running.
+
+:::
+
 ## Usage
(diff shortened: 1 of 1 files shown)


src/content/docs/plugin/single-instance.mdx
```diff
@@ -70,6 +70,13 @@ Install the Single Instance plugin to get started.
 
 </Tabs>
 
+
+:::info
+
+The Single Instance plugin must be the first one to be init()'ed to work well. Doing so prevents other plugins to initiate and eventually interfering with the app's instance that is already running.
```
FabianLars:
```suggestion
The Single Instance plugin must be the first one to be registered to work well. This assures that it runs before other plugins can interfere.
```

src/content/docs/plugin/single-instance.mdx
```diff
@@ -70,6 +70,13 @@ Install the Single Instance plugin to get started.
 
 </Tabs>
 
+
```
FabianLars:
```suggestion
```

PR #3335, opened 2025-06-06
Add notice that the restart is about pwsh not sys

Legend-Master:
#### Description

- Closes # 

I got confused for a second if that's asking me to restart my computer 😂

Also changed the folder selection to select the latest one in case there're multiple version of NDK installed

### DIFF
--- a/src/content/docs/start/prerequisites.mdx
+++ b/src/content/docs/start/prerequisites.mdx
@@ -356,8 +356,8 @@ $VERSION = Get-ChildItem -Name "$env:LocalAppData\Android\Sdk\ndk" | Select-Obje
 ```
 
 :::tip
-PowerShell will not pick up the new environment variables until reboot or logout (the PowerShell, not your system),
-however you can refresh the current session:
+Most apps don't refresh their environment variables automatically, so to let them pickup the changes,
+you can either restart your terminal and IDE or for your current PowerShell session, you can refresh it with
 
 ```ps
 [System.Environment]::GetEnvironmentVariables("User").GetEnumerator() | % { Set-Item -Path "Env:\$($_.key)" -Value $_.value }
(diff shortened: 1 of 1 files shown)


src/content/docs/start/prerequisites.mdx
```diff
@@ -351,16 +351,18 @@ export NDK_HOME="$ANDROID_HOME/ndk/$(ls -1 $ANDROID_HOME/ndk)"
...
-$VERSION = Get-ChildItem -Name "$env:LocalAppData\Android\Sdk\ndk"
+$VERSION = Get-ChildItem -Name "$env:LocalAppData\Android\Sdk\ndk" | Select-Object -Last 1
 [System.Environment]::SetEnvironmentVariable("NDK_HOME", "$env:LocalAppData\Android\Sdk\ndk\$VERSION", "User")
 ```
 
 :::tip
-PowerShell will not pick up the new environment variables until reboot or logout,
-However you can refresh the current session:
+PowerShell will not pick up the new environment variables until reboot or logout (the PowerShell, not your system),
+however you can refresh the current session:
```
FabianLars: This is more in line of the old style but without the incorrect assumption
```suggestion
PowerShell will not pick up the new environment variables until it is restarted,
however you can refresh the current session:
```
the `however` sounds a bit wrong there, so maybe something like:
```
PowerShell will not pick up the new environment variables until it is restarted,
or until you manually refresh the session:
```
or:
```
To make PowerShell pick up the new environment variables you can either restart it
or run this command:
```

i'm not too happy with any of them tbh but the PR's suggestion basically tries to correct an incorrect statement (reboot) visibly without actually fixing it if that makes sense?

Legend-Master: Makes sense to me, the first thing popped up in my head was this should be about the powershell not the system, so that's why I edited it like this, I think maybe:

```md
Most apps don't refresh their environment variables automatically, so to let them pickup the changes, you can either restart your terminal and IDE or for your current PowerShell session, you can refresh it with

[System.Environment]::GetEnvironmentVariables("User").GetEnumerator() | % { Set-Item -Path "Env:\$($_.key)" -Value $_.value }
```

FabianLars: It's a bit long but works for me until we get an english professor to proof read our docs 😂

FabianLars (approved):

PR #3352, opened 2025-06-12
Add a note on plugin command scheduling on Android

DeTeam:
#### Description

This kind of note can be helpful to anyone hoping that Tauri will do some magic to prevent UI freezes, given that commands' `Invoke` API  implies support for async behavior and longer-running operations.

### DIFF
--- a/src/content/docs/develop/Plugins/develop-mobile.mdx
+++ b/src/content/docs/develop/Plugins/develop-mobile.mdx
@@ -216,6 +216,20 @@ class ExamplePlugin(private val activity: Activity): Plugin(activity) {
 }
 ```
 
+:::note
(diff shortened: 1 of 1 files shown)


DeTeam:
@FabianLars @lucasfernog what do you think?

src/content/docs/develop/Plugins/develop-mobile.mdx
```diff
@@ -260,6 +260,20 @@ impl<R: Runtime> <plugin-name;pascal-case><R> {
 }
 ```
 
+:::note
```
FabianLars: should this block perhaps be in the android block above so it's next to the other kotlin code?

DeTeam: good point, moved! 👌

FabianLars (approved):

PR #3363, opened 2025-06-22
fix: sidecar arguments documentation not mentioning true and false values

asdolo:
#### Description

This PR updates the [Sidecar](https://v2.tauri.app/develop/sidecar/) documentation page to mention a missing (but important) catch when declaring arguments for a sidecar.

As [the source code mentions](https://github.com/tauri-apps/tauri-plugin-shell/blob/v2/build.rs#L46) (and personally tested), a value of `true` for the `args` key in the Capabilities file will allow any arguments to be passed to the command, while `false` will disable all arguments. This is not mentioned in the current version of the docs.


FabianLars (approved):

FabianLars:
thank you

PR #3391, opened 2025-07-05
add another way to add migrations for sql plugin

situ2001:
#### Description

Add this since it might be useful for people using ORM like drizzle that generates migration files automatically.

### DIFF
--- a/src/content/docs/plugin/sql.mdx
+++ b/src/content/docs/plugin/sql.mdx
@@ -220,6 +220,19 @@ let migration = Migration {
 };
 ````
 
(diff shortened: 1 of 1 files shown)


src/content/docs/plugin/sql.mdx
```diff
@@ -220,6 +220,19 @@ let migration = Migration {
 };
 ````
 
+Or if you want to use SQL from file, you can include it by using `include_str!`:
```
FabianLars:
```suggestion
Or if you want to use SQL from a file, you can include it by using `include_str!`:
```

FabianLars (approved):
thanks, valuable tip!

PR #3396, opened 2025-07-09
chore(docs): Extend Tauri default capability permissions to support opener and core defaults

been-there-done-that:
chore(docs): Updating the permissions addons for window customization

This PR updates the `src/content/docs/learn/window-customization.mdx` file to enhance the default permissions for the main Tauri window.

🔧 Changes:
Renamed the identifier from "main-capability" to "default" for consistency with the template.

Expanded the permissions array to include the defaults:
```text
core:default
opener:default
```


Legend-Master:
The goal here was to keep things as minimal as possible so I decided not to take this PR, thanks anyways!

PR #3402, opened 2025-07-15
add plugin workspace to releases

vasfvitor:
current: [build] 5370 page(s) built in 237.36s

PR:       [build] 8665 page(s) built in 572.31s

that's a lot of pages for just the releases of plugins - not mergin this


vasfvitor:
#3987

PR #3406, opened 2025-07-19
fix(example): upgrades window customization example from deprecated cocoa to objc2-app-kit

seanwatters:
#### Description

> Updates docs for changing window color on MacOS to use objc2-app-kit instead of deprecated cocoa crate.


FabianLars (approved):

FabianLars:
Thanks

FabianLars:
Actually we have a background color setting now so we could remove the use of bindings here.

PR #3407, opened 2025-07-19
Update multi-level menu example to include MacOS gotchas

bholmesdev:
#### Description

I copied the multi-level menu example into my app, and was surprised to find some broken behavior on MacOS:
- the "Quit" option was ignored because it was a top-level property
- the "File" menu was swallowed into the application submenu

This updates the multi-level menu to better explain what multi-level menus accomplish, and how to avoid these pitfalls on MacOS. I also updated the JavaScript example to what I'd recommend for cross-system multi-level menus. I noticed the Rust example was entirely different from the JavaScript example, so I didn't touch that one. Curious if that should be updated to match.

### DIFF
--- a/src/content/docs/learn/window-menu.mdx
+++ b/src/content/docs/learn/window-menu.mdx
@@ -143,13 +143,29 @@ fn main() {
 
 ## Creating a multi-level menu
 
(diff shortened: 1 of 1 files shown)


FabianLars (approved):
thank you!

PR #3427, opened 2025-07-30
Rewords the macos notarization section

### LINKED ISSUE #3426: macOS notarization docs poorly worded
The docs state:

> APPLE_ID, APPLE_PASSWORD and APPLE_TEAM_ID: authenticate using your Apple ID
>
> Alternatively, to authenticate with your Apple ID, set the APPLE_ID to your Apple account email and the APPLE_PASSWORD to an [app-specific password](https://support.apple.com/en-ca/HT204397) for the Apple account

This makes it sound like I can use my real password, or as an alternative, an app-specific password. Not sure if this was the intention and actually worked before, or if it's just misworded. 
When trying to notarize with my actual apple ID password I get this error:

> failed to bundle project: Error: HTTP status code: 401. Invalid credentials. Username or password is incorrect. Use the app-specific password generated at appleid.apple.com. Ensure that all authentication arguments are correct.

PS: I was opting to use the real apple ID password because the account was created solely for the purpose of publishing this app. It's not my day-to-day password (in fact, I don't use a mac at all)

naps62:
#### Description

- Rewords macos notarization section
- Closes #3426

### DIFF
--- a/src/content/docs/distribute/Sign/macos.mdx
+++ b/src/content/docs/distribute/Sign/macos.mdx
@@ -6,6 +6,8 @@ sidebar:
 i18nReady: true
 ---
 
+import { Tabs, TabItem } from '@astrojs/starlight/components';
+
 Code signing is required on macOS to allow your application to be listed in the [Apple App Store] and to prevent a warning that your application is broken and can not be started, when downloaded from the browser.
 
 ## Prerequisites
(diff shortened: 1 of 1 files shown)


FabianLars (approved):

FabianLars:
thank you :)

PR #3441, opened 2025-08-04
fix(docs): correct malformed SVG path tag in Tauri documentation

idityaGE:
Fixed a malformed <path> tag in an SVG element within the Tauri documentation. The SVG had an incomplete second <path> tag with no attributes or closing syntax, which could lead to rendering issues or unexpected behavior.

Original (buggy):
```
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
  <path fill="currentColor" d="M19 13H5v-2h14z" /><path>
</svg>
```

Updated:
```
<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24">
  <path fill="currentColor" d="M19 13H5v-2h14z" />
</svg>
```

This change ensures that the SVG is valid and renders correctly.


FabianLars:
Thank you

PR #3450, opened 2025-08-11
feat(develop): update Webdriver tests section

lucasfernog:
- update all dependencies
- adapt CI example workflow for Windows
- update example repository to https://github.com/tauri-apps/webdriver-example

### DIFF
--- a/src/content/docs/develop/Tests/WebDriver/ci.md
+++ b/src/content/docs/develop/Tests/WebDriver/ci.md
@@ -53,7 +53,7 @@ jobs:
         run: |
           sudo apt-get update &&
           sudo apt-get install -y
-          webkit2gtk-4.1
+          libwebkit2gtk-4.1-dev
           libayatana-appindicator3-dev
           webkit2gtk-driver
           xvfb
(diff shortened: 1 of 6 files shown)


src/content/docs/develop/Tests/WebDriver/ci.md
```diff
@@ -42,55 +49,72 @@ jobs:
       # install system dependencies that Tauri needs to compile on Linux.
       # note the extra dependencies for `tauri-driver` to run which are: `webkit2gtk-driver` and `xvfb`
       - name: Tauri dependencies
+        if: matrix.platform == 'ubuntu-latest'
+        run: |
+          sudo apt-get update &&
+          sudo apt-get install -y
+          webkit2gtk-4.1
```
FabianLars:
```suggestion
          libwebkit2gtk-4.1-dev
```
I know webkit2gtk-4.1 works with apt-get's glob thingy but it still triggers me x) (also iirc it doesn't work with apt)

FabianLars (approved):

src/content/docs/develop/Tests/WebDriver/ci.md
lucasfernog: :| I always use webkit2gtk-4.1 for some reason - crazy copypasta

FabianLars: probably because most other distros (arch & fedora for example) don't use the lib prefix, especially arch uses exactly `webkit2gtk-4.1` (if you're still an arch user)

PR #3463, opened 2025-08-16
Update calling-rust.mdx

Erix4:
Replace renamed path_resolver() function with path()

#### Description

- I noticed while following the documentation that the path_resolver() function doesn't exist anymore, so I thought it ought to be updated here


FabianLars (approved):
thank you

PR #3472, opened 2025-08-24
chore(sidecar): shell:allow-open permission is not required

lucasfernog:


FabianLars:
that's from create-tauri-app actually iirc (when it was using shell instead of opener)

PR #3498, opened 2025-09-07
docs(deep-link) Clarifies app or universal links vs custom scheme mobile deep links

monadoid:
#### Description
Clarifies app or universal links vs custom scheme mobile deep links.
This change was introduced here: https://github.com/tauri-apps/plugins-workspace/pull/2870 
And I believe shipped in v 2.4.2.

Since the documentation doesn't reflect these changes, I spent a while trying to figure out how to properly set up just a custom scheme deep link on android, so I thought i'd update the docs.

I think I should also update the README in the plugins repo: https://github.com/monadoid/plugins-workspace/tree/v2/plugins/deep-link 
Could someone confirm if I should put up a second PR there as well?

### DIFF
--- a/src/content/docs/plugin/deep-linking.mdx
+++ b/src/content/docs/plugin/deep-linking.mdx
@@ -84,8 +84,8 @@ Install the deep-link plugin to get started.
 There are two ways to open your app from links on Android:
 
 1. **App Links (http/https + host, verified)**
-  For [app links](https://developer.android.com/training/app-links#android-app-links), you need a server with a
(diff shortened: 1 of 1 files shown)


vasfvitor (approved):
great, thank you

vasfvitor:
> I think I should also update the README in the plugins repo: https://github.com/monadoid/plugins-workspace/tree/v2/plugins/deep-link
Could someone confirm if I should put up a second PR there as well?

Might not be necessary, but you could if you want

PR #3500, opened 2025-09-10
i18n(ja): Japanese Texts for first 6 files in Plugins Section

J1-takai:
#### Description
i18n(ja): Japanese Texts in Plugins Section (part 1):
- Japanese translations of first 6 files: Overview, Autostart, Barcode Scanner, Biometric, CLI and Clipboard plugins.
- Closes #

@vasfvitor  **<Translator's Memo>**
Since the files in Plugins Section are partially generated by ASTROjs programs (and I do not know how it works), almost half of the contents remains untranslated and copied/extracted from the original English texts. For the fully translated version, it may be easier to create, for example, a PDF booklet on these Plugins (incl. both Tauri official ones and the community support version) .... Anyway there is no hurry on this.


src/content/docs/ja/plugin/autostart.mdx
```diff
@@ -0,0 +1,169 @@
...
+import Compatibility from '@components/plugins/Compatibility.astro';
+
+import { Tabs, TabItem, Steps } from '@astrojs/starlight/components';
+import CommandTabs from '@components/CommandTabs.astro';
+import PluginPermissions from '@components/PluginPermissions.astro';
+import TranslationNote from '@components/i18n/TranslationNote.astro';
+
+<TranslationNote lang="ja">
+
+**Plugin 説明内容の英語表記部分について**　Plugin 関連の各章は、「Astro Web フレームワーク」により原文データからページ内容の一部が自動生成されているため、英語表記のままの部分があります。
```
vasfvitor: lets reword these notes, no need to mention Astro as it is the tool we use, its not 'their fault'. Overall this is correct

J1-takai: @vasfvitor 
Thank you for your comment. I have no intention to "blame" ASTRO for the remaining EN texts, and agree with you that there is no need to mention Astro as I just want to pre-notify JA readers that some parts of the plugin docs remain untranslated due to the auto-generated from the EN pages.

J1-takai: My rephrased texts are as follows:

**Plugin 説明内容の英語表記部分について**　Plugin の各章は、原文データからページ内容の一部が自動生成されているため、英語表記のままの部分があります。

which means something like: 
**Note on Remaining EN texts on Plugin Docs**  Each chapter of Plugin Docs has untranslated EN texts because they are auto-generated/extracted from the original EN documents.

Removed Words:  
- 関連; relating to
- 「Astro Web フレームワーク」により; by (using) Astro Web FrameWork

J1-takai: This Pre-Notice will be shown in every Plugin Docs, because the readers check them as needed basis instead of reading through from the top one to the last one.

And, sorry to say that I have issued 12 docs with the old texts in two PRs #3500 and #3546. I will use the above re-worded phrase for the new docs to be translated. 

Thank you always for your careful checks on my JA translations not only this but other PRs.  :bow:

vasfvitor: would you fix the translations with the rephrased texts on both PRs?

J1-takai: I have just edited my PR #3500 and #3546 with the rephrased texts (plus minor improvements of the wording of the tab item lables in #3500).

vasfvitor: great, thanks

J1-takai:
@vasfvitor Thank you for your continuous support!!

PR #3553, opened 2025-10-23
Update macOS installation instructions for LLVM

iarata:
Clarified installation instructions for macOS by mentioning that LLD is provided in a separate formula in Homebrew.

#### Description

- Adds `lld` to `brew install llvm` for building `.msi` on macOS. This is because LLD is now being offered in a separate formula on brew


src/content/docs/distribute/windows-installer.mdx
```diff
@@ -102,10 +102,10 @@ Default Tauri apps should not require this.
 </TabItem>
 <TabItem label="macOS">
 ```sh title=macOS
-brew install llvm
+brew install llvm lld
 ```
 
-On macOS you also have to add `/opt/homebrew/opt/llvm/bin` to your `$PATH` as suggested in the install output.
+On macOS you also have to add `/opt/homebrew/opt/llvm/bin` to your `$PATH` as suggested in the install output (LLD is now provided in a separate formula in Homebrew).
```
FabianLars, line 108: What is this trying to convey? Do users need to add lld manually to the path as well?

iarata: No they just need to install it separately as it's no longer installed along LLVM
Perhaps the description edits wasn't clear enough 🤔

FabianLars: Yeah, i think it's fine to just remove this change. The edit of the `brew install line` should be enough.

PR #3568, opened 2025-10-30
Add troubleshooting information for persisted-scope

Picorims:
Add information regarding the importance of plug-in initialization order in order for this plug-in to work properly.

#### Description

    - This PR adds information to the persisted-scope plug-in page regarding the importance of plug-in initialization order for the plug-in to work properly. The goal is to make it more visible, at it is easily missed in the terminal logs. Maybe `Troubleshooting` is not the right title for this, but considering that it is a requirement, I would emphasize it in the doc somehow so that the user can check up-front. That would also improve its discoverability when looking for solutions online.
- Closes #  (no associated ticket)

### DIFF
--- a/src/content/docs/plugin/persisted-scope.mdx
+++ b/src/content/docs/plugin/persisted-scope.mdx
@@ -66,14 +66,27 @@ Install the persisted-scope plugin to get started.
   </TabItem>
 </Tabs>
 
-## Usage
+:::caution
+The `persisted-scope` plugin *must* be registered and initialized after the `fs` plugin, as illustrated by the example below:
 
-After setup the plugin will automatically save and restore filesystem and asset scopes.
+```rs
+#[cfg_attr(mobile, tauri::mobile_entry_point)]
+pub fn run() {
+    tauri::Builder::default()
+        .plugin(tauri_plugin_fs::init()) // fs MUST BE before persisted scope!
+        .plugin(tauri_plugin_persisted_scope::init())
+        .run(tauri::generate_context!())
+        .expect("error while running tauri application");
+}
+```
 
-## Troubleshooting
-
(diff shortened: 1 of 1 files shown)


src/content/docs/plugin/persisted-scope.mdx
```diff
@@ -69,3 +69,11 @@ Install the persisted-scope plugin to get started.
...
 
 After setup the plugin will automatically save and restore filesystem and asset scopes.
+
+## Troubleshooting
+
+If the plug-in does not seem to save the scope, make sure that this plug-in is initialized AFTER the fs plug-in. If it is not the case, you should see a warning message upon launching your app in dev mode, similar to this:
+
+```
+Please make sure to register the `fs` plugin before the `persisted-scope` plugin!
+```
```
vasfvitor: thanks for this, could you add the instructions directly into the setup step? right after the auto/manual setup tabs, something like: "after installation ensure so and so" with a codeblock illustration where and how it should be. Maybe an [aside](https://starlight.astro.build/guides/authoring-content/#asides) works too

Picorims: I moved them in the setup step below the tabs, within a caution aside (as it is both important and not very noticeable otherwise, I can change if it is too much).

I also added a stripped code block taken from the project where I had the issue and fixed it, so hopefully it should be right but I am not against a double-check.

At the same time I removed the dashes from "plug-in" to be consistent with its syntax on other parts of the page.

I kept a copy of the warning message, not sure how useful it is. I would believe that it gives a hint to browse in the terminal if seeing this after facing the problem, or for referencing it, but maybe that's not important?

vasfvitor: it is ok to keep it, thanks again

vasfvitor (approved):

PR #3573, opened 2025-11-04
Docs: Document transaction handling in migrations

MordechaiHadad:
Added note about transaction handling during migrations.

#### Description
- Why open this PR? I had a rather standard question regarding if the SQL plugin automatically executes migrations inside of transaction, I wasn't able to find an answer in the docs or via the discord channel so I had to look inside of sqlx' and this plugin's code to verify that it does.
- What does this PR change? Adds a small note that migrations are automatically being run inside of transactions.


FabianLars (approved):
thanks

PR #3574, opened 2025-11-04
docs: Update webkit2gtk package for Alpine installation

junovell:
#### Description

Fixed dependency name for alpine linux from `webkit2gtk` to `webkit2gtk-4.1-dev`.

I also want to note that when working with alpine in a container it's required to install a font manually, unlike when using a debian container, should this be mentioned under the prerequisites for alpine or not?

### DIFF
--- a/src/content/docs/start/prerequisites.mdx
+++ b/src/content/docs/start/prerequisites.mdx
@@ -125,7 +125,7 @@ sudo zypper in -t pattern devel_basis
 ```sh
 sudo apk add \
   build-base \
-  webkit2gtk \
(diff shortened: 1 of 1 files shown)


FabianLars:
> should this be mentioned under the prerequisites for alpine or not?

hmm why not

junovell:
i used chatgpt to generate the note so it sounds professional, hope you don't mind that.

FabianLars:
don't care much about that but at least you're giving a disclaimer, that's much appreciated :)

FabianLars (approved):

PR #3575, opened 2025-11-07
Add section on JNI / FFI, Android 16KB memory alignment

bclarke123:
#### Description

Hi, thanks for this fantastic project! I could only find a few code examples to explain how to call Rust directly from Swift / Kotlin, so I added a brief section to the mobile plugin development page. Also, I added an explanation on how to configure Android builds to be compatible with the new 16KB memory alignment requirements. All feedback welcome!

### DIFF
--- a/src/content/docs/develop/Plugins/develop-mobile.mdx
+++ b/src/content/docs/develop/Plugins/develop-mobile.mdx
@@ -355,6 +355,130 @@ Use a nullable type and set the default value on the command function instead.
 Required arguments are defined as `let <argumentName>: Type`
 :::
 
+## Calling Rust From Mobile Plugins
+
+It is often preferable to write plugin code in Rust, for performance and reusability. While Tauri doesn't directly provide a mechanism to call Rust from your plugin code, using JNI on Android and FFI on iOS allows plugins to call shared code, even when the application WebView is suspended.
+
+### Android
+
+In your plugin's `Cargo.toml`, add the jni crate as a dependency:
+
+```toml
+[target.'cfg(target_os = "android")'.dependencies]
+jni = "0.21"
+```
+
+Load the application library statically and define native functions in your Kotlin code:
+
+```kotlin
+private const val TAG = "MyPlugin"
+
+init {
+  try {
+    // Load the native library (libapp_lib.so)
+    // This is the shared library built by Cargo with crate-type = ["cdylib"]
+    System.loadLibrary("app_lib")
+    Log.d(TAG, "Successfully loaded libapp_lib.so")
+  } catch (e: UnsatisfiedLinkError) {
+    Log.e(TAG, "Failed to load libapp_lib.so", e)
+    throw e
+  }
+}
+
+external fun helloWorld(name: String): String?
+```
+
+Then in your plugin's Rust code, define the function JNI will look for:
+
+```rust
+#[cfg(target_os = "android")]
+#[no_mangle]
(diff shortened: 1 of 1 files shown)


src/content/docs/develop/Plugins/develop-mobile.mdx
```diff
@@ -355,6 +355,129 @@ Use a nullable type and set the default value on the command function instead.
...
+    match env.new_string(result) {
+        Ok(jstr) => jstr.into_raw(),
+        Err(e) => {
+            log::error!("Failed to create JString: {}", e);
+            std::ptr::null_mut()
+        }
+    }
+}
+```
+
```
vasfvitor: does `Java_com_example_HelloWorld_helloWorld` has any correlation with the app name, if so, could you add an aside or note about it. Same about the ios section. 

I see patterns but I really don't know if is for illustration or code convention/requirement

thanks

src/content/docs/develop/Plugins/develop-mobile.mdx
```diff
@@ -355,6 +355,129 @@ Use a nullable type and set the default value on the command function instead.
...
+## Android 16KB Memory Pages
+
+Google is moving to make 16KB memory pages a requirement in all new Android app submissions. In order to build Tauri apps that meet this requirement, add the following to `.cargo/config.toml`:
+
+```toml
+[target.aarch64-linux-android]
+rustflags = ["-C", "link-arg=-Wl,-z,max-page-size=16384"]
+```
```
FabianLars, line 479: Is this really required? This seems to be the default in rust already, no? I didn't hear any about any issues with that yet on the rust side while we _did_ hear about it for the barcode-scanner's java dependency for example

bclarke123: It's a new check, that's becoming a requirement - this is a screenshot from the Google Play Console from my app's first Tauri app bundle submission, without that setting:

image

Same console, with this setting in a later build:

image

Learn More link goes [here](https://developer.android.com/guide/practices/page-sizes#build)

Further reading [here](https://developer.android.com/guide/practices/page-sizes)

src/content/docs/develop/Plugins/develop-mobile.mdx
bclarke123: Good note, updated to clarify naming conventions. Thanks for the feedback!

src/content/docs/develop/Plugins/develop-mobile.mdx
FabianLars: yeah i know about the requirement itself, i just didn't see it complain about the rust .so yet.

Here's the analyze output of a fresh tauri project WITHOUT the rustflags on my end
grafik

bclarke123: Updated per chat in Discord, thanks!

FabianLars: Oh wait a seconddddd, does your app make use of the jni stuff you wrote about here? Maybe that's what causing it?

FabianLars: Also these flags will be ignored in a plugin's config so in theory this is the wrong place for this section. i don't think we have a better place for that though x)

vasfvitor (approved):

FabianLars (approved):

PR #3600, opened 2025-11-18
hardcode `tauri ios` help output in the cli doc generator

FabianLars:
guess we should think about exposing these commands on all platforms but returning errors on non-macos devices.


vasfvitor (approved):

vasfvitor:
>  guess we should think about exposing these commands on all platforms but returning errors on non-macos devices.


I don't get it, was this for Tauri repo? Do you mean having a "Windows", "Linux" section? or something like that

FabianLars:
What mean is that `tauri ios build --help` should work on Windows and Linux as well even though the actual commands won't work.
Currently the whole ios command just doesn't exist in the cli on Windows/Linux

PR #3605, opened 2025-11-20
fix(geolocation/ios): proper `Info.plist` key

Vexcited:
#### Description

On iOS, the key in Info.plist for geolocation usage is `NSLocationWhenInUseUsageDescription`.
When mistyped, an error can be logged saying geolocation can not be used because this key is missing.


FabianLars (approved):
my bad - to be fair though, our versions kinda sounds more correct x)

thanks

PR #3631, opened 2025-12-04
docs(window): add userland example for detecting drag-stopped and res…

Khyati-Kapil:
docs(window): add userland example for detecting drag-stopped and resize-stopped events

This PR adds a userland JavaScript example demonstrating how developers can
detect when a window drag or resize operation has *stopped* by listening to
the continuously emitted `tauri://move` and `tauri://resize` events and
debouncing them.

Since native drag-end / resize-end events are not currently exposed by Tauri
and JS event detection is not reliable enough to be implemented in the
runtime, this example provides a recommended, practical solution for
developers needing this behavior.

The new section is added to `learn/window-customization.mdx` where other
window interaction examples (dragging, maximizing, custom titlebars) are
already documented.

### DIFF
--- a/src/content/docs/learn/window-customization.mdx
+++ b/src/content/docs/learn/window-customization.mdx
@@ -306,4 +306,42 @@ Create the main window and change its background color:
     		.run(tauri::generate_context!())
     		.expect("error while running tauri application");
     }
+
+
+    ## Detecting When Window Drag or Resize Has Stopped (Userland Example)
+
+Tauri does not currently expose native “drag finished” or “resize finished” events
+from the operating system. However, you can implement this behavior in userland
+by listening to the continuously emitted `tauri://move` and `tauri://resize`
+events and detecting when movement or resizing stops.
+
+```ts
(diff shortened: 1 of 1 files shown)


Legend-Master:
Hey, thanks for working on this, but I'm not sure how useful is it to leave a very much incomplete solution like this, and by 

> I think this should be done on user land

I meant if a user wants to do it, they can, I don't mean we should document this in our guides, since those guides are meant to be the approaches we recommend

Sorry for being harsh on this, hope you understand 🙏

Khyati-Kapil:
Thank you for the clarification — that makes sense.

I misunderstood the original intention behind “should be done in userland”.
Totally understand your point about the guides reflecting recommended
approaches only. No worries at all — thanks for explaining it clearly.
 
If a more complete or reliable approach becomes possible in the future, I’d be
happy to revisit the idea in a way that aligns with Tauri’s recommendations.

Legend-Master:
I'll close this for now, feel free to re-open if you came up with something more complete in the future, thanks anyways!

PR #3638, opened 2025-12-08
Update Doc Links to /plugin/file-system/ ja version.

J1-takai:
#### Description
"i18n(ja): Updating the Doc links " -->
- Links to /ja/plugin/file-system/#setup" etc are modified to match its ja version (PR#3546).
- Closes #


vasfvitor (approved):
I had to format again, keep an eye when making changes so the formatting doesn't change unnecessarily, it is also easier to see the diff this way.

There are some files that are excluded from prettier auto-formatting because it breaks.

Could you confirm that changes are supposed to be only those 2?

btw, deep-link change was already applied on the other PR, that's why it is missing here

vasfvitor (changes requested):
> 4:38:32 PM: 19:38:32 [ERROR] ✗ Found 2 invalid links in 1 file.
> 4:38:32 PM: 19:38:32 ▶ ja/learn/security/using-plugin-permissions/
> 4:38:32 PM: 19:38:32   ├─ /ja/plugin/file-system/#default-permission - invalid hash
> 4:38:32 PM: 19:38:32   └─ /ja/plugin/file-system/#permission-table - invalid hash
> 

I had to format again, keep an eye when making changes so the formatting doesn't change unnecessarily, it is also easier to see the diff this way.

btw, deep-link change was already applied on the other PR, that's why it is missing here. Also, you need to update the anchor to match the ja page anchor. You can visit the live website if these pages already exists and fix

J1-takai:
@vasfvitor - Checked the above links and found that the current JA version has links to these sections in the EN version (i.e. *without "/ja/"*) as follows:
(probably because not noticing that the links to the auto-generated part is possible, and left the EN links as is):

+ /ja/plugin/file-system/#default-permission - invalid hash
  - Current JA version (Line 122) is: `[fs default](/plugin/file-system/#default-permission)`
+ /ja/plugin/file-system/#permission-table - invalid hash
  - Current JA version (Line 188) is: `[公式ドキュメント](/plugin/file-system/#permission-table)`

As for Deep-Linking plugin, I will update the following link:

+ JA version (line 513): `[single instance]: /plugin/single-instance/` to be `[single instance]: /ja/plugin/single-instance/`

I will prepare new PRs for the above modifications accordingly.

Thanks for your information.

vasfvitor (approved):

J1-takai:
@vasfvitor - sorry for so many adjustments needed for my PR and thank you for your kind assistance. 👍 👍 👍

vasfvitor:
it's ok 👍

Issue #3622, opened 2025-12-01
[docs] [plugins] Community plugins aren't visible enough


IT-ess:
Following this [discussion](https://github.com/orgs/tauri-apps/discussions/14531#discussioncomment-15126863) with @FabianLars , I create this issue. 
I mainly use Tauri for the mobile part, and while Tauri mobile is still quite young, its growth is, imho, probably undermined by the lack of two things : 

- visibility : some plugins that cover (sometimes) basic stuff must be dug on github or crates.io ([mobile task scheduler](https://github.com/charlesschaefer/tauri-plugin-schedule-task), [mobile intents](https://crates.io/crates/tauri-plugin-intent)...)
- consensus: a lot of plugins do the same thing. While it's normal in an open source community to see multiple approaches of the same problem, newcomers should be able to select the default, battletested, community approved approach. The [Push Notifications](https://github.com/tauri-apps/tauri/issues/11651#issuecomment-3324914560) feature depicts well this case. 

Here are some possible actions to address these two issues, from quite to less easy : 
- **Reorganize the [community ressources](https://tauri.app/plugin/#community-resources) page**. Plugins should be filtered or searched by platforms at list, or with a basic search term. This would allow to search among the curated list, and that's a good start.
- **make a kind of crates.io wrapper site**, that would query crates.io results by only matching tauri-plugin-* results. This way we could search among all the published crates, even when their owned hasn't opened a PR on awesome tauri. With this you could also view and compare different plugins that treat similar use cases.
- **On this wrapper site, "Rank" plugins with a metric / label that shows consensus.** Github stars are a good start, but this is platform centric. Labeling good plugins could also be done, maybe a kind of "seal of approval" given by the maintainers or the community for a given usecase ? This may be done through surveys to the community

vasfvitor:
> Reorganize the [community ressources](https://tauri.app/plugin/#community-resources) page. Plugins should be filtered or searched by platforms at list, or with a basic search term. This would allow to search among the curated list, and that's a good start.

we have a specific search box on this page and I just noticed it's broken, it should search among all sections. I'll fix it. It may not be clear how that search works, not sure how to improve that.

About having a platform filter is possible for Tauri official plugins, but not easily for community as we don't have this information readily available.

Thanks for the ideas, will take a look

FabianLars:
> but not easily for community as we don't have this information readily available.

[...]

PR #3642, opened 2025-12-09
feat: community plugin list

vasfvitor:
related #3622

preview: https://deploy-preview-3642--tauri-v2.netlify.app/plugin/index-external-resources/

it should list all tauri-plugin-* found on npm or crates.io. That should cover all plugins that follows our naming convention


packages/community-resources/build.ts
```diff
@@ -0,0 +1,197 @@
...
+  if (!headers) {
+    headers = new Headers();
+  }
+  if (!headers.has('User-Agent')) {
+    headers.set(
+      'User-Agent',
+      'tauri-docs-plugins-discover (https://github.com/tauri-apps/tauri-docs)'
+    );
+  }
+
```
vasfvitor: https://crates.io/data-access#api requires something like that if I understood correctly

FabianLars: correct

.github/workflows/syncCommunityResources.yml
vasfvitor, line 1: should update weekly

src/data/communityResources.json
vasfvitor, line 1: data fetched is committed

PR #3656, opened 2025-12-16
Update example config for Nuxt 4

GigantPro:
- Updated the version with maximum compatibility
- Fixed paths to static files
- Adjusted Nuxt config to ensure proper static site builds

Summary: Update docs for using tauri with Nuxt 4

### DIFF
--- a/src/content/docs/start/frontend/nuxt.mdx
+++ b/src/content/docs/start/frontend/nuxt.mdx
@@ -13,7 +13,7 @@ Nuxt is a meta framework for Vue. Learn more about Nuxt at https://nuxt.com. Thi
 ## Checklist
 
 - Use SSG by setting `ssr: false`. Tauri doesn't support server based solutions.
-- Use default `../.output/public` as `frontendDist` in `tauri.conf.json`.
+- Use default `../dist` as `frontendDist` in `tauri.conf.json`.
 - Compile using `nuxi build`.
 - (Optional): Disable telemetry by setting `telemetry: false` in `nuxt.config.ts`.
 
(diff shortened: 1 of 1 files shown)


src/content/docs/start/frontend/nuxt.mdx
```diff
@@ -64,7 +64,7 @@ Nuxt is a meta framework for Vue. Learn more about Nuxt at https://nuxt.com. Thi
             "beforeDevCommand": "pnpm dev",
             "beforeBuildCommand": "pnpm build",
```
FabianLars: npm yarn and pnpm here use `build` while deno uses `generate`, which one is correct?

src/content/docs/start/frontend/nuxt.mdx
```diff
@@ -115,6 +115,15 @@ Nuxt is a meta framework for Vue. Learn more about Nuxt at https://nuxt.com. Thi
...
       ignore: ['**/src-tauri/**'],
+
+      nitro: {
+        // Generate a static site for Tauri
+        static: true,
+        output: {
+          // Set a hard path for statics
+          publicDir: './.output/public'
+        }
+      }
```
FabianLars: can't it inherit if from ssr: false above (which btw is the only thing they documented here https://nuxt.com/docs/4.x/getting-started/deployment#client-side-only-rendering)

src/content/docs/start/frontend/nuxt.mdx
GigantPro: Indeed! That’s the core mistake. `nuxt generate` builds the app as a static site, while `nuxt build` builds it as an executable server. Therefore, with yarn/npm/pnpm the command listed is incorrect. If you specify the correct command and also set `ssr: false`, everything will work properly. There’s no need to provide separate Nitro settings for this.

FabianLars (approved):
nuxt is such a blackbox for me :D i didn't test these changes but i'm gonna trust you :P Thanks for the PR!

PR #3657, opened 2025-12-17
fix: Corrected the icon specifying an invalid color parameter

normal-coder:
#### Description

fix(distribute): Corrected the APP icon specifying an invalid color parameter.

In the `tauri icon path_to_icon.png --ios-color #fff` command, the parameter value of `--ios-color` is mistakenly recognized as a command comment by the terminal, leading to a failure in execution. It is necessary to add quotes around the color value to avoid this issue.

```
➜  app git:(master) ✗ yarn tauri icon ./src-tauri/icons/icon.png --ios-color #fff
error: a value is required for '--ios-color <IOS_COLOR>' but none was supplied
```

### DIFF
--- a/src/content/docs/distribute/app-store.mdx
+++ b/src/content/docs/distribute/app-store.mdx
@@ -24,12 +24,12 @@ Additionally, you must setup code signing for [macOS][macOS code signing] and [i
 After running `tauri ios init` to setup the Xcode project, you can use the `tauri icon` command to update the app icons.
 
 <CommandTabs
-  npm="npm run tauri icon /path/to/app-icon.png -- --ios-color #fff"
-  yarn="yarn tauri icon /path/to/app-icon.png --ios-color #fff"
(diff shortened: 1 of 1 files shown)


FabianLars:
just out of interest, on which OS and terminal did you see that issue? First time i hear about this - though i can see why this may be problematic so i'm not against merging it.

normal-coder:
> just out of interest, on which OS and terminal did you see that issue? First time i hear about this - though i can see why this may be problematic so i'm not against merging it.

My Environment: `macOS 26.1`+ `iTerm 2`+ `zsh` + `oh-my-zsh`. For most modern terminal applications and shells, I understand that this situation indeed exists.

FabianLars (approved):

FabianLars:
hmm okay thanks :) i'd say that's a pretty common setup so a bit sad nobody reported it before.

PR #3680, opened 2026-01-17
Fix build command in "Build" section of Apple App Store instructions

timrogers:
#### Description

The docs currently include a `tauri build --bundles...` command which doesn't work. This switches to use `npm run tauri build -- --bundles...`, matching all the other example commands on the page.


FabianLars:
Thanks for your pr :)
> matching all the other example commands on the page.

can you make that actually match them by using the CommandTabs component instead of hardcoding npm? would be much appreciated.

PR #3775, opened 2026-04-06
Tweak accent color for light theme

Legend-Master:
The starlight default accent color is a bit too high in chroma (too saturated/contrasty) and since we're already using a custom one for the dark theme, let's also do it for the light theme

Feel free to suggest other colors, I just went with a color offset from the Tauri logo blue using oklch

### DIFF
--- a/src/components/overrides/Footer.astro
+++ b/src/components/overrides/Footer.astro
@@ -28,4 +28,12 @@ const year = new Date().getFullYear();
     margin-inline: auto;
     margin-block: 1rem;
   }
+
+  a {
+    color: var(--sl-color-accent);
+  }
+
+  a:hover {
+    color: var(--sl-color-accent-high);
+  }
 </style>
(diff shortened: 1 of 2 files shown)


FabianLars:
it's a bit too grey-ish/green-ish for me but i'm also colorblind so idk 😂

Legend-Master:
Yeah, the tauri logo blue simply can't be not washed out while having a lightness ~50% (needed for the contrasts)

image

Maybe we can change the hue a bit

Legend-Master:
Tried to target a chroma of 0.15, it should be lesser green/grey now

> https://oklch.com/#0.5,0.15,253,100
>
> image

FabianLars:
i think that's better but imo, just use what you prefer. Maybe some day we'll have a designer in the team again

FabianLars (approved):

PR #3786, opened 2026-04-14
feat: add documentation for 2.11 features

lucasfernog:
mobile multi-window: https://github.com/tauri-apps/tauri/pull/14484

file association: https://github.com/tauri-apps/tauri/pull/14486

### DIFF
--- a/src/content/docs/learn/mobile-file-associations.mdx
+++ b/src/content/docs/learn/mobile-file-associations.mdx
@@ -1,0 +1,137 @@
+---
+title: File Associations on Mobile
+sidebar:
+  order: 3
+i18nReady: false
+---
+
+Tauri supports file associations on Android and iOS, allowing your app to be registered as a handler for specific file types. When a user opens a file that matches your declared associations, the operating system launches your app and delivers the file URL.
+
+On **Android**, file associations are implemented using [intent filters](https://developer.android.com/training/app-links/deep-linking) that the Tauri build system generates automatically from your configuration.
+
+On **iOS**, file associations use [CFBundleDocumentTypes](https://developer.apple.com/documentation/bundleresources/information-property-list/cfbundledocumenttypes) and optionally [UTExportedTypeDeclarations](https://developer.apple.com/documentation/bundleresources/information-property-list/utexportedtypedeclarations) for custom file types.
+
+## Configuration
+
+File associations are declared in `tauri.conf.json` under `bundle.fileAssociations`. The Tauri CLI uses this configuration to generate the appropriate platform-specific metadata (Android intent filters in `AndroidManifest.xml`, iOS document types in `Info.plist`).
+
+Each entry in the array represents a file type your app can handle:
+
+```json title="src-tauri/tauri.conf.json"
+{
+  "bundle": {
+    "fileAssociations": [
+      {
+        "ext": ["png"],
+        "mimeType": "image/png"
+      },
+      {
+        "ext": ["jpg", "jpeg"],
+        "mimeType": "image/jpeg"
+      }
+    ]
+  }
+}
+```
+
+### Configuration Options
+
+- `ext` — list of file extensions to associate (without leading dot).
+- `mimeType` — the MIME type for the file (e.g. `image/png`). Required on Android for intent filter matching. Tauri infers common MIME types from extensions when not specified.
+- `role` — the app's role with respect to the file type. Maps to `CFBundleTypeRole` on Apple platforms. Values: `Editor` (default), `Viewer`, `Shell`, `QLGenerator`, `None`.
+- `rank` — the ranking among apps that handle this file type. Maps to `LSHandlerRank` on Apple platforms. Values: `Default` (default), `Owner`, `Alternate`, `None`.
+- `name` — display name for the file type. Defaults to the first extension.
+- `exportedType` — defines a custom file type owned by your app. Required on Apple platforms when associating with non-standard file extensions.
+- `androidIntentActionFilters` — which Android intent actions to register. Values: `Send`, `SendMultiple`, `View`. All three are used by default.
+
+### Custom File Types
+
+For non-standard file extensions, you should define an `exportedType` so Apple platforms can identify the file type. The `identifier` should be a reverse-DNS string unique to your app, and `conformsTo` lists the parent types:
+
+```json title="src-tauri/tauri.conf.json"
+{
+  "bundle": {
+    "fileAssociations": [
+      {
+        "ext": ["mydata"],
+        "mimeType": "application/octet-stream",
+        "exportedType": {
+          "identifier": "com.example.myapp.mydata",
+          "conformsTo": ["public.data"]
+        }
+      }
+    ]
+  }
+}
+```
+
+Common `conformsTo` values include `public.data`, `public.image`, `public.json`, and `public.plain-text`.
+
+## Handling Opened Files
+
+When a file is opened with your app, Tauri emits a `RunEvent::Opened` event containing the file URLs. This event is available on macOS, iOS, and Android.
+
+You need to handle two cases:
+
+1. **App is already running** — the event is delivered at runtime.
+2. **App is launched by the file open** — the event fires during startup, so you should store the URLs and make them available to your frontend.
+
+### Rust
+
+Store incoming URLs in managed state, expose them with a command the frontend can call on startup, and emit a Tauri event whenever `RunEvent::Opened` fires so the frontend can react while the app is already running:
+
+```rust title="src-tauri/src/lib.rs"
+use std::sync::Mutex;
+use tauri::Manager;
+
+struct OpenedUrls(Mutex<Vec<tauri::Url>>);
+
+#[tauri::command]
+fn opened_urls(app: tauri::AppHandle) -> Vec<tauri::Url> {
+    app.state::<OpenedUrls>().0.lock().unwrap().clone()
+}
+
+#[cfg_attr(mobile, tauri::mobile_entry_point)]
+pub fn run() {
+    tauri::Builder::default()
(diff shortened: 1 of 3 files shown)


FabianLars:
only commenting this so i don't forget it over night. do we also want a few sentences on multi-window support on phones (so without side-by-side)? Not sure if that's even a realistic use-case but it's probably a question that will come up.

lucasfernog:
> only commenting this so i don't forget it over night. do we also want a few sentences on multi-window support on phones (so without side-by-side)? Not sure if that's even a realistic use-case but it's probably a question that will come up.

good catch, pushed

src/content/docs/learn/mobile-multiwindow.mdx
```diff
@@ -0,0 +1,328 @@
...
+
+Tauri supports multiple windows on Android and iOS, allowing your app to display content side-by-side on tablets or in separate scenes on iPad.
+
+On **Android**, multi-window is implemented using [Activity Embedding](https://developer.android.com/guide/topics/large-screens/activity-embedding), which lets the system display two activities side by side on large screens.
+
+On **iOS**, multi-window uses the [UIScene](https://developer.apple.com/documentation/uikit/uiscene) API, which allows iPad users to open multiple instances of your app in separate windows.
+
+On **phones**, the system usually does not lay out two windows side by side. On **Android**, creating another window still launches a separate activity, but on handset-sized displays it is typically **pushed onto the activity back stack**—so **Back** returns to the previous activity instead of closing a split. On **iOS** (especially iPhone), opening or creating another window often **replaces the current UI** with the new scene’s content rather than keeping both visible at once; true concurrent windows remain an **iPad** (and Stage Manager) experience.
```
velocitysystems, line 16: I'm struggling to reconcile this comment with the `app.supportsMultipleWindows` flag. Since this flag is the source of truth as to whether multi-window is supported, should we simply return `false` for phones?

Applications will likely use this flag to detect whether to _enable_ the feature for their users. If it returns `true` on phones but isn't really supported this somewhat defeats its purpose.

src/content/docs/learn/mobile-file-associations.mdx
```diff
@@ -0,0 +1,137 @@
...
+      }
+    ]
+  }
+}
+```
+
+### Configuration Options
+
+- `ext` — list of file extensions to associate (without leading dot).
+- `mimeType` — the MIME type for the file (e.g. `image/png`). Required on Android for intent filter matching. Tauri infers common MIME types from extensions when not specified.
```
velocitysystems, line 40: Comparing this against the Rust docs, I noticed we have a stale comment [here](https://github.com/tauri-apps/tauri/blob/dev/crates/tauri-utils/src/config.rs#L1200):
```rust
/// The mime-type e.g. 'image/png' or 'text/plain'. Linux-only.
```

src/content/docs/learn/mobile-multiwindow.mdx
```diff
@@ -0,0 +1,328 @@
...
+   Create a Kotlin class for each additional window type. Each activity must extend `TauriActivity`:
+
+   ```kotlin title="src-tauri/gen/android/app/src/main/java/com/example/app/DetailActivity.kt"
+   package com.example.app
+
+   import android.os.Bundle
+   import android.os.PersistableBundle
+
+   class DetailActivity: TauriActivity() {
+     override fun onCreate(savedInstanceState: Bundle?, persistentState: PersistableBundle?) {
```
velocitysystems: Should be the following `onCreate` signature to match WRY [implementation](https://github.com/tauri-apps/wry/blob/dev/src/android/kotlin/WryActivity.kt#L112-L117):

```kotlin
override fun onCreate(savedInstanceState: Bundle?) {
   super.onCreate(savedInstanceState)                
}     
```

src/content/docs/learn/mobile-multiwindow.mdx
```diff
@@ -0,0 +1,328 @@
...
+}
+```
+
+</TabItem>
+<TabItem label="Rust">
+
+```rust
+use tauri::Manager;
+
+let main_window = app.get_window("main").unwrap();
```
velocitysystems: Shouldn't this be` get_webview_window()`?` get_window()` was deprecated in V2 and is [gated](https://github.com/tauri-apps/tauri/blob/dev/crates/tauri/src/lib.rs#L541-L545) behind `#[cfg(feature = "unstable")]`.

src/content/docs/learn/mobile-multiwindow.mdx
```diff
@@ -0,0 +1,328 @@
...
+
+## Window Instance APIs
+
+Once a window has been created, you can retrieve its platform-specific identifier:
+
+<Tabs syncKey="lang">
+<TabItem label="JavaScript">
+
+```javascript
+const activityName = await window.activityName();
```
velocitysystems, line 308: Should we add a cross-platform API which abstracts this complexity away from the consumer? How does this interact with the existing desktop APIs?

src/content/docs/learn/mobile-multiwindow.mdx
```diff
@@ -0,0 +1,328 @@
...
+  "windows": ["main"],
+  "permissions": ["core:default", "core:webview:allow-create-webview-window"]
+}
+```
+
+## Android
+
+Android multi-window uses [Activity Embedding](https://developer.android.com/guide/topics/large-screens/activity-embedding) to split activities side by side on large screens (tablets, foldables). You need to create an Android `Activity` for each window type, configure split rules, and register an initializer.
+
+<Steps>
```
velocitysystems, line 46: Android setup is currently five hand-wired steps (add dependencies, create activities, update AndroidManifest.xml, create split initializer, define split rules) versus iOS's one (`Info.ios.plist`).

There's already precedent inside `tauri-build` for generating this kind of plumbing — file associations inject their Android intent filters via `update_android_manifest` with a named marker (`crates/tauri-build/src/mobile.rs`). A similar helper could replace these manual steps.

A config-driven version might look like:

```json
{
  "app": {
    "android": {
      "activityEmbedding": {
        "enabled": true,
        "splitRules": [
          {
            "primary": "MainActivity",
            "secondary": "DetailActivity",
            "splitRatio": 0.33,
            "minWidthDp": 840
          }
        ]
      }
    }
  }
}
```

That would collapse the docs' five steps into "write your activity class, declare your rules." Beyond ergonomics, centralizing this in `tauri-build` would avoid the fragility of the current hand-wired approach.

A similar approach could be used for iOS to: (a) enable scene support, (b) handle scene requests.

velocitysystems (changes requested):

src/content/docs/learn/mobile-multiwindow.mdx
lucasfernog: app.supportsMultipleWindows returns false on iPhones (though for Android phones it returns true if API >= 32; in that case, the system will just push the new activity to the stack, that's just the fallback behavior defined by Android)
the docs just state what happens if you try to create another window even though supportsMultipleWindows return false

src/content/docs/learn/mobile-file-associations.mdx
lucasfernog: nice catch https://github.com/tauri-apps/tauri/pull/15253

src/content/docs/learn/mobile-multiwindow.mdx
lucasfernog: ends up being the same https://android.googlesource.com/platform/frameworks/base/+/master/core/java/android/app/Activity.java#1939 but i'll update for consistency

src/content/docs/learn/mobile-multiwindow.mdx
lucasfernog: nice catch, thanks

src/content/docs/learn/mobile-multiwindow.mdx
lucasfernog: well the abstraction is implemented by default.. by default we use the caller's window as the "parent" of the created window, so the relationship is defined automatically
this is for random freaks who wants to define the relationship manually - or create a window using the app handle instead

we could implement WindowBuilder::parent for mobile, but the relationship is a bit different than mobile's

src/content/docs/learn/mobile-multiwindow.mdx
lucasfernog: iOS is a bit easier, we HAVE to enforce scene support: https://github.com/tauri-apps/tauri/pull/15254 (iOS 27 will require it)

lucasfernog: I think that's a nice addition, tho not a requirement for 2.11 IMO
here's a PR: https://github.com/tauri-apps/tauri/pull/15255

velocitysystems: Great work! Agree - will take a look at the proposed PR.

src/content/docs/learn/mobile-multiwindow.mdx
velocitysystems: Thanks @lucasfernog  — that helps. I think there's still a gap worth addressing in the doc itself, though.                                                                        
                                                                                                                                                                                                           
The concern is that `app.supportsMultipleWindows` returns `true` on Android phones (API ≥ 32) even though the platform's behaviour is push-to-back-stack rather than split. Apps that gate "Open in new window" UI on this flag — which is how the doc presents it as a runtime check — will end up offering a feature that doesn't work the way users expect on phones.

lucasfernog: unfortunately there's no check AFAIK to properly determine whether the platform supports it or not - maybe we should just check if it is a tablet or not?

PR #3810, opened 2026-05-04
fix(debian): libwebkit2gtk version

nnnnoel:
#### Description

- Closes #


FabianLars:
Your PR is against the v1 branch which is for Tauri v1 which is actually still using webkitgtk 4.0 - in other words, our docs are correct - and yes, some distros have removed 4.0 from their repos which kinda forces devs to update to tauri v2.

Thanks for the PR nonetheless.

PR #3817, opened 2026-05-08
change npm package links from npmjs.com to npmx.dev

FabianLars:
We only talked about this _very_ briefly but there was no contra so i thought i'd start here :)


vasfvitor:
much needed, .net is down it seems

PR #3836, opened 2026-05-18
i18n(uz): add Uzbek localization and fix dynamic hero routing

cto-mindset:
Description
This PR introduces the initial Uzbek (uz) localization and fixes a navigation issue in the Hero component.

1. Uzbek Localization (i18n):

Added uz locale to locales.json.

Created Uzbek version of the landing page (src/content/docs/uz/index.mdx).

Added RSS feed page for Uzbek locale (src/content/docs/uz/rss.mdx).

2. Navigation Fix:

Fixed the "Get Started" button in Hero.astro. Previously, it was hardcoded to /start/, which redirected localized users back to the English docs.

The button now dynamically resolves the href based on the current Astro.url.pathname, ensuring users stay within their selected language (e.g., /uz/ -> /uz/start/).

Verified this fix across other locales (Spanish, Korean) to ensure it's a universal solution.

### DIFF
--- a/src/components/Hero.astro
+++ b/src/components/Hero.astro
@@ -5,6 +5,9 @@ import LogoDark from '@assets/logo.svg';
 import LogoOutline from '@assets/logo-outline.svg';
 
 const { tagline, getStarted, v1Docs } = Astro.props;
+const [, locale] = Astro.url.pathname.split('/');
+
+const baseLink = locale && locale !== 'start' ? `/${locale}/start/` : '/start/';
 ---
 
 <div class="hero-bg">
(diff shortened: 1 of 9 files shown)


src/components/Hero.astro
vasfvitor, line 1: can you move this fix to a new PR? also I think it currently only happens on ja locale, not sure why, it may be a configuration elsewhere

vasfvitor: also, do you have anyone to help us review the translation? it would be best, but not blocking if you don't

cto-mindset:
I closed the request to avoid conflicts.

PR #3850, opened 2026-05-26
Fix: update window-customization guide to use objc2_app_kit over cocoa

Downmoto:
#### Description
cocoa crate is deprecated in favour of objc2 crate, and specifically the objc2_app_kit crate for this example. This PR updates the example to use objc2_app_kit.

### DIFF
--- a/src/content/docs/learn/window-customization.mdx
+++ b/src/content/docs/learn/window-customization.mdx
@@ -284,19 +284,19 @@ Create the main window and change its background color:
 
     			// set background color only when building for macOS
     			#[cfg(target_os = "macos")]
-          {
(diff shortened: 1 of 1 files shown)


FabianLars:
long overdue, thanks! in theory i think tauri's built-in bg color api can handle this nowadays but i'd like to keep this for now as it's a good example to show how to use objc in tauri

FabianLars (approved):

Downmoto:
awesome! I noticed a check failing that didn't fail in my draft. Not sure what went wrong there, but it looks like i can use prettier to fix it, will do it this evening.

Downmoto:
sorry for the delay, committed the fix and looks like the check is passing now.

PR #3860, opened 2026-05-31
i18n(fr): Update of vscode.mdx in french

0ri0nexe:
#### Description

**Update**: the file was missing most of its current content, so linking the commits seems unnecessary but [here they are](https://github.com/tauri-apps/tauri-docs/commits/v2/src/content/docs/develop/Debug/vscode.mdx?since=2024-07-23T13:01:26.000Z)

**Notes**: 
- I figured the links to other files in the repo stayed the same by looking at the other translation, if not i might want to change line 5
- The file process-model isn't translated yet, is the translation `Processus Central` good enough for `Core Process` because it needs to be compliant with the translation that will be used in this file.

Anyway have a good one !

### DIFF
--- a/src/content/docs/fr/develop/Debug/vscode.mdx
+++ b/src/content/docs/fr/develop/Debug/vscode.mdx
@@ -2,27 +2,31 @@
 title: Débogage dans VS Code
 ---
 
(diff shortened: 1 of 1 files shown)


Legend-Master (approved):
Thanks

PR #3884, opened 2026-06-10
docs: recommend the Starlight i18n VS Code extension

### LINKED ISSUE #1776: [2.0] Add Starlight i18n VS Code extension recommendation in translating guide
### 📋 Explain your issue

So after @lorenzolewis mentioned the existence of [Starlight i18n](https://marketplace.visualstudio.com/items?itemName=hideoo.starlight-i18n) VS Code extension a while ago, I took a look at it and it does work as intended and would surely help a lot for contributors.

Only catch is that by default it tries to find locale setting directly inside Astro config and cannot reference to our locales.json. So the question would be should we change our config so that Starlight i18n would work by default, or should we try to open an upstream issue at [HiDeoo/starlight-i18n](https://github.com/HiDeoo/starlight-i18n) for overriding and also maybe try to fix it ourselves? 👀

rumitvn:
Closes #1776.

The translating guide already mentions the [Starlight i18n](https://marketplace.visualstudio.com/items?itemName=hideoo.starlight-i18n) extension, but it wasn't in the workspace recommendations. This adds `hideoo.starlight-i18n` to `.vscode/extensions.json` so VS Code suggests it automatically when contributors open the repo.

Per the discussion in #1776, the extension now reads an external `locales.json` without extra config, so no other changes are needed.


Legend-Master:
I see to get `Failed to find the last commit for the file at 'tauri-docs\src\content\docs\reference\javascript\positioner.md'.` when I try running `Starlight i18n` command of this plugin 🤔

PR #3886, opened 2026-06-12
docs: add Linux graphics troubleshooting page for NVIDIA and WebKitGTK

Zeus-Deus:
There is an open request to document the NVIDIA problems people keep hitting with WebKitGTK (tauri-apps/tauri#9394). I hit this myself while chasing input lag in my Tauri app on Arch with an RTX 3090 under Wayland, so I wrote up the known symptoms and workarounds from that issue plus what I found while debugging.

The page covers the common crashes and blank windows with the env var workarounds in order of cost, and a section on the silent failure modes that I have not seen documented anywhere: WebGL context creation succeeds even when it lands on a slow path, and WebKitGTK masks the renderer string (it reports Apple GPU on every Linux machine), so an app cannot check what is behind the context. The advice there is to ship a non WebGL fallback for WebGL heavy views instead of trusting the context.

Happy to adjust tone or move the page if Debug is not the right section.

### DIFF
--- a/src/content/docs/develop/Debug/linux-graphics.md
+++ b/src/content/docs/develop/Debug/linux-graphics.md
@@ -1,0 +1,48 @@
+---
+title: Linux Graphics Issues
+---
+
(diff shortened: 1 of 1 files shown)


FabianLars:
Thanks! This is honestly long overdue so thank you for biting the bullet :D 

> Happy to adjust tone or move the page if Debug is not the right section.

I think it's the most fitting category we currently have.

Zeus-Deus:
Thanks! Debug works for me, I'll leave it there. Also pushed a fix for the prettier check, the workflow just needs another approval to run.

FabianLars (approved):

PR #3903, opened 2026-06-19
Make theme select icon slightly bigger

Legend-Master:
This makes it visually the same as the rest of icons


FabianLars:
i somehow prefer the smaller size there but maybe that's more because the lang dropdown feels so misplaced

FabianLars (approved):

Legend-Master:
Yeah, the font size of that drop down is smaller then the rest of things

PR #3905, opened 2026-06-19
Tweak link styles

Legend-Master:
### Changed links from underlines to just colors 

We have many links in our docs and the underlines made the pages look cluttered

### Removed duplicated `heading-link`

This changed the behavior from `wrap` (entire heading is a link) to a slug besides the title, also fixes the duplicated slugs

### Refreshed the look of card links

I find the default starlight cards to be frankly quite ugly compared to other similar frameworks like docusaurus/vitepress

Before

image

After

image

I have also removed the shadows, if you prefer having the shadow, we can make it a bit lighter instead

### Slightly rounded `code`'s corners

Since everything else's got rounded corners, I don't know why `code` in markdowns are not


FabianLars:
i feel like we should revisit the link colors as well. without the underlines it doesn't register as a link with that color in my head idk - too close to the normal header color maybe

FabianLars (approved):

Legend-Master:
Yeah, planning to address that next

Issue #3788, opened 2026-04-15
[request] add plugin releases to release page


Legend-Master:
It would be nice to have a central place to find change log of plugins from https://github.com/tauri-apps/plugins-workspace in https://v2.tauri.app/release/

vasfvitor:
I've looked into it previously and it becomes quite cumbersome because IIRC Starlight generates one fallback locale page for each page. I'll take a look at it, but I think I'll bring back alive the project that have all releases in a dedicated app, another repo

https://vasfvitor.github.io/tauri-releases/

vasfvitor:
live here https://tauri.app/release/

will leave this issue open because now I see we don't have a link directly to each section

PR #3916, opened 2026-06-22
Re-layout title bar and side bar

Legend-Master:
- Hide `.starlight-sidebar-topics` on desktop
- Removed custom site title since it's too buggy to be useful (it shows `Menu` on every page by default which is not very useful, and clicking on it gives you a drop down which is the same what the hamburger menu gives you)
- Highlights links in header instead 
- Made starlight mobile menu show on home page through overrides
- Moved `Releases` to its own topic, we could eventually migrate it to https://github.com/tauri-apps/tauri-docs/issues/3788#issuecomment-4255160033

---

Before

image

After

image

---

Before

image

After

image


Legend-Master:
~~Seems like we can't simply remove the custom site title yet because of https://github.com/withastro/starlight/issues/977~~ Made it work by overriding starlight components to always show the side bar

FabianLars (approved):
yesss

PR #3935, opened 2026-06-27
Ignore tauri and plugins-workspace submodule changes

Legend-Master:
Seems like they always contain changes, probably because of `node_modules`


dreyfus92 (approved):
nice catch, let's test this out 👀

vasfvitor:
you know this always bugged (assuming is the same issue) me when working on the docs, but it wasn't enough to make me look into it. For tauri repo is the pnpm lockfile, plugins-workspace I don't know

PR #3937, opened 2026-06-28
docs(create-project): add vite.config.ts step to ignore src-tauri fro…

jvhuaiying:
…m Vite watch

- Add step 5 to configure server.watch.ignored in vite.config.ts
- Renumber original step 5 to step 6
- Sync changes between English and Chinese docs

#### Description

- Closes #


Legend-Master (commented):
~~Since this template is from the tauri-cli, let's edit that instead so we don't need to do this manually here~~ never mind, we don't ship vite configs, sorry!

Legend-Master (approved):
Thanks

PR #3958, opened 2026-07-07
Fix header width media queries to align with starlight `md-*`

Legend-Master:
They were using `min-width: 50rem` and it translates to `width >= 50rem`

> https://github.com/withastro/starlight/blob/60e1d28c8521e30b8c321f86caba7e35efabfb62/packages/starlight/style/util.css#L23


vasfvitor:
what happens at = 50rem? it uses sm-*?

vasfvitor (approved):

Legend-Master:
I really don't like those utility css 😂, it's `sl-hidden md:sl-flex` which means always hidden, but when `width >= 50rem`, change it to `display: flex`

PR #3968, opened 2026-07-12
[i18nIgnore] use satteri in components, and grammar fix

vasfvitor:
before/after

nsh
image

image

---
pbxproj
image

image
changed the example to match the syntax
image


Legend-Master (approved):
Awesome, I had a branch working on this but it was gone from a git error `fatal: invalid commit position. commit-graph is likely corrupt` for some reason, anyways, this is better than my version anyways!

astro.config.mjs
```diff
@@ -11,6 +11,19 @@ import path from 'path';
 import { fileURLToPath } from 'url';
 import lunaria from '@lunariajs/starlight';
 import { readFileSync } from 'fs';
+import nsisGrammar from './src/langs/nsis.tmLanguage.json';
```
Legend-Master, line 14: A small nitpick: did you write this yourself or is it from somewhere? If it's from somewhere, we should add a comment about it

vasfvitor: no I didn't, I'd imagine it would be necessary, In the same day I went looking but gave up because I had to sleep. Btw I'm sorry second time I stepped on your feet, this is the last stuff I'm working on for now I just noticed it when running locally so I decided to fix

help me out, for nsis I took from here https://github.com/idleberg/vscode-nsis/blob/12a87f5ae1bf86676d517ace47e3ce79d9272259/syntaxes/nsis.tmLanguage it has a license file on the same directory

https://github.com/idleberg/vscode-nsis/blob/12a87f5ae1bf86676d517ace47e3ce79d9272259/syntaxes/nsis-license.txt

then I converted from tmLanguage to json

pbxproj probably from here
https://github.com/hyperupcall-projects/vscode-assorted-languages/blob/main/syntaxes/pbxproj.tmLanguage.json

do you think a link is enough or what?

Legend-Master: > Btw I'm sorry second time I stepped on your feet

Please absolutely don't mind this, they were better than mine anyways 🙃

> do you think a link is enough or what?

I think just documenting the source and the licenses would be enough since they're all on permissive licenses

vasfvitor: ok, just did, it ytou think it is ok please merge, if anything please nit https://github.com/tauri-apps/tauri-docs/pull/3968/commits/58cdd17c5e2526bdb8685c1afbd1b587b107b5f1

dreyfus92 (approved):
so cool to see satteri here, awesome job @vasfvitor 🤘🏻

astro.config.mjs
Legend-Master: That's good, nicely done 🙃

PR #3974, opened 2026-07-16
Fix TypeScript error: StoreOptions requires the `defaults` property

jvhuaiying:
#### Description

- Closes #


Legend-Master:
Thanks for bring this up! This is a bug and will be fixed in https://github.com/tauri-apps/plugins-workspace/pull/3499

PR #3979, opened 2026-07-18
ci: optimize PR workflows (ubuntu-slim, triage consolidation)

vasfvitor:
to sum up I was annoyed with the long running ci and wanted to optmize, claude was quick on the trigger and opened a pr, I'm still reviewing, but changes were planned. There's one more think I'd like to maybe add and is astro-check

bonus point it found a bug

edit: it also got the changes from my other branch lol

here is the issue:
image

todo: bump action labeler

----
yada yada below:

`TAURI_DOCS_SKIP_IMAGE_OPT=true` to avoid sharp in preview
- Concurrency cancel-in-progress, `timeout-minutes: 30`
- `pnpm check` (`astro check`) runs as an informational step (`continue-on-error: true`) until the pre-existing type-error backlog is cleared.

2. Light jobs → `ubuntu-slim`

3. Automation consolidated into `triage.yml`
`assign.yml` + `labeler.yml` + `priority.yml` + `status.yml` 

fix: always-false `.status == 'open'` guard — the webhook field is `.state`, so **project Status/Priority field-setting has never actually run until now** (704f107c). 

this fix probably will mutate lots of issues/pr status


vasfvitor:
dam claude is quick on the trigger

.github/workflows/build.yml
```diff
@@ -0,0 +1,58 @@
+name: 'Build'
+
+# Full production-shaped build on every PR: runs the reference/config/cli/release
+# generators and `astro build` (includes starlight-links-validator). Catches
+# broken builds and dead internal links before merge — previously only Netlify
+# caught them, post-merge. Image optimization and git-history-derived data
+# (last-updated dates, Lunaria status) are intentionally degraded here for
+# speed — Netlify remains the source of truth for those.
```
vasfvitor: netlify assumption is wrong, we catch errors on preview

.github/workflows/triage.yml
vasfvitor, line 1: aside from merging them all together, I've made a few changes, I hope nothing breaks

package.json
```diff
@@ -50,5 +51,9 @@
   "packageManager": "pnpm@10.34.5",
   "engines": {
     "pnpm": "^10.16.0"
+  },
+  "devDependencies": {
+    "@astrojs/check": "^0.9.9",
+    "typescript": "^5.9.3"
```
vasfvitor, line 57: will bump to typescript 7 later. The security note is about astro-check, currently it is only used locally

FabianLars (commented):
This was added for Simon who's no longer active. Do we still need this? I never understood the value of it.

FabianLars:
I don't think the comment ended up where I wanted it to be. I hate this mobile app. I meant the auto assign pr author thing

FabianLars:
Similarly are we even using the project board?

dreyfus92:
we're not using the project board at all! so ig we can drop that workflow.

vasfvitor:
about the auto assign I feel the same. About the board, I'm also not using, thought about dropping this whole thing but then thought about giving it a try and see if it can be useful

vasfvitor:
I'll drop the workflow and revert the unrelated submoudle changes

vasfvitor:
removed auto assign, priority and status

FabianLars (approved):

dreyfus92 (approved):
awesome job @vasfvitor 👀

PR #3985, opened 2026-07-19
update to typescript 7

vasfvitor:
~starlight needs typescript 5 || 6 due to i18next from starlight (will see if it is a easy fix upstream) but since we still need 5.5 for typedoc I reused it.~ found a workaround, just removed i18next from lockfile and installed again and boom, we're golden

for typedoc with ts7 it seems that it will take longer

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -22,6 +22,12 @@
     {
       "matchPackageNames": ["typedoc", "typedoc-plugin-markdown", "typedoc-plugin-mdn-links"],
       "enabled": false
+    },
+    {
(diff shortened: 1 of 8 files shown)


renovate.json
```diff
@@ -22,6 +22,12 @@
     {
       "matchPackageNames": ["typedoc", "typedoc-plugin-markdown", "typedoc-plugin-mdn-links"],
       "enabled": false
+    },
+    {
+      "description": "typedoc needs the TypeScript JS compiler API, which typescript >= 7 (native compiler) no longer ships. The typescript instance in the typedoc package context must stay on the last JS-based line; the repo root's typescript follows the latest release.",
+      "matchFileNames": ["packages/js-api-generator/**"],
```
vasfvitor, line 28: later in #3986 replace the match to the new typedoc generator

FabianLars (approved):

PR #3987, opened 2026-07-19
feat: releases site under /release

vasfvitor:
adapted https://vasfvitor.github.io/tauri-releases/ (that itself was an extension of our current release generator)  into a separated starlight site that will be served under /release with same ui as the docs. 

it is a separated website to avoid i18n fallback, this makes possible adding many more packages we didn't have under releases.

~working on fixing smaller ui details and some other stuff~

- [x] Create the releases Netlify tauri-releases.netlify.app (base dir packages/releases-site) rest of config netlify will get from netlify.toml 
- [x] Update base brancho on "tauri-releases" netlify

using "tauri-releases" on the config, but can be anything, maybe v2-tauri-releases

demo: https://tauri-docs-releases-demo.netlify.app/release/


vasfvitor:
make-fetch-happen is from npm itself, using it to cache the requests from npm... and others

.github/workflows/refresh-releases.yml
```diff
@@ -0,0 +1,106 @@
+name: Refresh release data
+
+on:
+  schedule:
+    - cron: '00 7 * * *'
+  workflow_dispatch:
```
FabianLars, line 6: do you want us to trigger this from the other repos on release?

packages/releases-site/generator/config.ts
```diff
@@ -0,0 +1,372 @@
...
+export function changelogFilePath(pkg: RepoPackage): string {
+  return pkg.githubPath === '__root__' ? 'CHANGELOG.md' : `${pkg.githubPath}/CHANGELOG.md`;
+}
+
+export function cratesWebUrl(cratesPath: string): string {
+  return `https://crates.io/crates/${cratesPath}`;
+}
+
+export function npmWebUrl(npmPath: string): string {
+  return `https://www.npmjs.com/package/${npmPath}`;
```
FabianLars: can we maybe use npmx.dev like we recently starting using in  the docs?

FabianLars:
https://tauri-releases.netlify.app/ is up now - please remind me to change its base branch after merge

.github/workflows/refresh-releases.yml
```diff
@@ -0,0 +1,106 @@
...
+          git config --global user.name "tauri-bot"
+          git config --global user.email "tauri-bot@tauri.app"
+
+      # Only data.json is committed — pages, tableData.json and latestVersions.ts
+      # are regenerated from it at build time by the releases Netlify site.
+      # Merging this PR only triggers the releases site build (the docs site
+      # build is skipped by the ignore rule in the repo root netlify.toml).
+      # Bot PAT + automerge, same flow as syncSponsorsData.yml (a PAT-created
+      # PR also triggers CI checks, which GITHUB_TOKEN-created PRs don't).
+      - name: Open release refresh PR
```
FabianLars, line 84: can we somehow make it only open the pr when there were changes?

packages/releases-site/generator/config.ts
vasfvitor: sure

.github/workflows/refresh-releases.yml
vasfvitor: not sure, there are so many packages, what do you think? perhaps there's a way to make it like the dependabot

.github/workflows/refresh-releases.yml
vasfvitor: I was pretty sure it already does that by default just like the sponsor, will check it again

FabianLars: i didn't really check. i just didn't see anything at first glance, sorry if i missed it.

.github/workflows/refresh-releases.yml
FabianLars: in tauri we already have similar triggers in the workflow so adding another one isn't that bad. All the plugins share a single workflow file as well so no biggie. create-tauri-app, wry, and tao imo aren't that important and can wait for the daily cron.

.github/workflows/refresh-releases.yml
vasfvitor: I haven't checked yet, but it is the default behavior, if it happens to open empty prs I'll fix

.github/workflows/refresh-releases.yml
vasfvitor: is addding workflow_dispatch: enough?

FabianLars: on the docs side? yes, should be.

vasfvitor: then it is ok

vasfvitor: if you prefer I can open the prs upstream. For plugins and tauri

FabianLars (approved):

.github/workflows/refresh-releases.yml
FabianLars: would be appreciated. the less work for me the better, i'm lazy :D but really, we can just keep it on the cron for now and see if it's good enough for us.

vasfvitor:
@FabianLars please update netlify

FabianLars:
done

Legend-Master:
I wonder the reason why this is in a separate package/startlight site instead of part of the main site?

vasfvitor:
initially it was to avoid i18n fallback for each release page. I found it worked better as a package because now we can build docs and releases separately. Btw, to avoid the fallback issue it should be possible with .astro custom pages. 

did you noticed any issue or regression?

Legend-Master:
There's small layout shift in the header because of the lack of language select dropdown and the page layout looks a bit different in #3991. I'm mostly just concerned about duplications and minor style differences, also the bundles will probably be different so when you go onto one of those pages, the cached js and css can't get re-used.

vasfvitor:
hmmmmm I see, I tried to avoid duplication, so most of the stuff on the releases is imported from the main website, only real duplication is the favicon, and the 2 overrides to remove language select and make the logo redirect to / instead of /release 

I'll take a look at the layout difference you mentioned

so build is decoupled but ui is coupled

Legend-Master:
image

We should also remove those groupings by the way.

The changelog table page https://tauri.app/release/table/ has a `Link` row which seems to be the same as `Changelog` row?

vasfvitor:
agree about the grouping.

the link and modal are the same content yes, now that you mention I see a layout shift when selecting because of a "Rese filters", normally I would ignore but now I'm paying more attention to that 
details

anyways, what do you suggest? While they're the same content, one is a link directly to the release, the other is a quick way to read the notes. Maybe just add a "copy link" directly in the same column as 'see more'.

Legend-Master:
I would probably say make the entire column clickable and remove both see more and link rows.

And I'm planning on gathering some suggestions tomorrow in an issue and ping you 🙃

vasfvitor:
perfect, do it. I have a small list of stuff I want to improve too but first I wanted to get it live

PR #1791, opened 2024-01-04
New Security Concept Page

tillmann-crabnebula:
#### What kind of changes does this PR include?

- New or updated content

#### Description

This revamps the security concept page. It removes the development security content page,
adds new knowledge, removes old and outdated knowledge, centralizes the security concepts
into a single page.

I am happy with the major work on this PR but we need to add v1 specific content like the allowlist.
I am not happy about us having multiple places for v1/v2 selection as we should do this at the root level, but this is out of scope for this PR

PR #3988, opened 2026-07-20
Fix example permission scopes toml

Legend-Master:
Noticed this in https://github.com/tauri-apps/plugins-workspace/pull/3504

I can not believe this was wrong from the very beginning in #1791


vasfvitor (approved):
I'm working on auto generating those files directly from the toml

Legend-Master:
> I'm working on auto generating those files directly from the toml

Awesome

PR #3997, opened 2026-07-21
document plugin metadata

vasfvitor:
related https://github.com/tauri-apps/tauri-docs/issues/3622

in practice, for community plugins won't make a difference in our current docs, we'd need to also fetch the metadata for all plugins we don't own

fix a bug where the search hides community plguins

related https://github.com/tauri-apps/tauri-docs/pull/3642/


src/content/docs/develop/Plugins/index.mdx
vasfvitor, line 1: doc content

dreyfus92 (approved):
thank you @vasfvitor ✌🏻

Issue #3999, opened 2026-07-22
https://tauri.app/release/ feedback


Legend-Master:
Reference: https://github.com/tauri-apps/tauri-docs/pull/3987#issuecomment-5035537034

The layout is a bit different from the rest of the pages, probably because of the custom components.

<img width="2556" height="1190" alt="Image" src="https://github.com/user-attachments/assets/4d5f4c78-846f-4713-ab0f-c85021be1a32" />

<img width="2558" height="1228" alt="Image" src="https://github.com/user-attachments/assets/4a363e3a-d047-4b1b-9fed-07dc95b69acd" />

---

Remove the grouping of tao, wry, create-tauri-app

<img width="415" height="601" alt="Image" src="https://github.com/user-attachments/assets/d46009a5-a342-48ac-8933-6aedba9ed4bd" />

---

Probably a starlight thing, but the cards don't look that good, especially the spacing, version tags and alignments

<img width="1203" height="823" alt="Image" src="https://github.com/user-attachments/assets/0481d007-c499-48b9-b987-6ecb8dbb182f" />

---

No side bar highlight in version pages, this will probably need a rework in terms of how things are placed in the side bar. Also I think we should expand all groups by default in this topic.

<img width="1640" height="835" alt="Image" src="https://github.com/user-attachments/assets/8f358182-857c-4334-8271-72d88cae6a01" />

<img width="1916" height="934" alt="Image" src="https://github.com/user-attachments/assets/09562ed9-5d53-41b1-8bb2-fe240fd7b202" />

---

Wrong escapes, probably a typedoc thing?

<img width="1838" height="858" alt="Image" src="https://github.com/user-attachments/assets/db9b4a31-2e98-4e20-8983-4fc8715c081a" />

---

I know it's a bit of a hard ask, but I find it much easier to look at all related tauri package updates at once, say all 2.11 changes similar to what you can view in https://github.com/tauri-apps/tauri/pull/15041

We are trying to make sure that everything in tauri-utils, tauri-runtime, tauri-runtime-wry changes are reflected in tauri's change log, so probably just grouping the same version tauri/api/cli changes together would be good enough

Thinking something like https://docusaurus.io/blog/releases/3.10 (it's manually written so not 1:1, just an idea on how it might look like)

---

[...]

Issue #3992, opened 2026-07-21
review layout shift on /release vs /


vasfvitor:
also check if styles are correctly shared

vasfvitor:
note to self: rename cta -> create tauri app

vasfvitor:
https://github.com/tauri-apps/tauri-docs/pull/3987

PR #4000, opened 2026-07-23
improve releases page

vasfvitor:
related #3999  #3992 

- [x] toc
- [x] fix shared config and styles
- [x] fix escaping
- [x] fix sidebar highlight
- [x] inline sidebar with only 1 children
- [x] improve cards
- [x] group changelog 

pending: 

- [ ] see about how to display full release first instead of single version, ux/ui-wise, I guess a followup PR
- [ ] other stuff
- [ ] full changelog
- [ ] improve release table

~big problem: search is not shared :/ and worse is that users wouldn't know that~ I think I found a solution, next PR

new card (well?)
image

grouped releases, without fetching the release notes (working on layout)
https://deploy-preview-4000--tauri-releases.netlify.app/release/core/

create tauri and inline if only 1 pkg
image

highlight
image


packages/releases-site/astro.config.mjs
```diff
@@ -16,28 +17,21 @@ export default defineConfig({
...
         Header: './src/components/overrides/Header.astro',
+        // sidebar override for highlighting
+        Sidebar: './src/components/overrides/Sidebar.astro',
+
+        // shared overrides
         Footer: '../../src/components/overrides/Footer.astro',
         ThemeSelect: '../../src/components/overrides/ThemeSelect.astro',
         PageFrame: '../../src/components/overrides/PageFrame.astro',
+        TwoColumnContent: '../../src/components/overrides/TwoColumnContent.astro',
       },
```
vasfvitor, line 33: missing override that caused some layout differences

vasfvitor:
@Legend-Master I'm thinking about the full changelog. Place it partially (latest 5 or so) on the page with a 'see more' that leads to the full page. Below that the current `#Versions`

Legend-Master:
> @Legend-Master I'm thinking about the full changelog. Place it partially (latest 5 or so) on the page with a 'see more' that leads to the full page. Below that the current `#Versions`

That sounds good to me, maybe a bit more than 5 like 20 or so that it wouldn't be too long anyways?

Awesome work by the way, I could take a proper review next Monday 🙃

src/shared-config.mjs
```diff
@@ -0,0 +1,39 @@
+// Starlight config values shared between the docs site (astro.config.mjs)
```
Legend-Master: Still thinking about merging the two sites

Legend-Master (commented):
Image

We'll probably need to reduce the heading levels in the changelog markdowns and provide TOC for the patch versions as well

src/shared-config.mjs
vasfvitor: I'll do one and see if it works, what is your biggest grip with it being separated? (aside from the maintenance)

vasfvitor:
> We'll probably need to reduce the heading levels in the changelog markdowns and provide TOC for the patch versions as well

what do you mean? going deepr than this and showing individual packages?
image

Legend-Master:
- 2.11
  - 2.11.1
  - 2.11.2
  - 2.11.3
- 2.10

Something like this if we're going with the current structure (we have patch versions like 2.11.1 under this TOC with separators right now)

src/shared-config.mjs
Legend-Master: Well, mainly just maintenance, and some thoughts I talked about before: https://github.com/tauri-apps/tauri-docs/pull/3987#issuecomment-5029830382

2 different sites can leads to subtle differences very easily and making it harder to maintain, I still don't quite get the reasoning for 2 sites

vasfvitor:
I see

src/shared-config.mjs
vasfvitor: well, initially was because of i18n https://github.com/tauri-apps/tauri-docs/pull/3402 making the build time almost double when adding more packages so I had my mind set to it. The part that I said about making .astro pages isntead of .md I haven't tried and it should be possible, for me it was naturally to just split to make it work with what I had, as I have done in a separated repo and then decided to bring here

Considering your opinion I'll do it, I agree that ideally it should be together as so far it brought the unexpected issue of the search being split and what else could come and bite me lol. Plus, so far build time isn't a real problem since we never build releases on preview anyway 

For reference, currently the release deploys in 32s and main website in around 120s netlify time, not build time.

vasfvitor:
image


I think this is bad because of the second scrollbar


what if on sidebar, this way:

Changelog Table

   - Core Releases
      - 2.11
      - 2.10
     ...

   - tauri
   - tauri-cli
   - @tauri-apps/api



---

well, there are some inconsistencies because of release date, some minors of the group were release separated, I guess the above solution is cleaner

image

Legend-Master:
> well, there are some inconsistencies because of release date, some minors of the group were release separated, I guess the above solution is cleaner

The last image you posted looks good if things are properly grouped?

vasfvitor:
image


I'll adopt this above pattern for the all releases too, what do you think?
image

Legend-Master:
This looks really good, nicely done! 

> I'll adopt this above pattern for the all releases too, what do you think?

Fantastic, go ahead.

A small nitpick would be 

image

to

image

but it's probably a bit hard.

vasfvitor:
I actually went through that but with 2 vertical lines, in the end left only the outer one

Legend-Master:
Nice!

vasfvitor:
will move to main site and have the 'all releases' in view in next pr. For merging into main website I took a look and most of the code will still live in the /packages. I think its gonna work pretty well

PR #4016, opened 2026-07-27
chore(deps): remove unused deps

vasfvitor:
~https://github.com/tauri-apps/tauri-docs/commit/d07cc5693bcc7e70f95efbff8ae56f1ada0908fc
remove jsdom and use mdast-util instead (there are some visual differences, will annotate, if anythin will split into 2 prs)~

and remove other pkgs that weren't being used


packages/fetch-sponsors/config.ts
```diff
@@ -1,11 +1,18 @@
...
+    if (existsSync(path.join(dir, 'pnpm-workspace.yaml'))) return dir;
+    if (path.dirname(dir) === dir) throw new Error(`no workspace root above ${from}`);
+  }
+}
 
 // FIXME: This is used by both this package and the starlight build environment,
 // so we can't use `import ProjectContext from 'virtual:starlight/project-context'` here
 // (starlight seems to copy all the dependency files out so dynamic relative paths don't work)
-const workspaceRoot = searchForWorkspaceRoot(process.cwd());
+const workspaceRoot = findWorkspaceRoot(process.cwd());
```
vasfvitor, line 15: will address de FIXME in another PR

package.json
```diff
@@ -30,22 +30,18 @@
...
     "prettier": "3.9.6",
     "prettier-plugin-astro": "^0.14.1",
     "sass": "^1.89.0",
     "sharp": "^0.35.3",
-    "shiki": "^4.3.1",
     "starlight-blog": "^0.28.0",
     "starlight-links-validator": "^0.25.2",
     "starlight-llms-txt": "^0.11.0",
-    "starlight-sidebar-topics": "^0.8.0",
-    "vite": "^8.1.4"
```
vasfvitor: json schema not sure why we have it, didn't see any import

astro-feelback was never used
shiki and vite we're redeclaration, fetch-sponsors used it but no need to have specified just to use one function

packages/js-api-generator/package.json
```diff
@@ -11,7 +11,6 @@
   "license": "MIT",
   "private": "true",
   "dependencies": {
-    "github-slugger": "^2.0.0",
```
vasfvitor, line 14: unused here, but still used in other packages

PR #4031, opened 2026-08-04
Move profiles to root Cargo.toml

AntoniosBarotsis:
Profiles should be specified at the workspace root, not in sub-crates.

- Build warning:
  ```
  warning: profiles for the non root package will be ignored, specify profiles at the workspace root:
  package:   <PROJECT_DIR>/src-tauri/Cargo.toml
  workspace: <PROJECT_DIR>/Cargo.toml
  ```
- [Rust docs](https://doc.rust-lang.org/cargo/reference/workspaces.html): "_The [...] `[profile.*]` section in `Cargo.toml` are only recognized in the root manifest, and ignored in member crates’ manifests._"

### DIFF
--- a/src/content/docs/concept/size.mdx
+++ b/src/content/docs/concept/size.mdx
@@ -19,7 +19,7 @@ Dependent on whether you use the stable or nightly Rust toolchain the options av
 <TabItem label="Stable">
 
 ```toml
-# src-tauri/Cargo.toml
+# Cargo.toml
 [profile.dev]
 incremental = true # Compile your binary in smaller steps.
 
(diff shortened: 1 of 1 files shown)


FabianLars:
The vast majority of tauri projects have the root Cargo.toml file in src-tauri/ since those devs are the target audience of our docs i'd keep it as is (if you know how to add a root cargo.toml workspace you likely also know how to read the warning and fix it ^^ )

I wouldn't mind a small sentence saying that this has to be in the workspace root (if there's one) - can't think of the wording rn.

dreyfus92 (commented):
we could add a `:::note` along the lines of "if `src-tauri` is a member of a cargo workspace, profiles must go in the workspace root `Cargo.toml` instead. cargo ignores profile sections in member crates and warns about it. this way covers both audiences accurately instead of trading one group's correctnes for the other's. wdty @FabianLars?

AntoniosBarotsis:
Oh I was unaware, I'm pretty sure I made my project using whatever commands the docs mention and I ended up like this. I added another crate at one point so I probably made/changed the workspace at that point. 

We can probably close this then if tauri's default behaviour doesn't cause this 😅

PR #3986, opened 2026-07-19
feat: use starlight-typedoc

vasfvitor:
superseeds #3410 #3424 #3910

there are a few workarounds that I'll see about fixing upstream

depends on #3985 

there are slight differences for the better

https://deploy-preview-3986--tauri-v2.netlify.app/reference/javascript/autostart/
https://tauri.app/reference/javascript/autostart/

https://deploy-preview-3986--tauri-v2.netlify.app/reference/javascript/dialog/
https://tauri.app/reference/javascript/dialog/

`<T>` from Channel being removed is not a regression, was a fix in typedoc
https://deploy-preview-3986--tauri-v2.netlify.app/reference/javascript/api/namespacecore/
https://tauri.app/reference/javascript/api/namespacecore/

todo:
~bug on code block~
https://deploy-preview-3986--tauri-v2.netlify.app/reference/javascript/opener/
https://tauri.app/reference/javascript/opener/

new page
https://deploy-preview-3986--tauri-v2.netlify.app/reference/javascript/shell/

PR #4042, opened 2026-08-06
feat: use starlight-typedoc - #3986

FabianLars:
Rebase of https://github.com/tauri-apps/tauri-docs/pull/3986

Closes #4052


Legend-Master:
I wonder if it's possible to show the parameter types inline, instead of in a table below

image

compared to 

image

vasfvitor:
yeah afaik there's an option, btw should I bring it closer to v1 like the example section being up top?

Legend-Master:
> btw should I bring it closer to v1 like the example section being up top?

yeah if the parameter types are inlined with just the type references below, then they should be below everything else

vasfvitor:
enum as table
image

expand objects
image
image
at [reference/javascript/api/namespacedpi](https://deploy-preview-4042--tauri-v2.netlify.app/reference/javascript/api/namespacedpi/#__tauri_to_ipc_key__)


---
https://deploy-preview-4042--tauri-v2.netlify.app/reference/javascript/api/namespacedpi/
without expand
image
with expand objects
image

vasfvitor:
I'm going to ~rebase~ merge and do a bit of work on the css

Legend-Master:
Also it seems like currently the functions were shown at the end, would it make sense to show them first? Since most of the useful things we provide are the functions as classes, the types, enums are all helpers

vasfvitor:
you mean this?
image

specifically for this page https://deploy-preview-4042--tauri-v2.netlify.app/reference/javascript/api/namespaceapp/#type-aliases I'd argue that moving functions up would bury the other down below on the TOC. Unless the opposite is also true in other pages (many enum and helpers that hides the function on TOC), in that case either way this would be an issue so it ends as a matter of preference

vasfvitor:
or I could reduce the toc depth to 2 levels

Legend-Master:
Yeah that's what I meant. I wouldn't mind this too much either ways that this should only be for looking up certain APIs anyways

vasfvitor:
@Legend-Master what do you think about the toc depth? It is also possible to remove the 'Functions', 'Enumerations' headings but I tried and it not so clear at glance

before/after toc depth (not same image but close)
image

image


before/after toc depth
image
image

PR #4056, opened 2026-08-13
docs(configuration-files): clarify that arrays are replaced, not merged

aesopfrom0:
Adds a note on how platform-specific configuration merges, and adjusts the example so the behaviour is visible.

The RFC 7396 link is already there, but the current example can't show the difference: `plugins` is an object so it merges key by key, and `bundle.resources` has a single element on both sides, so replacing and merging produce the same result. Adding a second element to the base `resources` makes the Linux override visibly drop it.

`app.windows` is where this bites, since its entries are objects and look like they would merge field by field. I hit this porting two apps to Windows: the entry in `tauri.windows.conf.json` only set a few fields, and the width, height and minimum size from the base config were gone. The configuration is valid and builds.


src/content/docs/develop/configuration-files.mdx
```diff
@@ -90,13 +90,22 @@ In addition to the default configuration file, Tauri can read a platform-specifi
 
 The platform-specific configuration file gets merged with the main configuration object following the [JSON Merge Patch (RFC 7396)] specification.
 
+:::note
+Objects are merged key by key, but arrays are replaced as a whole, including arrays nested in objects.
+In the example below `bundle` is merged, while `bundle.resources` is replaced.
+This also applies to `app.windows`, even though its entries are objects: a platform-specific entry
+replaces the base entry instead of being merged into it, so any field you omit falls back to its default
+rather than the value in your base configuration.
+Repeat every field you want to keep.
```
dreyfus92: hey @aesopfrom0, thank you for working on this. i'd reword this since this is specific for the windows case, for arrays of strings it's elements, not fields.
```suggestion
Repeat everything you want to keep.
```

aesopfrom0:
Good catch, thanks. Scoped the "fields" sentence to arrays whose elements are objects, and left the string-array case on its own.

dreyfus92 (approved):
this looks good to me, but i'd like @FabianLars to have a second check on this ✌🏻

FabianLars (approved):

PR #4065, opened 2026-08-18
docs(fs): clarify that permissions require an explicit scope

### LINKED ISSUE #3536: [bug] Unscoped "fs" permissions don't work unless you specify an "allow"
### Describe the bug

The issue is that for an fs permission for example `fs:allow-exists` if you just enable the permission by itself under `capabilities/default.json` under "permissions", then that does not apply and you will still meet a "forbidden path" error (for example under your `/home/` directory).

**Not working for a directory `/assets` under home such as `/home/foo/assets`:**
(under `src-tauri/capabilities/default.json`)
```
"permissions": [
  "core:default",
  "fs:default",
  "fs:allow-exists"
]
```

**Working for a directory `/assets` under home such as `/home/foo/assets`:**
```
"permissions": [
  "core:default",
  "fs:default",
  {
    "identifier": "fs:allow-exists",
    "allow": [{ "path": "$HOME" }, { "path": "$HOME/**/*" }]
  }
]
```

In both examples using the following JS code with the `@tauri-apps/plugin-fs` and `@tauri-apps/api` packages installed:
```
const home = await path.homeDir();
const localAssetDir = `${home}/foo/assets`;
const assetDirExists = await exists(localAssetDir); // Throws an error "forbidden path: /home/<username>/foo/assets"
console.log(assetDirExists);
```

### Reproduction

1. Install the File System plugin using the steps provided in the docs here: https://v2.tauri.app/plugin/file-system/#setup
2. Add the `fs:allow-exists` permission to the capabilities/default.json file under "permissions".
3. Create a fake dir under the home directory such as `/foo/assets`
4. Use the JS code example given to check for the existence of that path.
5. Observe the console for a "forbidden path" error being thrown.
6. Now modify the permission to the format specifying the identifier as the same permission and specifying the home directory and anything within under "allow" like the example above.
7. Observe the console where now "true" is logged.

### Expected behavior

This is either a bug or if the behavior is intended, then it should be more explicitly stated in the docs around permissions and scopes that "allow" is required for it to work.

- If this is a bug then expected behavior is that setting an unscoped permission works effectively globally (since it's not scoped to any directory).
- If this is not a bug and is intended behavior, then this should be **explicitly stated** somewhere in the docs around permissions and scopes sections so it's easier to find for newcomers.

### Full `tauri info` output

```text
tauri info
[✔] Environment
    - OS: Ubuntu 22.4.0 x86_64 (X64) (Unknown DE on Unknown Session)
    - pnpm: 10.16.1
    - npm: 11.6.0

[-] Packages
    - tauri 🦀: 2.8.5
    - tauri-build 🦀: 2.4.1
    - wry 🦀: 0.53.3, (outdated, latest: 0.53.4)
    - tao 🦀: 0.34.3
    - @tauri-apps/api : 2.8.0
    - @tauri-apps/cli : 2.8.4

[-] Plugins
    - tauri-plugin-fs 🦀: 2.4.2
    - @tauri-apps/plugin-fs : 2.4.2
    - tauri-plugin-log 🦀: 2.7.0
    - @tauri-apps/plugin-log : not installed!

[-] App
    - build-type: bundle
    - CSP: unset
    - frontendDist: ../dist
    - devUrl: http://localhost:5173/
    - framework: Vue.js
    - bundler: Vite
```

### Stack trace

```text

```

### Additional context

Mentioned first here in another issue related to "forbidden path" issue from documentation confusion: https://github.com/tauri-apps/tauri/issues/11338#issuecomment-3374770559

Rohan170603:
- Adds a caution callout to the fs plugin's Permissions section clarifying that enabling a permission (e.g. `fs:allow-exists`) alone does not grant access to any path — an explicit `allow` scope is also required, or calls fail with a `forbidden path` error at runtime. Includes a working vs. non-working example.
- Closes #3536

### DIFF
--- a/src/content/docs/plugin/file-system.mdx
+++ b/src/content/docs/plugin/file-system.mdx
@@ -618,6 +618,30 @@ By default all potentially dangerous plugin commands and scopes are blocked and 
 
 See the [Capabilities Overview](/security/capabilities/) for more information and the [step by step guide](/learn/security/using-plugin-permissions/) to use plugin permissions.
 
+:::caution[Permissions alone do not grant a scope]
+Enabling a permission such as `fs:allow-exists` by itself does **not** allow access to any path. Most `fs` commands also require a **scope** that explicitly lists which paths the command may access. Without an `allow` scope, calls will fail at runtime with a `forbidden path` error, even though the permission is enabled.
+
+```json title="Not enough — permission has no scope, so no paths are allowed"
+{
+  "permissions": ["fs:default", "fs:allow-exists"]
+}
+```
+
+```json title="Works — the identifier form adds an explicit allow scope"
+{
+  "permissions": [
+    "fs:default",
+    {
+      "identifier": "fs:allow-exists",
+      "allow": [{ "path": "$HOME" }, { "path": "$HOME/**/*" }]
+    }
+  ]
+}
+```
+
+See [Scopes](#scopes) below for the full list of path variables you can use in `allow` and `deny`.
+:::
+
 ```json title="src-tauri/capabilities/default.json" ins={7-11}
 {
   "$schema": "../gen/schemas/desktop-schema.json",


Legend-Master (approved):
Thanks!

Rohan170603:
@Legend-Master! Thank you for reviewing and merging this! I appreciate the feedback and the quick turn-around. Looking forward to contributing more.

PR #4067, opened 2026-08-20
fix: remove apps sections from AwesomeTauri component

FabianLars:
https://github.com/tauri-apps/awesome-tauri/pull/876 / https://discord.com/channels/616186924390023171/662624589037436928/1539687911912374322

idk if any other changes are needed but didn't find anything

### DIFF
--- a/src/components/AwesomeTauri.astro
+++ b/src/components/AwesomeTauri.astro
@@ -14,23 +14,6 @@ const sections = {
   'guides-no-official-no-video': 'Guides & Tutorials',
   'guides-no-official-only-video': 'Guides & Tutorials',
   templates: 'Templates',
-  'applications-audio-video': 'Audio & Video',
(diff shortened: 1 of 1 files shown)


src/components/AwesomeTauri.astro
```diff
@@ -14,23 +14,6 @@ const sections = {
   'guides-no-official-no-video': 'Guides & Tutorials',
   'guides-no-official-only-video': 'Guides & Tutorials',
   templates: 'Templates',
-  'applications-audio-video': 'Audio & Video',
```
Legend-Master, line 17: I don't actually see this being used anywhere before either?

Legend-Master (approved):

src/components/AwesomeTauri.astro
FabianLars: yep, that was a "in case we need" it kinda thing i think? or maybe it was planned to use it idk

PR #3826, opened 2026-05-14
docs(plugins): document plugin and permission identifier syntax rules

MavenRain:
#### Description

  - Expands the existing "Naming Convention" section of the Plugin Development
  guide with an "Identifier Rules" subsection covering character constraints,
  length limits, the colon-prefix syntax, and a valid/invalid examples table.
  Surfaces the exact error message the build produces on a violation so readers
  can find this section by searching for it.
  - Addresses tauri-apps/tauri#11262 (level 1 of the issue's three-tier ask).
  The level-2 ask (improving the error message itself to identify the file and
  the plugin-name context) is a separate code change targeted at
  tauri-apps/tauri.

PR #4074, opened 2026-08-20
docs(plugins): document plugin and permission identifier syntax rules - #3826

FabianLars:
rebase of #3826


FabianLars:
i can't get these 3 commits out of the rebases, idk where they even come from

PR #4075, opened 2026-08-20
docs(prerequisites): note about static libs for Alpine/musl builds

FabianLars:
rebase of #3840


FabianLars:
idk how correct that is but don't care

PR #4089, opened 2026-08-28
docs: explain NSIS hook paths

AG064:
#### Description

- Explains how `${__FILEDIR__}` behaves at the top level of an NSIS hook and inside an inserted hook macro.
- Shows how to save the hook directory before defining a macro that needs a local file.
- Related to tauri-apps/tauri#11837.

#### Checks

- Minimal NSIS reproduction with MakeNSIS 3.11
- `pnpm exec prettier --check --plugin prettier-plugin-astro src/content/docs/distribute/windows-installer.mdx`
- `pnpm build`
- `pnpm build:astro` after the final wording review


src/content/docs/distribute/windows-installer.mdx
```diff
@@ -580,6 +580,17 @@ Then you must configure Tauri to use that hook file:
 }
 ```
 
+Tauri includes the hook file from its generated NSIS script. At the top level of a hook file, `${__FILEDIR__}` is the hook file's directory, so you can include a file beside it directly. NSIS expands hook macros in the generated script. Save the directory before defining a macro if the macro needs a file beside the hook:
+
+```nsh
+!include "${__FILEDIR__}\\adjacent.nsh"
```
dreyfus92: should be a backslash here. NSIS doesn't treat `\` as an escape in string literals, so this emits as `\\` in the path. 

```suggestion
!include "${__FILEDIR__}\adjacent.nsh"
```

src/content/docs/distribute/windows-installer.mdx
```diff
@@ -580,6 +580,17 @@ Then you must configure Tauri to use that hook file:
...
+!define HOOK_FILE_DIR "${__FILEDIR__}"
+
+!macro NSIS_HOOK_PREINSTALL
+  File "${HOOK_FILE_DIR}\\example.txt"
```
dreyfus92: same here

```suggestion
  File "${HOOK_FILE_DIR}\example.txt"
```

src/content/docs/distribute/windows-installer.mdx
```diff
@@ -580,6 +580,17 @@ Then you must configure Tauri to use that hook file:
 }
 ```
 
+Tauri includes the hook file from its generated NSIS script. At the top level of a hook file, `${__FILEDIR__}` is the hook file's directory, so you can include a file beside it directly. NSIS expands hook macros in the generated script. Save the directory before defining a macro if the macro needs a file beside the hook:
```
dreyfus92: i'd reword this a bit, nothing wrong with the explaination it's a bit too compressed.


```suggestion
#### Including Files Next to the Hook

Tauri `!include`s your hook file from its generated NSIS script. At the top level of the hook file, `${__FILEDIR__}` is the hook file's directory, so you can include a file beside it directly. Macro bodies, however, are expanded where Tauri inserts them in the generated script, so inside a hook macro `${__FILEDIR__}` resolves to the generated script's directory instead. If a macro needs a file beside the hook, capture the directory in a `!define` at the top level first:
```

dreyfus92 (changes requested):
awesome work @AG064, left a few comments! 😄

src/content/docs/distribute/windows-installer.mdx
AG064: Fixed in dccd36fc. Thank you for catching the extra backslash.

src/content/docs/distribute/windows-installer.mdx
AG064: Fixed here as well in dccd36fc.

src/content/docs/distribute/windows-installer.mdx
AG064: Reworked the explanation and added the subsection in dccd36fc. Thank you for the clear review.

dreyfus92 (approved):
thank you @AG064 ✌🏻

PR #4114, opened 2026-09-09
[i18nIgnore] docs: fix a/an article errors

wahidrizka:
Three spots in the docs use the wrong indefinite article. The choice between "a" and "an" follows the sound of the next word, so this corrects each:

- `develop/calling-frontend.mdx`: "we emit a open-file event" to "an open-file event"
- `security/future.mdx`: "written in an memory unsafe language" to "a memory unsafe language"
- `security/scope.mdx`: "in combination with a enabled command" to "an enabled command"

Text only, no code or output changes.


Legend-Master (approved):
thanks

PR #4126, opened 2026-09-13
feat: v3 docs (alpha)

lucasfernog:


src/content/docs/develop/cef.mdx
```diff
@@ -0,0 +1,277 @@
...
+
+- **Windows** and **Linux**: `libcef`, its resources, and the `en-US` locale pack are shipped with the app, next to the executable on Windows and in the AppImage, and under `/usr/share/<product name>` in the Debian and RPM packages. The Windows installers skip the WebView2 installation step, and the Debian and RPM packages do not depend on webkit2gtk.
+- **macOS**: `Chromium Embedded Framework.framework` is placed in the app bundle's `Contents/Frameworks`, along with five helper apps (`<App> Helper`, `Helper (GPU)`, `Helper (Renderer)`, `Helper (Plugin)`, and `Helper (Alerts)`). The bundler compiles the helper executable at bundle time against the `cef` crate version your app resolved. The app is signed with the entitlements Chromium's JIT needs under the hardened runtime (`com.apple.security.cs.allow-jit`, `com.apple.security.cs.allow-unsigned-executable-memory`, and `com.apple.security.cs.disable-library-validation`), so [code signing and notarization](/distribute/sign/macos/) work as usual.
+
+### Linux
+
+The Debian and RPM packages install Chromium's setuid `chrome-sandbox` helper with mode 4755, so packaged apps always have a working sandbox.
```
FabianLars, line 272: That reminds me that i wanted to add an AppArmor exception profile thingy to the installers.

src/content/docs/develop/cef.mdx
```diff
@@ -0,0 +1,277 @@
...
+An AppImage cannot: its payload is mounted `nosuid`, so it relies on unprivileged user namespaces, which some distributions restrict. See [Process Sandbox](#process-sandbox) for what happens there.
```
FabianLars, line 273: didn't check recent commits, did you remove the namespace hook? If not, then it will ask the user to create an apparmor exception for the appimage, which kinda sucks so i'm considering to remove it and do what the docs say. would be cool to hook that up to the `SandboxPolicy::Required` but that would require moving that policy to the config which i think sucks.

FabianLars:
@vasfvitor Do you remember any prior discussions about docs v3? I don't... Did we say we'll have them separate like we have v1 and v2?

Probably makes sense at least in the beginning or in general since we have no time to develop anything else

vasfvitor:
@FabianLars I don't recall much, just that it was mentioned

I just looked into https://starlight-versions.vercel.app and I guess it would just complicate stuff, even if it works based on branch.

(now to everyone) Do you have any preferences on how to present v3/v2 docs, visually or idk any other details? 

I will add a version selector, v1 already has it. Then a banner "check out tauri v3 alpha" or so on current docs, then deploy this branch v3 on netlify (I have permissions now)

- [ ] v2 - pin submodules and changelog to v2 only. (btw I already merged v3 alpha changelog)
- [ ] v3 - point v3 branch to v3.tauri.app
- [ ] uhh

do we want to point beta.tauri.app to v3.tauri.app too? Right now it goes to tauri.app

FabianLars:
I think the v1 and v2 split works well and would do the same for v3. Don't know if starlight has something better nowadays though.

I'd disable beta.tauri.app completely in favor of sharing explicit vX.tauri.app links.

